PR #1, opened 2022-11-05
Add package.json and ts

mfucci:
Add package.json and minimal ts shell so this can be published to npm

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
     "url": "https://github.com/project-chip/matter.js.git"
   },
   "scripts": {
-    "clean": "rm -rf build",
+    "clean": "rm -rf dist",
     "build": "npm run clean; tsc",
     "test": "TZ=utc mocha"
   },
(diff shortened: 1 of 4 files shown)


turon (approved):
LGTM :+1:

package.json
```diff
@@ -0,0 +1,42 @@
+{
+  "name": "matter.js",
+  "version": "0.0.0",
```
turon, line 3: Perhaps 0.0.1?

tsconfig.json
```diff
@@ -0,0 +1,13 @@
+{
+  "compilerOptions": {
+    "target": "es6",
+    "module": "CommonJS",
+    "outDir": "build",
```
turon, line 5: I thought `dist` is more common?

package.json
mfucci: Discussed offline, 0.0.0 is to remove any expectation on npm that this contains something functional

tsconfig.json
mfucci: Correct. Updated to dist

PR #2, opened 2022-11-05
Add Schema and BitmapSchema

mfucci:
This adds the core definition of a generic schema, with one implementation to encode / decode Bitmaps.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,5 +1,5 @@
 /**
  * @license
- * Copyright 2000 Example, Inc
+ * Copyright 2022 Project CHIP Authors
  * SPDX-License-Identifier: Apache-2.0
  */
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,0 +1,5 @@
+/**
+ * @license
+ * Copyright 2000 Example, Inc
+ * SPDX-License-Identifier: Apache-2.0
+ */
--- a/src/schema/BitmapSchema.ts
+++ b/src/schema/BitmapSchema.ts
@@ -14,21 +14,24 @@ type BitSchema = {[key: string]: BitPosition};
 type TypeFromBitSchema<T extends BitSchema> = {[K in keyof T]: boolean};
 
 /** Declares a bitmap schema by indicating the bit position and their names. */
-export interface BitmapSchema<T extends BitSchema> extends Schema<TypeFromBitSchema<T>, number> {}
-export const BitmapSchema = <T extends BitSchema>(bitSchemas: T) => ({
-    encode: (value: TypeFromBitSchema<T>) => {
-        let result = 0;
-        for (const name in bitSchemas) {
-            if (value[name]) result |= 1 << bitSchemas[name].position;
-        }
-        return result;
-    },
+export const BitmapSchema = (bitSchemas: BitSchema) => new class<T extends BitSchema> extends Schema<TypeFromBitSchema<T>, number>{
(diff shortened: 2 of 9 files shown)


src/schema/BitmapSchema.ts
```diff
@@ -0,0 +1,34 @@
+/**
+ * @file Define schema for a bitmap.
+ * @copyright Project CHIP Authors 2022
+ * @license Apache-2.0
+ */
```
mfucci, line 5: What do you think of using those short license / copyrights declarations?
They follow the JsDoc 3 from here: https://jsdoc.app/

turon: This is the compliant license style:
```
/**
 * @license
 * Copyright 2000 Example, Inc
 */
```

mfucci: Compliant with Matter license declaration?

Per JsDoc 3 doc, the SPDX Id can be used directly after @license:
https://jsdoc.app/tags-license.html

For the @copyright tag, it is supposed to use in combination with a @file tag:
https://jsdoc.app/tags-copyright.html
Maybe should be @copyright 2022 Project CHIP Authors to follow the recommendation from inside Apache 2.0 license
https://github.com/project-chip/matter.js/blob/main/LICENSE#L189

I don't mind using your suggestion above or the full license declaration, I just thought the short version was neat...

turon: Yes, I like the proposed format too, but for now it is most expedient to stay with the format that has been pre-approved by legal.

mfucci: Ok, I will update that when I get a break in my day time job! (In 3h...)

mfucci: Updated

src/index.ts
```diff
@@ -0,0 +1,5 @@
+/**
+ * @license
+ * Copyright 2000 Example, Inc
```
turon, line 3:
```suggestion
 * Copyright 2022 Project CHIP Authors
```

test/schema/BitmapSchemaTest.ts
```diff
@@ -0,0 +1,41 @@
+/**
+ * @license
+ * Copyright 2000 Example, Inc
```
turon, line 3:
```suggestion
 * Copyright 2022 Project CHIP Authors
```

src/schema/BitmapSchema.ts
```diff
@@ -0,0 +1,37 @@
...
+
+/** Specifies the position of the bit. */
+export interface BitPosition { position: number }
+export const BitPosition = (position: number) => ({ position });
+
+type BitSchema = {[key: string]: BitPosition};
+type TypeFromBitSchema<T extends BitSchema> = {[K in keyof T]: boolean};
+
+/** Declares a bitmap schema by indicating the bit position and their names. */
+export const BitmapSchema = (bitSchemas: BitSchema) => new class<T extends BitSchema> extends Schema<TypeFromBitSchema<T>, number>{
```
turon, line 17: Would it be possible to simplify this definition to use simple `export class<T> ...` syntax?

turon (approved):

src/schema/BitmapSchema.ts
mfucci: Unfortunately not and this declaration was actually buggy.
Fixed now but even more complex...

PR #3, opened 2022-11-06
Add tlv schema for unsigned integer

mfucci:
After making the schema generic, it made more sense to merge TlvCodec / TlvObjectCodec and Template into the Schema object.

This PR implements an unsigned integer encoding / decoding only without the validate() method.
After we are converged on the code structure, I will implement validate() in the next PR, splitting UnsignedNumberTlv into UInt8Tlv / UInt16Tlv, UInt32Tlv, UInt64Tlv than add other data type Tlv schemas.


src/schema/TlvSchema.ts
```diff
@@ -0,0 +1,242 @@
...
+ * Schema to encode an unsigned integer in TLV.
+ * 
+ * @see {@link MatterCoreSpecificationV1_0} § A.11.1
+ */
+class UnsignedNumberSchema extends TlvSchema<number | bigint> {
+    /** @override */
+    protected encodeTlv(writer: DataWriterLE, value: number | bigint, tag: TlvTag = {}): void {
+        const size = this.getUnsignedIntSize(value);
+        this.encodeControlByteAndTag(writer, ElementType.UnsignedInt | size, tag);
+        this.encodeUnsignedIntBytes(writer, value, size);
```
turon, line 216: Could these be calls to static methods in `TlvCodec`?

src/schema/TlvSchema.ts
```diff
@@ -0,0 +1,242 @@
...
+    }
+
+    /** @override */
+    protected decodeTlv(encoded: DataReaderLE) {
```
turon, line 220: Could this implementation be moved to a static method in `TlvCodec`?

test/schema/TlvSchemaTest.ts
```diff
@@ -0,0 +1,64 @@
...
+
+import assert from "assert";
+import { UnsignedNumberTlv } from "../../src/schema/TlvSchema";
+import { arrayBufferFromHex, arrayBufferToHex } from "../../src/util/ArrayBuffer";
+
+describe("TlvSchema", () => {
+
+    context("encode", () => {
+        it("encodes an 1 byte unsigned int using in TLV", () => {
+            const result = UnsignedNumberTlv.encode(1);
```
turon, line 15: Could these tests get simplified into one small test driven by a testVector[] of `{input:, output:}`?
Example:
https://github.com/project-chip/matternode/blob/main/test/tlv/TestTlvCodec.js#L24

src/schema/TlvSchema.ts
mfucci: Will do.

test/schema/TlvSchemaTest.ts
mfucci: Will do.

src/schema/TlvSchema.ts
mfucci: Will do.

src/schema/TlvSchema.ts
```diff
@@ -0,0 +1,242 @@
...
+    /** Encodes a TLV value. */
+    protected abstract encodeTlv(writer: DataWriterLE, value: T, tag?: TlvTag): void;
+}
+
+/**
+ * Schema to encode an unsigned integer in TLV.
+ * 
+ * @see {@link MatterCoreSpecificationV1_0} § A.11.1
+ */
+class UnsignedNumberSchema extends TlvSchema<number | bigint> {
```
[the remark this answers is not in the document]
mfucci: Will do.

src/schema/TlvSchema.ts
```diff
@@ -0,0 +1,238 @@
...
+}
+
+const COMMON_PROFILE = 0x00000000;
+const UNSPECIFIED_PROFILE = 0xFFFFFFFF;
+const ANONYMOUS_ID = 0xFFFFFFFF;
+
+abstract class TlvSchema<T> extends Schema<T, ArrayBuffer> {
+
+    /** @see {@link MatterCoreSpecificationV1_0} § A.7 */
+    protected decodeTagAndTypeSize(reader: DataReaderLE) {
```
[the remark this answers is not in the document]
mfucci: Done

mfucci:
@turon Updated per your comments

src/schema/TlvSchema.ts
```diff
@@ -0,0 +1,238 @@
...
+    FullyQualified8Bytes = 0xE0,
+}
+
+
+interface TlvTag {
+    profile?: number,
+    id?: number,
+}
+
+const COMMON_PROFILE = 0x00000000;
```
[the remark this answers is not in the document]
mfucci: Done

src/tlv/TlvCodec.ts
```diff
@@ -0,0 +1,162 @@
...
+import { BitmapSchema, EnumBits } from "../util/schema/BitmapSchema";
+
+
+/**
+ * TLV element types.
+ * 
+ * @see {@link MatterCoreSpecificationV1_0} § A.7.1
+ */
+export enum TlvType {
+    SignedInt_1OctetValue = 0x00,
```
turon, line 18: Rename `1OctetValue` --> `U8Value` here and all similar locations?

src/util/schema/BitmapSchema.ts
```diff
@@ -0,0 +1,63 @@
...
+    Flag,
+    Enum,
+}
+
+/** Defines the bit position of a boolean flag. */
+export interface FlagBit { type: BitRangeType.Flag, offset: number }
+export const FlagBit = (offset: number) => ({ type: BitRangeType.Flag, offset }) as FlagBit;
+
+/** Defines the bit position and bit lenght of an enum flag. */
+export interface EnumBits<E extends number> { type: BitRangeType.Enum, offset: number, length: number }
```
turon, line 19: Perhaps rename `EnumBits` -> `BitField` or `BitFieldEnum`?

src/tlv/TlvCodec.ts
```diff
@@ -0,0 +1,162 @@
...
+    ContextSpecific = 1,
+    CommonProfile2Bytes = 2,
+    CommonProfile4Bytes = 3,
+    ImplicitProfile2Bytes = 4,
+    ImplicitProfile4Bytes = 5,
+    FullyQualified6Bytes = 6,
+    FullyQualified8Bytes = 7,
+}
+
+const ControlOctetSchema = BitmapSchema({
```
turon, line 60: I think we should be consistent and always use `Byte` rather than mix terms and use `Octet` sometimes.

src/util/schema/BitmapSchema.ts
```diff
@@ -0,0 +1,63 @@
...
+import { Schema } from "./Schema";
+
+const enum BitRangeType {
+    Flag,
+    Enum,
+}
+
+/** Defines the bit position of a boolean flag. */
+export interface FlagBit { type: BitRangeType.Flag, offset: number }
+export const FlagBit = (offset: number) => ({ type: BitRangeType.Flag, offset }) as FlagBit;
```
turon, line 16: Rename `FlagBit` to `BitFlag`?  I like grouping all BitmapSchema subtypes under the prefix `Bit`.

src/tlv/TlvCodec.ts
mfucci: I tried to stay close to the name used in the specs "Signed Integer, 1-octet value" but fine to shorten it.

src/tlv/TlvCodec.ts
mfucci: Agreed with consistency.
I think the specs use Octet everywhere. I personally prefer Byte...
If we use Byte and want to be consistent, this means that we will use ByteString type instead of OctetString type everywhere.

So the choices / tradeoffs are:
1 _ use Byte everywhere but not be consistent with Matter spec naming
2 _  use Octet everywhere but this is less common than Byte
3 _  use Byte in our codebase and only Octet when there is a need to respect Matter specs naming

Probably the best choice is 2, so I should rename Byte to Octet...
WDYT?

src/tlv/TlvCodec.ts
mfucci: Done, I just used 8 / 16 / 32 / 64 to described the variants.

src/tlv/TlvCodec.ts
mfucci: Done, I renamed Octet to Byte everywhere

src/util/schema/BitmapSchema.ts
mfucci: Done

src/util/schema/BitmapSchema.ts
mfucci: Done

turon (approved):

PR #6, opened 2022-11-07
Add unsigned int validation

mfucci:
I think the next PR should split UnsignedIntegerTlv in its own file since TlvSchema file starts to be too big.

I suggest to move them to this directory, since this is Matter-specific:
/src/matter/tlv

### DIFF
--- a/.github/workflows/build-test.js.yml
+++ b/.github/workflows/build-test.js.yml
@@ -1,0 +1,18 @@
+name: Build and test
+
+on:
+  push:
+    branches: [ "main" ]
+  pull_request:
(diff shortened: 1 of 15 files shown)


turon:
> I suggest to move them to this directory, since this is Matter-specific: /src/matter/tlv

I agree with the spirit of moving all Matter TLV related codec, parsing, and schema handling to `src/matter/tlv` or even `src/tlv`.

mfucci:
Merging this down to the base branch (add-tlv-schema) to be able to refactor to /tlv directory

PR #14, opened 2022-11-11
Ensures that the code works in a browser

mfucci:
I did this now since I started to be afraid that I was using APIs that don't exist in browsers.

This was much harder than I thought:

1. CommonJs modules (node.js default) are not supported in browser
2. Since ES6/ES2015, import statement has been added that is now well supported by browsers
3. Most node.js tools never migrated to ES6 modules and even in node.js support is experimental

So I was faced with the choice:

1. Keep CommonJs which will require a polyfill library to be able to run in a browser
2. Switch to ES6 modules with more difficulties to run on node.js

Since I believe that ES6 modules are the future, I decided to go with solution 2.

Mocha doesn't support ES6 modules, so I had to migrate to jest instead
ES6 modules require the name file to be specified in the import `import "./MyModule.js"` instead of `import ./MyModule`
But node.js doesn't support .js extension in some situations, while this is mandatory to run in a browser.


turon:
Given browsers don't support udp, should running in a browser remain a goal?

turon (approved):

mfucci:
The main goal of this PR is to ensure that the codebase is not nods.js specific.
Works in the browser so there are no hidden node.js API dependencies!

PR #15, opened 2022-11-11
Remove assert dependency since jest include similar APIs

mfucci:


turon:
I like using the assert module because it seems more standard to me than using package-specific alternative.
Is it that important to optimize dependencies on the test code?  I would expect the runtime code which does need to be optimal to not use either assert nor jest.

turon (approved):

mfucci:
The assert library seemed pretty primitive: seriously, I have to write my own code to test if two floats are close to each other?
And it requires to have an extra tsconfig.test.json file to be able to import it because it is distributed as an ES6 module.

jest define describe / it / expect as global so no need to import them which is nice.

expect syntax is the same for jest and jasmine so it seems standard enough, and easy to read as well.

PR #17, opened 2022-11-12
Add TLV array

mfucci:


test/tlv/TlvArray.ts
```diff
@@ -0,0 +1,53 @@
...
+        it("encodes an array", () => {
+            const result = schema.encode(["a", "b", "c"]);
+
+            expect(result.toHex()).toBe("160c01610c01620c016318");
+        });
+    });
+
+    describe("decode", () => {
+        it("decodes an array", () => {
+            const result = schema.decode(ByteArray.fromHex("160c01610c01620c016318"));
```
turon, line 32: I'd suggest adding a codec test vector and adding a few more tests (arrays of various types perhaps).

turon (approved):

test/tlv/TlvArray.ts
mfucci: Added in PR #20

PR #27, opened 2022-11-17
Report spec names in jasmine tests in web and node.js tests

mfucci:


mfucci:
I was not able to get the coverage working with nyc for node.js unit tests: problems between jasmine / nyc / ts / ES6 modules.
It is working fro the same unit tests in the browser through karma, so I guess this is not essential.

turon (approved):

PR #28, opened 2022-11-26
Add additional APIs to support node-matter

mfucci:


src/tlv/TlvNumber.ts
```diff
@@ -0,0 +1,101 @@
...
+/** Unsigned integer TLV schema. */
+export const TlvFloat = new TlvShortNumberSchema(TlvType.Float, _value => TlvLength.FourBytes, FLOAT32_MIN, FLOAT32_MAX);
+export const TlvDouble = new TlvShortNumberSchema(TlvType.Float, _value => TlvLength.EightBytes);
+export const TlvInt8 = new TlvShortNumberSchema(TlvType.SignedInt, value => TlvCodec.getIntTlvLength(value), INT8_MIN, INT8_MAX);
+export const TlvInt16 = new TlvShortNumberSchema(TlvType.SignedInt, value => TlvCodec.getIntTlvLength(value), INT16_MIN, INT16_MAX);
+export const TlvInt32 = new TlvShortNumberSchema(TlvType.SignedInt, value => TlvCodec.getIntTlvLength(value), INT32_MIN, INT32_MAX);
+export const TlvInt64 = new TlvLongNumberSchema(TlvType.SignedInt, value => TlvCodec.getIntTlvLength(value), INT64_MIN, INT64_MAX);
+export const TlvUInt8 = new TlvShortNumberSchema(TlvType.UnsignedInt, value => TlvCodec.getUIntTlvLength(value), 0, UINT8_MAX);
+export const TlvUInt16 = new TlvShortNumberSchema(TlvType.UnsignedInt, value => TlvCodec.getUIntTlvLength(value), 0, UINT16_MAX);
+export const TlvUInt32 = new TlvShortNumberSchema(TlvType.UnsignedInt, value => TlvCodec.getUIntTlvLength(value), 0, UINT32_MAX);
```
turon, line 95: `TlvShortNumberSchema` seems like an odd name for a 32-bit type.
I'd suggest just calling the 32-bit or less one `TlvNumberSchema`.

src/tlv/TlvCodec.ts
```diff
@@ -140,15 +141,15 @@ export class TlvCodec {
     }
 
     /** @see {@link MatterCoreSpecificationV1_0} § A.7 */
-    public static readTagType(reader: DataReaderLE): { tag: TlvTag, typeLength: TlvTypeLength } {
+    public static readTagType(reader: DataReader): { tag?: TlvTag, typeLength: TlvTypeLength } {
```
turon, line 144: Where does Little-Endian get specified now?

turon (approved):

src/tlv/TlvCodec.ts
turon: I see the DataReader constructor has endian argument and defaults to little.

src/tlv/TlvNumber.ts
mfucci: Done

src/tlv/TlvCodec.ts
mfucci: Little-Endian is specified when the DataReader is constructed and there is no default.
I generalized DataReader to support LE and BE since node-matter uses both and matter.js will probably soon.

Looking at this code now, I realized that this change was not correct, this reader has to be LE for the codec to work, so generalizing it weaken this constraint, fixing this now by create again DataReaderLE extending DataReader

PR #29, opened 2022-11-27
Add missing dist-cjs files

mfucci:


package.json
```diff
@@ -40,7 +40,8 @@
     "typescript": "^4.9.3"
   },
   "files": [
-    "dist/**/*"
+    "dist/**/*",
+    "dist-cjs/**/*"
```
turon, line 44: Can we have all distribution files under a singular `dist/` with subdirectories for cjs, es, and coverage?

turon (approved):

package.json
mfucci: Yes, I can change that in next release since this will be transparent for matter.js users

PR #36, opened 2022-12-04
Fix bug when passing ByteArray.fromHex an empty string

mfucci:


turon (approved):

src/util/ByteArray.ts
```diff
@@ -42,6 +42,7 @@ Uint8Array.prototype.equals = function (other: Uint8Array) {
 }
 
 Uint8Array.fromHex = function (hexString: string) {
+    if (hexString.length === 0) return new ByteArray(0);
```
turon, line 45: Would a regression test be hard to write?

mfucci: Yep! I just added all the unit tests for ByteArray, regression test included!

PR #37, opened 2022-12-08
Add ByteArrayBitmap schema

mfucci:
It uses the same syntax as Bitmap schema but the result is stored in a ByteArray rather than a number.
This is made to support cases where bitmap is longer than 32 bits.
I need this for the QR pairing code data in node-matter...

### DIFF
--- a/src/schema/BitmapSchema.ts
+++ b/src/schema/BitmapSchema.ts
@@ -4,15 +4,17 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import { Schema } from "./Schema.js";
+import { ByteArray } from "../util/ByteArray";
+import { Schema } from "./Schema";
 
 const enum BitRangeType {
     Flag,
     Number,
     Enum,
 }
-type BitRange<T> = { type: BitRangeType, offset: number, mask: number };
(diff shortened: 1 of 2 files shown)


src/schema/BitmapSchema.ts
```diff
@@ -4,45 +4,59 @@
...
+const BitRange = <T>(type: BitRangeType, offset: number, length: number ) => ({ type, offset, length } as BitRange<T>);
 
 /** Defines the bit position of a boolean flag. */
 export interface BitFlag extends BitRange<boolean> { type: BitRangeType.Flag }
 export const BitFlag = (offset: number) => BitRange(BitRangeType.Flag, offset, 1) as BitFlag;
 
 /** Defines the bit position and bit length of a numeric value. */
 export interface BitField extends BitRange<number> { type: BitRangeType.Number }
-export const BitField = (offset: number, length: number) => BitRange(BitRangeType.Number, offset, length) as BitField;
+export const BitField = (offset: number, length: number ) => BitRange(BitRangeType.Number, offset, length) as BitField;
```
turon, line 25: Is this trailing space before `)` imposed by lint / style guide?  Or just a stray character?

turon (approved):

src/schema/BitmapSchema.ts
mfucci: Just a stray... removed!

PR #38, opened 2022-12-12
Ignore unknown TLV object fields

mfucci:
This crashes node-matter when using it with HA Matter plugin.


src/tlv/TlvObject.ts
```diff
@@ -71,8 +72,12 @@ export class ObjectSchema<F extends TlvFields> extends TlvSchema<TypeFromFields<
             if (profile !== undefined) throw new Error("Structure element tags should be context-specific.");
             if (id === undefined) throw new Error("Structure element tags should have an id.");
             const fieldName = this.fieldById[id];
-            if (fieldName === undefined) throw new Error(`Unknown field ${id}`);
-            const { field, name }= fieldName;
+            if (fieldName === undefined) {
+                // Ignore unknown field
```
turon, line 76:
```suggestion
                // Ignore unknown field by decoding it so we skip forward the proper length.
```

turon (approved):

src/tlv/TlvObject.ts
mfucci: Done!

PR #42, opened 2022-12-29
Fix Descriptor attributes IDs

mfucci:


Apollon77 (approved):

turon (approved):
Was this an error in the spec vs. SDK or just a typo in the initial checkin?

Apollon77:
@turon a typ in initial commit

PR #44, opened 2023-01-15
OptionalWritableAttribute should be also return type

Apollon77:


Apollon77:
@mfucci Same error in node-matter, will also provide PR soon

mfucci (approved):
Great catch!

PR #46, opened 2023-01-23
Validate decoded data (as also done for Numbers)

Apollon77:


src/tlv/TlvString.ts
```diff
@@ -39,7 +39,9 @@ export class StringSchema<T extends TlvType.ByteString | TlvType.Utf8String> ext
 
     override decodeTlvInternalValue(reader: TlvReader, typeLength: TlvTypeLength) {
         if (typeLength.type !== this.type) throw new Error(`Unexpected type ${typeLength.type}.`);
-        return reader.readPrimitive(typeLength) as TlvToPrimitive[T];
+        const value = reader.readPrimitive(typeLength) as TlvToPrimitive[T];
+        this.validate(value);
```
mfucci, line 43: The whole validation chain is messed up so should probably be addressed as a whole with a consistent logic rather than  just here.

The Schema entry point decode() calls validate() after decoding so adding here as well will actually call validate twice on String TLV.

The whole complexity comes from shallow vs recursive validation. If validate() is recursive, calling it in decode() should enough. However, we will traverse the tree twice, which is sub-optimal. I tried here to make shallow validate, validating only the current tree level, so it can be called during the decoding of the tree.
However validate() is also called at other times, for instance when a attribute is set.

So:
* the whole current design is incorrect
* the current implementation is completely wrong

Two solutions:
* gave up on optimization, so validate() is recursive and only called it from decode() is fine. Work to fix -> modify all validate methods to make them recursive
* fix the current design by introduce a validateShallow() method. Validate shallow is called decoding and validate() should not be called in decode(), current validate() method should be made recursive calling shallowValidate() then validate(0 on children.

With solution 2, the problem is the Schema is also used for non-TLV data structure that are not recursive. So validate() should be called in decode().
So maybe better to go with solution 1 ?

mfucci (changes requested):

src/tlv/TlvString.ts
Apollon77: Ok, In this I need to look in deeper then :-)

Apollon77:
I close this here ... lets continue in https://github.com/project-chip/matter.js/pull/49 :-)

PR #49, opened 2023-02-05
Adjust logic to validate once optionally after decoding

Apollon77:
Ok, as discussed in https://github.com/project-chip/matter.js/pull/46/files#r1094125357 here is my PR to change the following:
* decoding is "just" doing decode and after the decoding we do a recursive validation
* the decode method has an option to skip te validation after decoding.

You could already have a look if you have time. I currently work to add a complex Object/Array/different type test

### DIFF
--- a/src/tlv/TlvNumber.ts
+++ b/src/tlv/TlvNumber.ts
@@ -39,6 +39,10 @@ export class TlvNumericSchema<T extends bigint | number> extends TlvSchema<T> {
 
     override validate(value: T): void {
         if (typeof value !== "number" && typeof value !== 'bigint') throw new Error(`Expected number, got ${typeof value}.`);
+        this.validateBoundaries(value);
+    }
+
+    validateBoundaries(value: T): void {
         if (this.min !== undefined && value < this.min) throw new Error(`Invalid value: ${value} is below the minimum, ${this.min}.`);
         if (this.max !== undefined && value > this.max) throw new Error(`Invalid value: ${value} is above the maximum, ${this.max}.`);
     }
@@ -80,8 +84,7 @@ export class TlvNumberSchema extends TlvNumericSchema<number> {
 
     override validate(value: number): void {
         if (typeof value !== "number") throw new Error(`Expected number, got ${typeof value}.`);
-        if (this.min !== undefined && value < this.min) throw new Error(`Invalid value: ${value} is below the minimum, ${this.min}.`);
-        if (this.max !== undefined && value > this.max) throw new Error(`Invalid value: ${value} is above the maximum, ${this.max}.`);
+        this.validateBoundaries(value);
     }
 }
 
--- a/src/tlv/TlvNumber.ts
+++ b/src/tlv/TlvNumber.ts
@@ -77,6 +77,12 @@ export class TlvNumberSchema extends TlvNumericSchema<number> {
             minValue(max, this.max),
         );
     }
+
+    override validate(value: number): void {
+        if (typeof value !== "number") throw new Error(`Expected number, got ${typeof value}.`);
+        if (this.min !== undefined && value < this.min) throw new Error(`Invalid value: ${value} is below the minimum, ${this.min}.`);
+        if (this.max !== undefined && value > this.max) throw new Error(`Invalid value: ${value} is above the maximum, ${this.max}.`);
+    }
 }
 
 export const TlvLongNumberSchema = TlvNumericSchema<number | bigint>;
--- a/src/tlv/TlvNumber.ts
+++ b/src/tlv/TlvNumber.ts
@@ -12,7 +12,7 @@ import { BitmapSchema, BitSchema, TypeFromBitSchema } from "../schema/BitmapSche
 
 /**
  * Schema to encode an unsigned integer in TLV.
- * 
+ *
  * @see {@link MatterCoreSpecificationV1_0} § A.11.1
  */
 export class TlvNumericSchema<T extends bigint | number> extends TlvSchema<T> {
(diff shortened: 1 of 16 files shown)


Apollon77:
Ok, Logic enhanced, tests too ... Now we also check that (mainly for encoding) the datatypes are correct.
@mfucci @turon happy reviewing

src/tlv/TlvObject.ts
```diff
@@ -80,13 +80,14 @@ export class ObjectSchema<F extends TlvFields> extends TlvSchema<TypeFromFields<
             const { field, name } = fieldName;
             result[name] = field.schema.decodeTlvInternalValue(reader, elementTypeLength);
         }
-        this.validate(result);
         return result as TypeFromFields<F>;
     }
 
     override validate(value: TypeFromFields<F>): void {
         for (const name in this.fieldDefinitions) {
+            if (this.fieldDefinitions[name].optional && (value as any)[name] === undefined) continue;
```
mfucci, line 88: nit: you can extract this instead of repeating it.
const fieldValue = (value as any)[name];
const { optional, schema } = this.fieldDefinitions[name];

src/tlv/TlvString.ts
```diff
@@ -42,9 +42,10 @@ export class StringSchema<T extends TlvType.ByteString | TlvType.Utf8String> ext
         return reader.readPrimitive(typeLength) as TlvToPrimitive[T];
     }
 
-    override validate({ length }: TlvToPrimitive[T]): void {
-        if (length > this.maxLength) throw new Error(`String is too long: ${length}, max ${this.maxLength}.`);
-        if (length < this.minLength) throw new Error(`String is too short: ${length}, min ${this.minLength}.`);
+    override validate(value: TlvToPrimitive[T]): void {
+        if (typeof value !== "string" && !(value instanceof Uint8Array) && !Buffer.isBuffer(value)) throw new Error(`Expected string, got ${typeof value}.`);
```
mfucci, line 46: This is incorrect since the value type is linked to this.type:

if (this.type === TlvType.Utf8String && typeof value !== "string") throw ...
if (this.type === TlvType.ByteString && !(value instanceof ByteArray)) throw ...

src/tlv/TlvNumber.ts
```diff
@@ -34,12 +34,11 @@ export class TlvNumericSchema<T extends bigint | number> extends TlvSchema<T> {
     override decodeTlvInternalValue(reader: TlvReader, typeLength: TlvTypeLength) {
         if (typeLength.type !== this.type) throw new Error(`Unexpected type ${typeLength.type}, was expecting ${this.type}.`);
         const value = reader.readPrimitive(typeLength) as T;
-        this.validate(value);
         return value;
     }
 
     override validate(value: T): void {
-        super.validate(value);
+        if (typeof value !== "number" && typeof value !== 'bigint') throw new Error(`Expected number, got ${typeof value}.`);
```
mfucci, line 41: Also validate in TlvNumberSchema to check that it is a number.

mfucci (changes requested):
You should also add validate() for TlvAny (check if TlvStream instance) TlvVoid (always throws).

src/tlv/TlvString.ts
Apollon77: Thank you , makes sense

src/tlv/TlvNumber.ts
Apollon77: Oh, I missed that second class in the file, added

Apollon77:
@mfucci All feedback is addressed

src/tlv/TlvNumber.ts
```diff
@@ -78,6 +77,12 @@ export class TlvNumberSchema extends TlvNumericSchema<number> {
             minValue(max, this.max),
         );
     }
+
+    override validate(value: number): void {
+        if (typeof value !== "number") throw new Error(`Expected number, got ${typeof value}.`);
+        if (this.min !== undefined && value < this.min) throw new Error(`Invalid value: ${value} is below the minimum, ${this.min}.`);
```
mfucci, line 83: Extract this into a validateBoundary method part of TlvNumericSchema to avoid repeating this code.

src/tlv/TlvString.ts
```diff
@@ -42,9 +42,11 @@ export class StringSchema<T extends TlvType.ByteString | TlvType.Utf8String> ext
         return reader.readPrimitive(typeLength) as TlvToPrimitive[T];
     }
 
-    override validate({ length }: TlvToPrimitive[T]): void {
-        if (length > this.maxLength) throw new Error(`String is too long: ${length}, max ${this.maxLength}.`);
-        if (length < this.minLength) throw new Error(`String is too short: ${length}, min ${this.minLength}.`);
+    override validate(value: TlvToPrimitive[T]): void {
+        if (this.type === TlvType.Utf8String && typeof value !== "string") throw new Error(`Expected string, got ${typeof value}.`);
+        if (this.type === TlvType.ByteString && !(value instanceof Uint8Array)) throw new Error(`Expected Uint8Array, got ${typeof value}.`);
```
mfucci, line 47: instanceof ByteArray ?

ByteArray is for now an alias for Uint8Array but if this is changed, it should not break this code.

mfucci (approved):

src/tlv/TlvNumber.ts
Apollon77: makes sense

Apollon77:
@mfucci @turon Comments adjusted, can be merged and ideally (with the other PR) new version released. Thank you

PR #50, opened 2023-02-07
Add fallbacks for mandatory fields

mfucci:
chip-tools doesn't populate some mandatory fields in some cases.

See: https://github.com/mfucci/node-matter/pull/222

### DIFF
--- a/src/tlv/TlvObject.ts
+++ b/src/tlv/TlvObject.ts
@@ -107,7 +107,11 @@ export const TlvObject = <F extends TlvFields>(fields: F) => new ObjectSchema(fi
 /** List TLV schema. */
 export const TlvList = <F extends TlvFields>(fields: F) => new ObjectSchema(fields, TlvType.List);
 
-/** Object TLV mandatory field. A fallback value can be defined for backward compatibility.  */
+/** 
+ * Object TLV mandatory field. Optionally provide a fallback value to initialize the field value when devices omit
+ * providing a value against the specifications or in special usecases. Make sure to use a value that is an equivalent
+ * to the value being empty.
+ */
 export const TlvField = <T>(id: number, schema: TlvSchema<T>, fallback?: T) => ({ id, schema, fallback, optional: false }) as FieldType<T>;
 
 /** Object TLV optional field. */
--- a/src/tlv/TlvObject.ts
+++ b/src/tlv/TlvObject.ts
@@ -14,6 +14,7 @@ export interface FieldType<T> {
     id: number,
     schema: TlvSchema<T>,
     optional?: boolean,
+    fallback?: T,
 };
 
 export interface OptionalFieldType<T> extends FieldType<T> {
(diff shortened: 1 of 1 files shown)


Apollon77 (approved):

src/tlv/TlvObject.ts
```diff
@@ -97,8 +107,8 @@ export const TlvObject = <F extends TlvFields>(fields: F) => new ObjectSchema(fi
 /** List TLV schema. */
 export const TlvList = <F extends TlvFields>(fields: F) => new ObjectSchema(fields, TlvType.List);
 
-/** Object TLV mandatory field. */
-export const TlvField = <T>(id: number, schema: TlvSchema<T>) => ({ id, schema, optional: false }) as FieldType<T>;
+/** Object TLV mandatory field. A fallback value can be defined for backward compatibility.  */
```
turon, line 110: Isn't `defaultValue` a better, more self-explanatory term than `fallback`?

src/tlv/TlvObject.ts
```diff
@@ -80,6 +81,15 @@ export class ObjectSchema<F extends TlvFields> extends TlvSchema<TypeFromFields<
             const { field, name } = fieldName;
             result[name] = field.schema.decodeTlvInternalValue(reader, elementTypeLength);
         }
+        // Check mandatory fields and, if missing, populate with fallback value if defined.
+        for (const name in this.fieldDefinitions) {
```
turon, line 85: I'm confused why this logic is under OptionalFieldType as opposed to FieldType, especially given the comment that it only applies to mandatory fields.

src/tlv/TlvObject.ts
Apollon77: In this case it is formally not a default value "as per specs" (and we use this term "default" already for Default values of attributes in the clusters). Sp from my POV "fallback" is fine because it is the "if the value was discarded while sending it is easier for our code to use this fallback value" :-)

src/tlv/TlvObject.ts
Apollon77: Are you sure that not GitHub tricked you? ;-) It is in the "ObjectsSchema" which is generic

turon: oh right, my bad.  thanks!

src/tlv/TlvObject.ts
turon: Would it be possible to expand the comment to explain better to a layperson how it is used by the API to inform the caller how to set it with intention?

Apollon77, line 110: This would be my (non native speaking) try :-)

```suggestion
/** 
  * Object TLV mandatory field. Optionally provide a fallback value to initialize the field value when devices omit
  * providing a value against the specifications or in special usecases. Make sure to use a value that is an equivalent
  * to the value being empty.
  */
```

Apollon77: my try see https://github.com/project-chip/matter.js/pull/50/files#r1103324471

turon (approved):

Apollon77:
Cool, Then I would be so free and merge :-)

Apollon77:
@mfucci I also fixed the merge conflict after merging my PR, so please do final check and merge. Then a new release on npm would be awesome

PR #18, opened 2022-11-13
Add Tlv nullable

mfucci:

PR #52, opened 2023-02-19
Migrate matter.js to monorepo

Apollon77:
This is my counter-approach to #51 

it is basically "just" using the workspaces feature of npm and project references of typescript

"nom install" is only called on root level and not anymore in the packges themself. This is important. All rest should work automagically.

What does this PR do:
1.) It creates the packages directory and moves "matter.js" into it; a second dummy package should represend node-matter.content (as Marco did too)
2.) Then tsconfig and build stuff is adjusted to build in this environment
3.) build/test command sin root package.json will execute all sub scripts

Thats basically it


tsconfig.json
```diff
@@ -1,9 +1,15 @@
-/* Dev configuration */
 {
-  "extends": "./tsconfig.base.json",
-  "compilerOptions": {
-    "esModuleInterop": true, // Required fto remove warnings or jest test
-  },
-  /* During dev, src are included so they are also modified during refactorings. */
-  "include": ["src/**/*.ts", "test/**/*.ts"],
+    "compilerOptions": {
```
AlCalzone, line 2: I'd recommend extending from a `@tsconfig/*` package, e.g. https://github.com/tsconfig/bases/#node-lts-tsconfigjson and only add necessary changes.

tsconfig.json
```diff
@@ -1,9 +1,15 @@
...
+        "incremental": true,
```
AlCalzone, line 3: composite implies incremental

tsconfig.build.json
```diff
@@ -0,0 +1,18 @@
+{
```
AlCalzone, line 1: It is a bit cumbersome to manage references everywhere, but I'd recommend basing the `build` config on the non-build config and only make it actually build. E.g. like this.

```jsonc
{
	"extends": "./tsconfig.json",
	// project references here.
	"compilerOptions": {
		"rootDir": "src",
		"outDir": "build"
	},
	"include": ["src/**/*.ts"],
	"exclude": [ /* test files, etc */ ]
}
```

AlCalzone: Upside of this, the editor experience is using the same settings, so no surprises at build time

tsconfig.build.json
```diff
@@ -0,0 +1,18 @@
...
+        "noImplicitOverride": true,
+        "noUnusedParameters": true,
+        "noUnusedLocals": true,
+        "strictNullChecks": true,
+        "strict": true
+    },
+    "references": [
+        { "path": "packages/matter.js" },
+        { "path": "packages/node-matter" }
+    ]
```
AlCalzone, line 17: If there's a clear "entrypoint" to the monorepo, one consideration would be to have no references at the top level, and build in the "entrypoint" package instead.

packages/matter.js/package.json
```diff
@@ -0,0 +1,48 @@
+{
```
mfucci, line 1: Can you tell git that this file is from the original package.json and not a new file?
There is a command for that, but I don't remember it.
This will help a lot to see what is different here.

packages/node-matter/package.json
```diff
@@ -0,0 +1,42 @@
...
+        "clean": "rm -rf dist",
+        "build": "npm run clean; tsc",
+        "test": "TZ=utc true || mocha"
+    },
+    "devDependencies": {
+        "ts-node": "^10.9.1",
+        "typescript": "^4.9.5"
+    },
+    "dependencies": {
+        "@project-chip/matter.js": "^0.1.0"
```
mfucci, line 35: How does typescript knows to link against the local matter.js build vs the published one?

tsconfig.json
```diff
@@ -1,9 +1,15 @@
...
+        "target": "es2019",
+        "module": "commonjs",
```
mfucci, line 5: If this file is the base file for matter.js and node-matter, it should contain only the common parameters: it is more difficult to follow if parameters should be overridden.

tsconfig.json
```diff
@@ -1,9 +1,15 @@
...
+        "declaration": true,
+        "sourceMap": true,
+        "composite": true,
+        "strict": true,
+        "moduleResolution": "node",
+        "esModuleInterop": true,
```
mfucci, line 11: This is not be able be default: this are only very specific cases why this is needed that should clearly be listed here. Otherwise, this will create less optimal code...

tsconfig.build.json
```diff
@@ -0,0 +1,18 @@
...
+        "noImplicitAny": true,
+        "noImplicitOverride": true,
+        "noUnusedParameters": true,
+        "noUnusedLocals": true,
+        "strictNullChecks": true,
+        "strict": true
+    },
+    "references": [
+        { "path": "packages/matter.js" },
+        { "path": "packages/node-matter" }
```
mfucci, line 16: packages/node ?

packages/node-matter/package.json
AlCalzone: ~~magic~~ [npm workspaces](https://docs.npmjs.com/cli/v9/using-npm/workspaces)

TL;DR: When a dependency matches the package name of a package referenced by the top-level `package.json` `workspaces`, it automatically gets symlinked instead of installed from elsewhere.

TypeScript just works, it doesn't care where the package comes from btw.

tsconfig.json
AlCalzone: Interestingly, this is in the `recommended` base tsconfig:
https://github.com/tsconfig/bases/blob/main/bases/recommended.json

FWIW, the TypeScript team even recommends it in the [release notes](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-7.html#support-for-import-d-from-cjs-from-commonjs-modules-with---esmoduleinterop) and imply the recommendation in [the docs](https://www.typescriptlang.org/tsconfig#esModuleInterop):
> Turning on esModuleInterop will fix both of these problems in the code transpiled by TypeScript.

So if anything, this should at least be the default for Node.js

packages/node-matter/package.json
AlCalzone: If you want to try yourself:
```jsonc
// file: package.json
{
	"name": "monorepo",
	"private": true,
	"workspaces": [
		"packages/*"
	]
}
```

```jsonc
// file: packages/a/package.json
{
	"name": "package-a",
	"version": "1.0.0"
}
```

```jsonc
// file: packages/b/package.json
{
	"name": "package-b",
	"version": "1.0.0",
	"dependencies": {
		"package-a": "1.0.0"
	}
}
```

run `npm install` and see how the folders in `node_modules` are actually symlinks to the `packages` subdirectories.

packages/matter.js/package.json
Apollon77: This is the same problem we already had in an other PR: When you move a file AND change it in the same PR then Git will not see it as a "Move" anymore somehow. I did not found a way to prevent that last ime and this time we can not bypass t by using two PRs, lets please accept that behaviour of git/github

tsconfig.json
Apollon77: Yes I will clean this up

tsconfig.build.json
Apollon77: This is the naming discussion we need to do and ideally the directly is names as the package in the end ... The question is how the "new" node-matter package is named ...

Apollon77: @turon @mfucci Any ideas? I think Marco had @project-hip/matter-node.js in this PR, we could also leave node-matter ... This is then for the "final" PR

packages/matter.js/package.json
Apollon77: PS: Any yes I remember that you proofed me in another PR that it works if you do that ... but I still did not found out why it is not working when I do that ... no clue and honestly I would love to not spend hours in finding the reason. Maybe my git is too old with 2.19.0 ... need to see how to best update, so will retry in the finalPR ... this is for me just the "proof of concept"

tsconfig.json
Apollon77: We could go with @tsconfig/node-lts or with @tsconfig/node16 ... @mfucci @turon What do you think?

tsconfig.build.json
Apollon77: Because of the fact that we really publish two libraryies and none is the real "entrypoint" I would keep this

packages/node-matter/package.json
Apollon77: Exxactly. This is the magic of the workspaces.

Apollon77:
Ok, I updated and tried to cover the comments best possible ... I especially collected the tsconfigs in the base file ... for the open comments @mfucci @turon need your opinions or ok or whatever.

In my eyes this works and we can optimize configs and such also later.

packages/matter.js/package.json
Apollon77: PS: I assume in this case it is even more special because the "original place" stays also with a package.json with changed content, so it is "moved file and created new in the same spot". maybe would had been registered by git if it would had been two commits? I would not dig into that deeper because feels lost time

packages/node-matter/package.json
Apollon77: Bildschirm­foto 2023-02-24 um 17 58 18

Here it how it looks in the FS in exactly our case

tsconfig.build.json
Apollon77: Ideally the directories are named like the packages that are inside the directories ... So How the naming should be?

* @project-chip/matter.js we already have

Options for the node package in my eyes
* @project-chip/matter.js-node
* @project-chip/matter-node.js
* @project-chip/matter-nodejs
* ...

Apollon77: Ok we decided for @project-chip/matter-node.js

packages/matter.js/package.json
turon: I've run into the same problem when an existing file is moved away and replaced, git diff just can't figure out how to show the simplest capture of the changes.

.gitignore
```diff
@@ -1,4 +1 @@
-/dist
-/node_modules
-/out
```
turon, line 3: How does git know to ignore `/out`?

README.md
```diff
@@ -1,9 +1,22 @@
-# matter.js
+# JavaScript/TypeScript based Matter Implementation
 
 ![experimental](https://img.shields.io/badge/status-Experimental-red) [![license](https://img.shields.io/badge/license-Apache2-green.svg)](https://raw.githubusercontent.com/project-chip/matter.js/master/LICENSE)
 
 Implementation of Matter protocol in typescript with no native dependencies (and very limited dependencies).
 
+## Monorepo Overview
+
+This repository contains multiple packages (and because of this it is a monorepo). The packages are contained in the `packages` directory and arer all published separately to NPM.
```
turon, line 9:
```suggestion
This repository contains multiple packages (and because of this it is a monorepo). The packages are contained in the `packages` directory and are all published separately to NPM.
```

tsconfig.json
turon: I seem to recall `matternode`, which is mostly `js` as opposed to `ts`, required `esModuleInterop: true` as well to link to `matter.js` properly.

README.md
```diff
@@ -1,9 +1,22 @@
...
+
+* matter.js: the core Matter implementation in typescript which is JavaScript only and has no native dependencies.
+* matter-node.js: a node.js implementation of a Matter
+
+This repository uses the workspaces feature on npm to manage the dependencies between the packages. Because of this please only use `npm install` on the root of the repository. This will install all the dependencies for all the packages and also take care to create relevant symlinks between the packages.
```
turon, line 14: RE: `only use`, it looks like `npm run build` and `npm test` will work fine from the top-level as well, and simply build or test all workspaces, no?

turon (approved):
LGTM :+1:

.gitignore
Apollon77: out should not exist on root level anymore becau it only exists in /packages/matter.js/out and it is ignored there

README.md
Apollon77: yes works too, see in line #18 below.This section is about "npm install" to NOT run npm install in one of the "/packages/* directories

mfucci (approved):

PR #53, opened 2023-02-25
[codegen] Initial version of Cluster code generation.

turon:
Initial version of using Matter SDK code generation to create the matter.js cluster definitions.
This doesn't generate perfect output yet, but provides an initial framework for automatic generation of the cluster handlers.

### DIFF
--- a/tools/codegen/matter_idl_plugin/__init__.py
+++ b/tools/codegen/matter_idl_plugin/__init__.py
@@ -1,0 +1,187 @@
+"""Plugin module to matter_idl for auto-generation of .proto files.
+
+This module provides a plugin to the matter_idl tool from the Matter SDK
+(https://github.com/project-chip/connectedhomeip/tree/master/scripts/py_matter_idl)
+to assist auto-generating matter.js TLV schema definitions from the
+Matter SDK XML schema.
+"""
+
+import os
+from typing import List
+
+import jinja2
+from matter_idl.generators import CodeGenerator
+from matter_idl.generators import GeneratorStorage
+from matter_idl.matter_idl_types import Attribute
+from matter_idl.matter_idl_types import Cluster
+from matter_idl.matter_idl_types import ClusterSide
+from matter_idl.matter_idl_types import Command
+from matter_idl.matter_idl_types import Field
+from matter_idl.matter_idl_types import Idl
+
+
+def toUpperCamelCase(s: str) -> str:
+  """ Converts a lowerCamelCase string to UpperCamelCase."""
+  return s[0].upper() + s[1:]
+
+def toLowerCamelCase(s: str) -> str:
+  """ Lower the first character of the given string."""
+  return s[0].lower() + s[1:]
+
+def toTlvSchemaType(zap_type: str) -> str:
+  """Convert zap type to protobuf type."""
+  u8_types = [
+      "int8u",
+      "enum8",
+      "bitmap8",
+  ]
+
+  u16_types = [
+      "int16u",
+      "enum16",
+      "bitmap16",
+      "group_id",
+  ]
+
+  u32_types = [
+      "int32u",
+      "int24u",
+      "uint32",
+      "enum32",
+      "bitmap32",
+      "cluster_id",
+      "attrib_id",
+      "event_id",
+      "command_id",
+      "endpoint_no",
+      "devtype_id",
+      "fabric_idx",
+      "vendor_id",
+      "status_code",
+      "faulttype",
+      "levelcontroloptions",
+      "percent100ths",
+      "percent",
+      "epoch_s",
+  ]
+  u64_types = [
+      "uint64",
+      "enum64",
+      "bitmap64",
+      "node_id",
+      "fabric_id",
+      "int40u",
+      "int48u",
+      "int56u",
+      "int64u",
+      "epoch_us",
+  ]
+  i8_types = ["int8s"]
+  i16_types = ["int16s"]
+  i32_types = ["int32", "int24s", "int32s"]
+  i64_types = ["int64", "int40s", "int48s", "int56s", "int64s"]
+  float_types = ["float", "single"]
+  string_types = ["char_string", "long_char_string"]
+  bytes_types = ["octet_string", "long_octet_string"]
+
+  zap_type_lower = zap_type.lower()
+  if zap_type_lower in u8_types:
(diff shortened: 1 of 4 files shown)


Apollon77:
@turon Thats cool. It would - in any case - simplify the initial creation of the files which is basically 60-70% of the work. The rest is then specs reading and potential adjustments ... or do we trust the XML 1000%? I found several things in the XMLs that do not make sense sometimes ... if they are the truth then  the specs are wrong in several places :-)

Maybe let the two of us have a call next week to discuss the goal and next steps. In anycase thtis is cool. I will review the generation template later today!

tools/codegen/matter_idl_plugin/__init__.py
```diff
@@ -0,0 +1,187 @@
...
+  if zap_type_lower in float_types:
+    return "TlvFloat"
+  if zap_type_lower == "double":
+    return "TlvDouble"
+  if zap_type_lower == "boolean":
+    return "TlvBoolean"
+  if zap_type_lower in string_types:
+    return "TlvString"
+  if zap_type_lower in bytes_types:
+    return "TlvBbytes"
```
Apollon77, line 113: Should be TlvByteString I think

turon:
There is a bug in the SDK that caches an older version of the matter_idl
lib in the pigweed .environment.  i should add a note to the README to
delete that to make it work.

tools/codegen/matter_idl_plugin/matter_js_schema.jinja
```diff
@@ -0,0 +1,79 @@
...
+export const enum {{entry.name}} {
+    {%-  for field in entry.entries %}
+    {{field.name}} = {{field.code}},{%   endfor %}
+}
+{%   endfor %}{% endif -%}
+{%- if cluster.bitmaps %}
+
+  // Bitmaps
+{%-   for entry in cluster.bitmaps %}
+export const enum {{entry.name}} {
```
Apollon77, line 29: BitMaps should look like

```
const TlvNameSupportBitmap = TlvBitmap(TlvUInt8, {
    groupNames: BitFlag(7),
});
```

tools/codegen/matter_idl_plugin/matter_js_schema.jinja
```diff
@@ -0,0 +1,79 @@
...
+import { TlvEnum, TlvUInt16, TlvUInt64 } from "../tlv/TlvNumber.js";
+import { TlvString } from "../tlv/TlvString.js";
+import { TypeFromSchema } from "../tlv/TlvSchema.js";
+import { TlvBoolean } from "../tlv/TlvBoolean.js";
+
+{%- if cluster.enums %}
+
+  // Enums
+{%-   for entry in cluster.enums %}
+export const enum {{entry.name}} {
```
Apollon77, line 20: should we export by default?

tools/codegen/matter_idl_plugin/matter_js_schema.jinja
```diff
@@ -0,0 +1,79 @@
...
+    {{field.name}} = {{field.code}},{%   endfor %}
+}
+{%   endfor %}{% endif -%}
+{%- if cluster.structs %}
+
+  // Structs
+{%-   for entry in cluster.structs %}{% if not entry.tag %}
+const Tlv{{entry.name}} = TlvObject({
+    {%-  for field in entry.fields %}
+    {{field.name}}:  TlvField({{field.code}}, {{field | toTlvSchemaFullType}}),{%   endfor %}
```
Apollon77, line 40: could we also get min/max and the others in here? Or if not programmatically maybe as comment?

tools/codegen/matter_idl_plugin/matter_js_schema.jinja
```diff
@@ -0,0 +1,79 @@
...
+ */
+export const {{cluster.name}}Cluster = Cluster({
+    id: {{cluster.code}},
+    name: "{{cluster.name}}",
+    revision: 1,
+{%- if cluster.attributes %}
+
+    attributes: {
+{%-   for attr in cluster.attributes %}
+        {{attr.definition.name}}: Attribute({{attr.definition.code}}),  {{attr.definition.data_type.name | toTlvSchemaType}} }),
```
Apollon77, line 69: Same here for optional and writable ...

tools/codegen/matter_idl_plugin/matter_js_schema.jinja
```diff
@@ -0,0 +1,79 @@
...
+{% endfor %}
+  }{% endif %}
+{%- if cluster.commands %}
+
+    commands: {
+{%-   for cmd in cluster.commands %}
+        {{cmd.name | toLowerCamelCase}}: Command({{cmd.code}}, Tlv{{cmd.name}}Request, {{cmd.code}}, Tlv{{cmd.name}}Response),
+{% endfor %}
```
Apollon77, line 77: could you also add the events?

Apollon77 (commented):
I will just leave my thoughts as comments because question is about the goal. this is a cool basis, but also a lot is not yet implemented which needs manual work afterwards in any case ... but would save time already

turon:
> or do we trust the XML 1000%? I found several things in the XMLs that do not make sense sometimes ... if they are the truth then the specs are wrong in several places :-)

The XML is somewhat the truth-on-the-ground because the Matter SDK is what is used to build and certify final products and it uses the XML codegen extensively for both client and server code.  Any spec bugs should be captured as an issue and we can correct it (most likely by correcting the spec to match the 1.0 SDK behavior at this point).  The XML lacks certain things where the manually crafted definitions will remain better such as `@see` spec references and documentation.

Apollon77:
Ok, then I can go through the XMLs and specs of the clusters we have and will open issues for all things I see differences ...

Apollon77:
based on latest learnings this is true, BUT the following things are nmot covereed by the XMLs:
* default values have a different meaning as in specs
* Non-volatile flag is not in XML

tools/codegen/matter_idl_plugin/matter_js_schema.jinja
turon: Looks like enums in hand-written clusters are all exported, so yes.

Apollon77:
@turon I will close here in favor of #192 . Thank you for your collaboration!

PR #55, opened 2023-03-03
Add eslint and formatter

Apollon77:
The first commit adds eslint with a config needed based on the current matter.js code and some fixes and adjustments based on findings.

The second commit adds Formatter including all relevant changes after running it

The third commit adds linter and formatter to github action and readme


.eslintrc.json
```diff
@@ -0,0 +1,76 @@
...
+        "plugin:import/errors",
+        "plugin:import/warnings",
+        "plugin:import/typescript"
+    ],
+    "plugins": ["@typescript-eslint", "import"],
+    "reportUnusedDisableDirectives": true,
+    "env": {
+        "node": true
+    },
+    "ignorePatterns": ["**/*.d.ts", "build/**/*"],
```
mfucci, line 21: Why would we ignore .d.ts files?
We are using some...

.eslintrc.json
```diff
@@ -0,0 +1,76 @@
...
+
+        "no-constant-condition":[
+            "error",
+            {
+                "checkLoops": false
+            }
+        ]
+    },
+    "settings": {
+        "import/extensions": [".ts", ".tsx"],
```
mfucci, line 63: You can remove .tsx since we are not using react...
Same below

.eslintrc.json
```diff
@@ -0,0 +1,76 @@
...
+        "import/parsers": {
+            "@typescript-eslint/parser": [".ts", ".tsx"]
+        },
+        "import/resolver": {
+            "typescript": {
+                "alwaysTryTypes": true, // always try to resolve types under `<root>@types` directory even it doesn't contain any source code, like `@types/unist`
```
mfucci, line 69: Do we really need this?
I am unclear what this does from the comment...

Apollon77:
Which import you mean?

mfucci:
All the ts imports.
I found some instructions here: https://medium.com/weekly-webtips/how-to-sort-imports-like-a-pro-in-typescript-4ee8afd7258a
But that should probably be another MR

Apollon77:
If we really need that ... it might be (again) my pragmatic view, but normally my ide handles imports ... I personally never read them that much 2which means I also do not care too much about their order ;-))

But no issue with adding it if it is taken care automatically as part of the formatting

.eslintrc.json
Apollon77: right

.eslintrc.json
Apollon77: adjusted, right ...

Apollon77:
Migrated to prettier, added import organization plugin, adressed review comments

PR #61, opened 2023-03-17
Controller/InteractionClient Process optimizations and refactoring

Apollon77:
This PR is reinstanciating https://github.com/mfucci/node-matter/pull/257

Most of the PR was discussed and left was that there is a good chance theat SubscriptionClient is added multiple times to the Exchange. This is right and this issue existed already nbefore this PR!

A change would need a deeper restructuring of how subscriptions are handled. In my eyes the subscription manager would somehow need to be moved to the MessageExchange. I would like to NOT blow up this PR with also adding this to it but rather move it to a follow up PR (and most likely the whole topic will change anyway with the Hight level API again).

Original PR content:
This PR collects some things that I could not split better. Ithink it might be easier to review the single Commits because they for themself are simpler.

In fact the PR does:
* Cleanup in ChannelManager (especially centrallizing the Channelkey generation. It also adds "removeChannel" which is needed when InteractionCLient is reconnecting
* Create and throw own typed error for Retransmission Reached case to allow more easy catching
* Make timeout to discover a device configurable - during commissioning we usually wait 5s, for resuming a session (preparation for persistence!) we should allow to wait longer, so I made the timeout configurable
* Add session-Reconnect logic to "withMessanger" for Interactionclient to renew session - If we have a running connection to a device in controller case and the device gets restarted then the current session becomes invalid. Apple/Google/Amazon are opening a new session when this happens so we simply do the same. The "withController" method which is used gets enhances by a method that also can handle a reconnect if the Retransmission-Reached Error is thrown for the send command.
* The interactionClient gets an optional "reconnect" callback for this case to have the logic itself in MatterController where the rest of the connection login is owned. I think thisan elegant way to keep the topics selarated - if you have a better one just tell.
* The InteractionClient now also gets a method to "resume" a session and because we do something like that also when doing CASE pairing I refactored to code that the "connect" method takes care of that. It will all be used later when persistence is there - then the code can know if we have are commissioned or not and depending on this enter commissioning or just do a session resume/reconnect. So this is here preparation


turon:
`In my eyes the subscription manager would somehow need to be moved to the MessageExchange.`

This sounds like a bad mixing of layers.  Exchanges are message layer while subscriptions are part of the interaction model.  They should not be mixed.

Apollon77:
Ok. Also true. But as said I would tackle this in a separate PR. The comment was mainly about how to fix the issue we have left open (about potentially duplicating Subscriptionmsnsger in interactionclient)

Apollon77:
In my eye sthe final High level API makes sure that there is only one interactionClient for a node at a certain timepoint which makes the issue less likely - and in this changes then I would also try to find a solution for that lft over topic  and leave the Todo in for now.

WDYT? @turon

packages/matter-node.js/src/matter/mdns/MdnsScanner.ts
```diff
@@ -38,15 +38,15 @@ export class MdnsScanner implements Scanner {
...
+    async findDevice({ operationalId }: Fabric, nodeId: NodeId, timeoutSeconds = 5): Promise<MatterServer | undefined> {
         const operationalIdString = operationalId.toHex().toUpperCase();
         const deviceMatterQname = getDeviceMatterQname(operationalIdString, nodeId.toString());
 
         const record = this.matterDeviceRecords.get(deviceMatterQname);
         if (record !== undefined) return { ip: record.ip, port: record.port };
 
         const { promise, resolver } = await getPromiseResolver<MatterServer | undefined>();
-        const timer = Time.getTimer(5 * 1000 /* 5 s*/, () => {
+        const timer = Time.getTimer(timeoutSeconds * 1000 /* timeOut seconds */, () => {
```
turon, line 49: Not sure what value the `/* timeOut seconds */` comment is adding.

turon: Also, if the underlying resolution of the timer is milliseconds, would it make sense to simply expose `timeoutMs` instead of `timeoutSeconds` and multiplying out?

turon (approved):
:+1: LGTM

packages/matter-node.js/test/IntegrationTest.ts
```diff
@@ -149,30 +149,30 @@ describe("Integration", () => {
             assert.equal(nodeId.id, BigInt(1));
         }, 60 * 1000 /* 1mn timeout */);
 
-        it("the session is resumed if it has been established previously", () => {
-            client.connect(new NodeId(BigInt(1)));
+        it("the session is resumed if it has been established previously", async () => {
+            await client.connect(new NodeId(BigInt(1)));
```
turon, line 153: Perhaps replace `BigInt(1)` with `1n`?

Apollon77: From my knowledge full Bigint support came in ES2020 ... but our typescript setting generates ES2015 ... with "BigInt()" there are no issues ... just the "1n" notation

Apollon77: (and we randomize the ID with one of the next PRs anyway" :-)

packages/matter-node.js/src/matter/mdns/MdnsScanner.ts
Apollon77: I removed the unneeded comment. The matter specs mostly use seconds so lets use seconds

turon: Yeah, align with spec is always good justification.

PR #63, opened 2023-03-19
Use persistence

Apollon77:
This PR adds usage of persistence for Controller and Device. Additionally it:
* Splits out RootCertificateManager to an own file in certificate directory
* Adds a way for Device to detect if the device was commissioned or is a new device
* Allows to specify the vendorid, productid, passcode, discriminator and storage file to use via parameters

Important-Infos:
* The tests will fail until #61 is merged which brings the persistence flag for attributes which is used and also checked in tests (and also because it already uses some moved types from that PR). I still decided to bring in that way because so review can already happen in parallel
* When a device got paired and is then restarted and reinitialized from persistence the announcement is done wrong because it will announce a "commissionable device" instead of "the already commissioned" device. I decided to not lso add all this to this PR, but will prepare another PR that will adjust commissioning

This all should not be issues to starta review and even approve when ok. I take care about the merge order


packages/matter-node.js/src/Device.ts
```diff
@@ -43,22 +43,29 @@ import { AdminCommissioningHandler } from "./matter/cluster/server/AdminCommissi
 import { NetworkCommissioningHandler } from "./matter/cluster/server/NetworkCommissioningServer";
 import { AttestationCertificateManager } from "./matter/certificate/AttestationCertificateManager";
 import { CertificationDeclarationManager } from "./matter/certificate/CertificationDeclarationManager";
+import { StorageNode } from "./persistence/StorageNode";
```
turon, line 46: Have you considered using the `node-localstorage` package as `matternode` did?  It is a ready-to-go package, and every field is a simple text file in storage (so nicely independent from async issues, and easy to view / patch by the user out-of-band).

https://github.com/project-chip/matternode/blob/main/src/platform/PersistantStore.js#L35

It would be good to make sure a platform abstraction is part of the design or at least there is a plan for how such a PAL could be added later.

packages/matter-node.js/src/Device.ts
```diff
@@ -163,28 +170,35 @@ class Device {
...
+            console.log(`Manual pairing code: ${ManualPairingCodeCodec.encode({ discriminator, passcode })}`);
+        } else {
+            console.log("Device is already commissioned. Waiting for controllers to connect ...");
+        }
     }
 }
 
 new Device().start().catch(error => logger.error(error));
+
+process.on("SIGINT", () => {
```
turon, line 202: It would be nice to hide this outside of the app.

packages/matter-node.js/src/Controller.ts
```diff
@@ -22,25 +22,41 @@ import { UdpInterface } from "./net/UdpInterface";
...
+        const persistenceManager = new PersistenceManager(storage);
+        await persistenceManager.initialize();
+
         const ip = getParameter("ip");
         if (ip === undefined) throw new Error("Please specify the IP of the device to commission with -ip");
         const port = getIntParameter("port") ?? 5540;
         const discriminator = getIntParameter("discriminator") ?? 3840;
         const setupPin = getIntParameter("pin") ?? 20202021;
-        const client = await MatterController.create(await MdnsScanner.create(), await UdpInterface.create(5540, "udp4"), await UdpInterface.create(5540, "udp6"));
+        const client = await MatterController.create(await MdnsScanner.create(), await UdpInterface.create(5540, "udp4"), await UdpInterface.create(5540, "udp6"), persistenceManager);
```
turon, line 44: Why does the network interface need the persistenceManager?

turon: Also why do we hard-code the port here?  The client port should be ephemeral and different from the default server port (which is okay to default to 5540, but should be defined as a const minimally rather than a magic number).

packages/matter-node.js/src/Device.ts
```diff
@@ -43,22 +43,29 @@ import { AdminCommissioningHandler } from "./matter/cluster/server/AdminCommissi
...
+        const persistenceManager = new PersistenceManager(storage);
+        await persistenceManager.initialize();
+
         const deviceName = "Matter test device";
         const deviceType = 257 /* Dimmable bulb */;
-        const vendorName = "node-matter";
-        const passcode = 20202021;
-        const discriminator = 3840;
+        const vendorName = "matter-node.js";
+        const passcode = getIntParameter("passcode") ?? 20202021;
```
turon, line 63: Shouldn't the sequence be:
- if passed via command line, use it
- else try to load from storage
- else use default

packages/matter-node.js/src/matter/MatterController.ts
```diff
@@ -36,32 +38,48 @@ const ADMIN_VENDOR_ID = new VendorId(752);
...
             .setIdentityProtectionKey(ipkValue)
             .setRootVendorId(ADMIN_VENDOR_ID);
         fabricBuilder.setOperationalCert(certificateManager.generateNoc(fabricBuilder.getPublicKey(), FABRIC_ID, CONTROLLER_NODE_ID));
-        const fabric = await fabricBuilder.build();
-        return new MatterController(scanner, netInterfaceIpv4, netInterfaceIpv6, certificateManager, fabric);
+
+        // Check if we have a fabric stored in the persistence, if yes initialize this one, else build a new one
+        const controllerPersistance = persistenceManager.createPersistence("MatterController");
+        if (controllerPersistance.has("fabric")) {
+            const storedFabric = Fabric.fromJson(controllerPersistance.get<FabricJsonObject>("fabric"));
```
turon, line 55: This API requires knowledge of the storage format and raw parsing of JSON.  The storage API should simply provide values from field name for you.

Ideally, this would look something like `fabric = storage.get("fabric")`.

packages/matter-node.js/src/matter/MatterController.ts
```diff
@@ -36,32 +38,48 @@ const ADMIN_VENDOR_ID = new VendorId(752);
...
         private readonly scanner: Scanner,
         private readonly netInterfaceIpv4: NetInterface,
         private readonly netInterfaceIpv6: NetInterface,
         private readonly certificateManager: RootCertificateManager,
         private readonly fabric: Fabric,
+        private readonly persistenceManager: PersistenceManager
     ) {
+        this.controllerPersistence = this.persistenceManager.createPersistence("MatterController");
+
+        this.sessionManager = new SessionManager(this, this.persistenceManager);
```
turon, line 79: Couldn't manager use `this` to ask for storageManager?

packages/matter-node.js/src/matter/certificate/RootCertificateManager.ts
```diff
@@ -0,0 +1,102 @@
+/**
```
turon, line 1: Is this existing functionality or new functionality unrelated to persistence?  It's best to keep PRs focused on one thing.

packages/matter-node.js/test/IntegrationTest.ts
```diff
@@ -212,6 +226,32 @@ describe("Integration", () => {
         });
     });
 
+    describe("storage", () => {
```
turon, line 229: Can unit tests for storage be dedicated unit test rather than mixed with Integration?

packages/matter-node.js/src/Device.ts
Apollon77: The approach from node-matter so far was "as less dependencies as possible", thats why no other packages were considered. It was decided to do a very lightweight nodejs storage as "defaule basic implementation". But in fact as soon as the Storage interface is implemented the dev later on can switch out the real Storage implementation (like Storagenode.ts now) by any other implementation.

packages/matter-node.js/src/Device.ts
Apollon77: Do you have an example how? Right now users start Device "file" and end with CTRL-C basically ... it was the only way I found (at least with the current structure) how to asure that the data are stored.

packages/matter-node.js/src/Controller.ts
Apollon77: 1.) Not the Network interface needs the persistenceManager but the "MatterController" class. thats why it is added here for the Controller builder. I think this all will be changed with teh Hight level API completely

2.) In one of my other PRs the port will be configurable. It was hardcoded until now in node-matter

packages/matter-node.js/src/Device.ts
Apollon77: Yes, it is not yet stored in persistence ... I can add that

packages/matter-node.js/src/matter/MatterController.ts
Apollon77: I agree, I decided for the "builder" approach because we already had the builder used in the process, so yes was "lowest effort".  Alternatively we need to add logic to storage on how to persist and also instanciate kind of any of the classes we have because t is not just data but the instace of a Fabric class. We coukld do that by defining a "Persistable" interface and classes implement that by providing toJson method and fromJson-static-builder-methods, the the storage can do that. WDYT?

packages/matter-node.js/src/matter/MatterController.ts
Apollon77: Yes, if we define the persistenceManager as public ...  Do we want that? ( I know JS do not care that much, but OOP wise it i smeeh in my eyes to have it public just for that, or?)

packages/matter-node.js/src/matter/certificate/RootCertificateManager.ts
Apollon77: In fact yes it is unrelated. I switched it out and then added persistence ... Would be cool to include here still because just moved code without any other change then persistence.

packages/matter-node.js/test/IntegrationTest.ts
Apollon77: In fact for me the Integration test is more an end-to-end test so I would also include verificstion that the right stuff is stored in there.

I currently provide a separate PR to more add testing for storage

packages/matter-node.js/src/matter/MatterController.ts
turon: There could be a getter for it.

packages/matter-node.js/test/IntegrationTest.ts
turon: I just like leading with functionality getting added as dedicated classes with dedicated tests.

packages/matter-node.js/src/Device.ts
turon: Yeah my concern is that there is quite a bit of new code to support this JSON format, all of which needs to have tests and docs, and will only potentially apply to the node impl, and not web or other platforms. `node-localstorage` has really easy to use and understand footprint on disk.  I'll have to play with a local copy of this PR to see what the storage files look like in practice.

packages/matter-node.js/src/matter/MatterController.ts
Apollon77: Ok, I will adjust and add a public getter

packages/matter-node.js/src/Device.ts
Apollon77: In fact it is "a big json". Does it need to be readable?

Apollon77: (But we can also use a lib, I have no issue with this)

packages/matter-node.js/test/IntegrationTest.ts
Apollon77: I agree, this was missed in the original PR that brought in Persistence ... I will add testing to this PR tomorrow

packages/matter-node.js/src/Device.ts
Apollon77: added

packages/matter-node.js/src/matter/MatterController.ts
Apollon77: The sessionmanager is using "this" as a generic Context and is used by Device and Controller, but in fact Sessionmanager do not know what it really is typing wise. I would like to leave it that way for now ...

Apollon77:
@turon @mfucci I thought a lot and it would be good I think to really have the persistence class to only accept strings as value - this means we move the responsibility to serialize single values to the "code that want to store it". This should allow to easiely use all sorts of storage backends like a JSON as we have now, or also a node-localstorage one.

If ok I can adjust that in tjis PR and also add testing to the persistence classes here. I think it would be helpful to have #70 in to split out "json de/serialization helper methods" because we still have places where we might have special dynamic datatypes to store (e.g. when it is about storing cluster attributes as example).

Makes that sense?

packages/matter-node.js/src/Device.ts
Apollon77: replaces by node-localstorage

Apollon77:
I thought about my latest changes a bit. I like the string only approach when it comes to the storage backend. But I do not like the string only when it comes to our code because we now have parseInt and toString everywhere. 

I had an idea this morning that we really define some types again on the Persistence class level. That would allow some basic types to be handled by Persistence which would simplify usage a lot ... and then we stringify there. 

I think that would be a good compromise which is still convenient for dev but simple for the storage backend.

I will adjust that tonight.

Apollon77:
Ok, adjusted, now typing and basic serialization support for "stringifiable" (and our added special types) is back on level of Persistence" class. The value is stored stringified which end up "near to before" and should be still easy to read for simple types.
happy reviewing.

PR #70 is also included and thats why closed

packages/matter-node.js/src/Device.ts
```diff
@@ -43,22 +43,36 @@ import { AdminCommissioningHandler } from "./matter/cluster/server/AdminCommissi
...
+const storage = new StorageNodeLocalstorage(getParameter("file") || "device.json");
+
 class Device {
     async start() {
         logger.info(`node-matter`);
 
+        const persistenceManager = new PersistenceManager(storage);
+        await persistenceManager.initialize();
+
+        const devicePersistence = persistenceManager.createPersistence("Device");
```
turon, line 60: I'd call this API: `deviceStorage = storageManager.createContext("device")`

packages/matter-node.js/src/persistence/StorageNodeLocalstorage.ts
```diff
@@ -0,0 +1,43 @@
...
+    async initialize() {
+        // nothing to do
+    }
+
+    async close() {
+        // nothing to do
+    }
+
+    buildStorageKey(context: string, key: string): string {
+        return context + "---" + key;
```
turon, line 29: I'd make the context separator simply `:`.

packages/matter-node.js/test/persistence/PersistenceTest.ts
```diff
@@ -0,0 +1,96 @@
+/**
+ * @license
+ * Copyright 2022-2023 Project CHIP Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { StorageInMemory } from "../../src/persistence/StorageInMemory";
+import { Persistence } from "../../src/persistence/Persistence";
+import * as assert from "assert";
+import { SupportedStorageTypes } from "../../src/persistence/JsonConverter";
```
turon, line 10: What is the JsonConverter used for now?  I thought we aren't using JSON format for storage anymore.

turon (commented):
Looks really good now.  My comments are mainly about the naming of the new storage APIs.

packages/matter-node.js/src/persistence/StorageNodeLocalstorage.ts
Apollon77: Is a colon a good idea because used in Filenames? I tried to use a char which do not make issues there and is still "unique" enough that no one use it. I also thought about "." And forbidding . In context. What about that?

packages/matter-node.js/test/persistence/PersistenceTest.ts
Apollon77: It is still used to convert the data into strings - check the node localstorage class.

packages/matter-node.js/README.md
```diff
@@ -61,6 +61,17 @@ or when starting from TS files:
...
 ```
+(Please note the "--" to separate commandline parameters between the npm run and the executed script. 
+
+The following parameters are available:
+* -passcode: the passcode to use for pairing (default: 20202021)
+* -discriminator: the discriminator to use for pairing (default: 3840)
+* -vendorid: the vendor ID as number to use for pairing (default: 65521 (0xFFF1))
+* -productid: the product ID as number to use for pairing (default: 32768 (0x8000))
+* -port: the port to listen on (default: 5540)
+* -file: the file to use for storing the pairing information (default: device.json). Delete the file or provide an alternative name to reset the device
```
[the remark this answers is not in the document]
Apollon77: The default was device.json before ... I adjusted it to be "device-node" and "contreoler-node".

I decided against a number in the name because formally the nodeid for "device usecase" is disctated by the controller that pairs, so so "node 1" will not be "node id 1". That might be misleading.
For Controller we could do that because there we will use ID 1 by default currently, but also will be randomized in next step, so would also just confuse.

packages/matter-node.js/src/Controller.ts
```diff
@@ -22,25 +22,49 @@ import { UdpInterface } from "./net/UdpInterface";
...
     async start() {
         logger.info(`node-matter`);
 
-        const ip = getParameter("ip");
-        if (ip === undefined) throw new Error("Please specify the IP of the device to commission with -ip");
-        const port = getIntParameter("port") ?? 5540;
-        const discriminator = getIntParameter("discriminator") ?? 3840;
-        const setupPin = getIntParameter("pin") ?? 20202021;
-        const client = await MatterController.create(await MdnsScanner.create(), await UdpInterface.create(5540, "udp4"), await UdpInterface.create(5540, "udp6"));
+        const persistenceManager = new PersistenceManager(storage);
```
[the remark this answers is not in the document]
Apollon77: Ok, I adjusted ... just the names with "-" do not really match to anything ... I Used StorageBackendMemory and StorageBackendDisk

packages/matter-node.js/src/Controller.ts
```diff
@@ -22,25 +22,49 @@ import { UdpInterface } from "./net/UdpInterface";
 import { getIntParameter, getParameter } from "./util/CommandLine";
 import { MdnsScanner } from "./matter/mdns/MdnsScanner";
 import { Logger } from "./log/Logger";
+import { StorageNodeLocalstorage } from "./persistence/StorageNodeLocalstorage";
```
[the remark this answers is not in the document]
Apollon77: solved via next comment and bigger renamings

packages/matter-node.js/src/Device.ts
Apollon77: adjusted

packages/matter-node.js/src/persistence/StorageNodeLocalstorage.ts
Apollon77: I adjusted based on my proposal with "." (yes the ":" should be handled by node-localstorage but .... ;-)

turon: Ok.  Fair enough.  `-` and `_` and `.` are far more standard than `:`.

packages/matter-node.js/README.md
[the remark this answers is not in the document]
turon: I'm not seeing the update to `device-node` in the code or docs.  Perhaps my pull of the latest version of this PR didn't take.  Either way, this can be a follow-on if it needs correcting.

packages/matter-node.js/src/Controller.ts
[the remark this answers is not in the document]
turon: thanks

turon (approved):
LGTM :+1:  Thanks for adding this critical feature!

packages/matter-node.js/README.md
```diff
@@ -61,6 +61,17 @@ or when starting from TS files:
...
 ```
+(Please note the "--" to separate commandline parameters between the npm run and the executed script. 
+
+The following parameters are available:
+* -passcode: the passcode to use for pairing (default: 20202021)
+* -discriminator: the discriminator to use for pairing (default: 3840)
+* -vendorid: the vendor ID as number to use for pairing (default: 65521 (0xFFF1))
+* -productid: the product ID as number to use for pairing (default: 32768 (0x8000))
+* -port: the port to listen on (default: 5540)
+* -store: the storage location (directory) to use for storing the pairing information (default: device-node). Delete the directory or provide an alternative name to reset the device
```
Apollon77, line 72: @turon here it is ... latest commit

PR #64, opened 2023-03-19
Announcements optimizations

Apollon77:
This PR brings several enhancements for announcements, like
* Allow to specify a device port to be used for the device and announcement (allows running multiple devices on the same host)
* Allow to specify an announcement network Interface to prevent iterating all interfaces (including some not relevant ones, prevents https://github.com/mfucci/node-matter/issues/184)
* Announce a device for 15 mins (in 1minute intervals), 15mins as by specs
* Announce all Fabrics when restarting a device from persistent fabrics (enhances #63)

Notes: The handling with overlapping of announcements and an opened additional commissioning window is not perfect, but that topic will be restructured soon anyway for high level API when bringing all together. It works as needed for now


vves:
@Apollon77 I take it that node-matter if fully abandoned now and this PR sorts out some of the issues you and I were working on in the previous library?

Apollon77:
@vves basicslly yes. I took over the announcement interface from your PR but left out the "still in discussion" IPv6 topic because i still do not understand why you have issues. 

That's why I left the PR in node-matter open and we continue discuss there. Or you open a new PR in matter.js (Code places are same-ish)

packages/matter-node.js/src/matter/MatterDevice.ts
```diff
@@ -60,8 +69,61 @@ export class MatterDevice {
...
+        this.startAnnouncement();
+    }
+
+    startAnnouncement() {
+        this.announcementStartedTime = Time.nowMs();
+        this.announceInterval = Time.getPeriodicTimer(DEVICE_ANNOUNCEMENT_INTERVAL, () => this.announce()).start();
+        this.announce();
+    }
+
+    announce(announceOnce = false) {
```
turon, line 86: This PR is mixing three completely different things:
1) configurable port
2) configurable filtering of network interface
3) add announce logic to device

In the future, please separate disparate functionality like this into separate PRs.  It make approving the small stuff (like configurable port) a lot easier.

packages/matter-node.js/src/matter/MatterDevice.ts
```diff
@@ -60,8 +69,61 @@ export class MatterDevice {
...
-        this.broadcasters.forEach(broadcaster => broadcaster.announce());
+    async start() {
+        /*for (const broadcaster of this.broadcasters) {
+            broadcaster.setCommissionMode(0, this.deviceName, this.deviceType, this.vendorId, this.productId, this.discriminator);
+            broadcaster.announce();
+        }*/
+        this.startAnnouncement();
+    }
+
+    startAnnouncement() {
```
turon, line 80: Shouldn't this announce logic live in the library somewhere rather than coded directly into the sample Device application?

packages/matter-node.js/README.md
```diff
@@ -40,6 +40,10 @@ For instance, on a Raspberry Pi, this will turn on / off the red LED:
 matter -on "echo 255 > /sys/class/leds/led1/brightness" -off "echo 0 > /sys/class/leds/led1/brightness"
 ```
 
+The following parameters are available:
+* -announceinterface: limit announcements to the provided network interface (default: all interfaces available)
```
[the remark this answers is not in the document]
Apollon77: Just en0. I add an example. Interface name as shown by ipconfig or ip a

packages/matter-node.js/src/matter/MatterDevice.ts
Apollon77: Yes sorry for that

packages/matter-node.js/src/matter/MatterDevice.ts
Apollon77: MatterDevice is the "device node" logic level right now. (Example device is Device.ts). That will move also with high level api anyway.

packages/matter-node.js/README.md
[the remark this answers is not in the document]
Apollon77: adjusted, and jetzt it is the interfacet to use for mdns announcements. This shhelps to supress errors like https://github.com/mfucci/node-matter/issues/184 (or we need to add a better filter logic for the network interfaces to use for announcements but no idea after which logic :-() So allowing users to limit to one was best odea we had

Apollon77:
@turon I adjusted Readme to better state what the announceinterface is for and I answered your other comments. Announcement will also be adjusted with high level API

vves:
Additionally - the importance of the announceinterface is that on multi-nic systems, each network interface will announce a different *.local mdns name.  In my case, I have two physical and one virtual nic and I have three different devices announced on mdns unless a single interface is selected.

turon (approved):
Thanks for moving away from hard-coded port!

PR #65, opened 2023-03-19
Handle subscription cancellations and error handling

Apollon77:
This PR adds:
* checking relevant StatusResponses for subscription datareports that should lead to a cancellation according to specs
* still catch errors from subscription update sending to prevent crashing with SmartLife app because this happens when Andoid devcie goes to sleep/offline and would crash our device right now

### DIFF
--- a/packages/matter-node.js/src/matter/interaction/SubscriptionHandler.ts
+++ b/packages/matter-node.js/src/matter/interaction/SubscriptionHandler.ts
@@ -5,18 +5,20 @@
  */
 
 import { MatterDevice } from "../MatterDevice";
-import { InteractionServerMessenger } from "./InteractionMessenger";
+import { InteractionServerMessenger, StatusResponseError } from "./InteractionMessenger";
 import { Fabric } from "../fabric/Fabric";
 import { AttributeWithPath, AttributePath, INTERACTION_PROTOCOL_ID, attributePathToId } from "./InteractionServer";
 import { Time, Timer } from "../../time/Time";
 import { NodeId } from "../common/NodeId";
 import { TlvSchema } from "@project-chip/matter.js";
+import { tryCatchAsync } from "../../error/TryCatchHandler";
+import { StatusCode } from "./InteractionMessages";
 import { Logger } from "../../log/Logger";
 import { SecureSession } from "../session/SecureSession";
 
 const logger = Logger.get("SubscriptionHandler");
 
-const SUBSCRIPTION_MAX_INTERVAL_PUBLISHER_LIMIT_MS = 1000 * 60 * 60; /** 1 hour */
+const SUBSCRIPTION_MAX_INTERVAL_PUBLISHER_LIMIT_MS = 3 * 60 * 1000; //1000 * 60 * 60; /** 1 hour */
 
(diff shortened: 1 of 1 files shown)


packages/matter-node.js/src/matter/interaction/SubscriptionHandler.ts
```diff
@@ -5,18 +5,20 @@
...
 import { TlvSchema } from "@project-chip/matter.js";
+import { tryCatchAsync } from "../../error/TryCatchHandler";
+import { StatusCode } from "./InteractionMessages";
 import { Logger } from "../../log/Logger";
 import { SecureSession } from "../session/SecureSession";
 
 const logger = Logger.get("SubscriptionHandler");
 
-const SUBSCRIPTION_MAX_INTERVAL_PUBLISHER_LIMIT_MS = 1000 * 60 * 60; /** 1 hour */
+const SUBSCRIPTION_MAX_INTERVAL_PUBLISHER_LIMIT_MS = 3 * 60 * 1000; //1000 * 60 * 60; /** 1 hour */
```
turon, line 21: Why did this change?

turon (approved):
Is there a test for this?

packages/matter-node.js/src/matter/interaction/SubscriptionHandler.ts
Apollon77: I will add a comment before I merge. 

In fact former behavior is as per spec. But this leads to the effect that if you restart the device the controllers wait the agreed max amount of time to resubscribe (because for them all is fine). This can be a long time and I found that not making sense in real usage because the device could so stay "unchanged" for up to 45 mins in our implementation.

That's why I lowered to 3 mins. 

Additionally inward experiencing that Chiptool is also not using the 1h max - Chiptool uses the "max time" send by device as max, so updates could come really often if the device requests so. This is also not making too much sense because so you spam the network with empty subscriptions info messages. That's why I tried to do something in the middle.

Apollon77:
Unfortunately no test but tested locally with iOS and google.

Apollon77:
"proof":

```
2023-04-03 00:55:27.213 DEBUG SubscriptionHandler Sending subscription update message for ID 2781829980 with 0 values
2023-04-03 00:55:27.213 DEBUG MessageExchange new MessageExchange 1 56156 300 5000 5
2023-04-03 00:55:27.213 DEBUG SubscriptionHandler Sending subscription changes for ID 2781829980: []
2023-04-03 00:55:27.213 DEBUG ExchangeManager sending id:44380/56156/1346505783 t:1/5 reqAck:true payload: 1526005c5bcfa5360118290424ff0118
2023-04-03 00:55:27.226 DEBUG MessageExchange onMessageReceived 1 id:20397/56156/94391172 t:1/1 acked:1346505783 reqAck:true payload: 1524007d24ff0118
2023-04-03 00:55:27.227 INFO SubscriptionHandler Subscription 2781829980 cancelled by peer.
```

tested with puttion my computer to sleep and wake up long time afterwards then next subscription message is declined by iOS (in this case) with error 7d (Invalid subscription) ... so we cancel in this case sending too

PR #68, opened 2023-03-23
Solve TS Config inconsistencies

Apollon77:
While checking with the new "main branch" and my PRs I found several issues:

1.) strictNullChecks was missing as settings when generating dist files and thats why "nullable" types were generated wrongly in d.ts files (without allowing null)
2.) matter-node.js was no longer generating d.ts files

After adding this i needed to fix some minor things.

Additionally I added strict:true to the dev checks because I think this also makes a lot of sense coding wise.

### DIFF
--- a/packages/matter-node.js/tsconfig.base.json
+++ b/packages/matter-node.js/tsconfig.base.json
@@ -3,6 +3,7 @@
   "compilerOptions": {
     "target": "ES2015",
     "module": "CommonJS",
+    "strictNullChecks": true
   },
   "references": [
       { "path": "../matter.js" }
--- a/packages/matter-node.js/tsconfig.base.json
+++ b/packages/matter-node.js/tsconfig.base.json
@@ -3,7 +3,7 @@
   "compilerOptions": {
     "target": "ES2015",
     "module": "CommonJS",
-    "strictNullChecks": true
+    "strictNullChecks": true, // Required to correctly generate Nullable types in d.ts
   },
   "references": [
       { "path": "../matter.js" }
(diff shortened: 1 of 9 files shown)


packages/matter-node.js/tsconfig.base.json
```diff
@@ -3,6 +3,7 @@
   "compilerOptions": {
     "target": "ES2015",
     "module": "CommonJS",
+    "strictNullChecks": true
```
mfucci, line 6: Add a comment what this is needed and what is the effect

packages/matter.js/tsconfig.base.json
```diff
@@ -3,5 +3,6 @@
   "compilerOptions": {
     "target": "ES2015",
     "module": "ES2015",
+    "strictNullChecks": true,
```
mfucci, line 6: Same, add comment

packages/matter.js/tsconfig.dist-cjs.json
```diff
@@ -3,7 +3,7 @@
   "extends": "./tsconfig.base.json",
   "compilerOptions": {
     "outDir": "dist/cjs",
-    "module": "CommonJS",
+    "module": "CommonJS"
```
mfucci, line 6: revert to keep trailing comma

mfucci (approved):

packages/matter-node.js/tsconfig.base.json
Apollon77: added  // Required to correctly generate Nullable types in d.ts

packages/matter.js/tsconfig.base.json
Apollon77: added  // Required to correctly generate Nullable types in d.ts

packages/matter.js/tsconfig.dist-cjs.json
Apollon77: added, also in two other places

PR #70, opened 2023-03-26
Enhance persistence Json to also support Maps

Apollon77:
This PR does the following:
* Split out the custom JSON Converter methods into an own file so that they can also be used by other storage implementations
* Add support to alaps persist and restore "Map" objects
* Add Tests for the json converter

### DIFF
--- a/packages/matter-node.js/test/persistence/JsonConverterTest.ts
+++ b/packages/matter-node.js/test/persistence/JsonConverterTest.ts
@@ -25,7 +25,7 @@ describe("JsonConverter", () => {
 
             const json = toJson(obj);
 
-            assert.equal(json, "{\"aNumber\":1,\"aNumberString\":\"2\",\"aString\":\"hello\",\"aBoolean\":true,\"aNull\":null,\"anObject\":{\"a\":1,\"b\":2,\"c\":3},\"anNumberArray\":[1,2,3],\"anStringArray\":[\"a\",\"b\",\"c\"],\"anObjectArray\":[{\"a\":1},{\"b\":2},{\"c\":3}]}");
+            assert.equal(json, `{"aNumber":1,"aNumberString":"2","aString":"hello","aBoolean":true,"aNull":null,"anObject":{"a":1,"b":2,"c":3},"anNumberArray":[1,2,3],"anStringArray":["a","b","c"],"anObjectArray":[{"a":1},{"b":2},{"c":3}]}`);
 
             const decodedObj = fromJson(json);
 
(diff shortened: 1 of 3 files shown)


packages/matter-node.js/test/persistence/JsonConverterTest.ts
```diff
@@ -0,0 +1,106 @@
...
+            assert.ok(decodedObj instanceof Map);
+            assert.equal(decodedObj.get("a"), 1);
+        });
+
+        it("encode/decide Map of Maps", () => {
+            const obj = new Map([[1, new Map([["a", 1]])], [2, new Map([["b", 2]])]]);
+
+            const json = toJson(obj);
+
+            assert.equal(json, "\"{\\\"__object__\\\":\\\"Map\\\",\\\"__value__\\\":\\\"[[1,\\\\\\\"{\\\\\\\\\\\\\\\"__object__\\\\\\\\\\\\\\\":\\\\\\\\\\\\\\\"Map\\\\\\\\\\\\\\\",\\\\\\\\\\\\\\\"__value__\\\\\\\\\\\\\\\":\\\\\\\\\\\\\\\"[[\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\"a\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\",1]]\\\\\\\\\\\\\\\"}\\\\\\\"],[2,\\\\\\\"{\\\\\\\\\\\\\\\"__object__\\\\\\\\\\\\\\\":\\\\\\\\\\\\\\\"Map\\\\\\\\\\\\\\\",\\\\\\\\\\\\\\\"__value__\\\\\\\\\\\\\\\":\\\\\\\\\\\\\\\"[[\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\"b\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\",2]]\\\\\\\\\\\\\\\"}\\\\\\\"]]\\\"}\"");
```
turon, line 93: Could you use  \` to quote the string w/quotes and avoid backslashes?

Apollon77: Adjusted and simplified one level ...

turon: It definitely looks better, but is there a way to get rid of the need for backslashes altogether if you use backtick `\`` or single-quote `'` around the outer string?

Apollon77: No, because they are there because it is a Json inside a JSON ... The idea of the encoding of these special values was to encode them as their own JSON ... so you end up in a "double encoding" when you stringify that again ... but so we are sure that the "special encoded values" are not touched by anything else ... 

We do that in the office if we have values that should be just passed through without anyone else should touch the - simpyl pass as string and done. But that approach in this case does "string in string" and so these backslashes are real :-)

Apollon77:
moved to #63

PR #73, opened 2023-03-28
Allow to run via ts-node

Apollon77:
This PR removed all file references that reference a js file where it should be ts that could prevent running via ts-node and adds npm scripts for matter and matter-controller to run from ts code directly and transpile on the fly.


mfucci:
I added all those ".js" in the imports when I added compatibility with es modules, since it seemed to be the only way to have dual support for commonjs & es modules.
Maybe I screwed up when I did that...
karma-jasmine unit tests on the browser are using commonjs module, so are not checking this.

Can you ensure that es module are still work as expected from a browser?
I think browsers are only resolving imports when the file name ends by ".js"...
Create a simple web project that checks this and ensures it is still working?

- create a new project and link matter.js
- in tsconfig.json, indicate "module": "ES2015"
- do a stupid operation using matter.js, like encoding an empty array over TLV
- link to it from an index.html page
- declare the linked script as a "module"
=> if es module is working correctly everywhere, this should work (this was working before this change)
I think, the page will crash because it cannot resolve the import without having .js at the end

If it is broken, we should figure out how to add an unit test for it.

Apollon77:
I will try but browser should use the generated files in "dist", or?! And they should have that correctly set when generated. And the web tests als run correctly.

Apollon77:
Maybe also https://www.npmjs.com/package/@open-wc/karma-esm could be an idea?

mfucci:
The distribution for es module will be in dist/es and in package.json, module is declared as "dist/es/matter.js".
So an es module project will pack the compiled from the right directory.
dts are the same for es and cjs distributions.

Yes, it is a mess... but I did get it work at some point.

For matter.js, I think it is important to support both es and cjs modules instead of being prescriptive.
For matter-node.js, it is fine to only be cjs modules since it is the module type natively supported by node.js. The es module support is only experimental.

Apollon77:
ooooookkkkkk ... that was a funny learnign experience today ...

Yes, if we want to support browser then we need to have ALL imports with ".js", because the TypeScript guys decided that they will never ever rewrite file extensions oif the target donot require it... so we now need to have .ts files that import .js files (sorry but WTF... THAT is a mess ;-) )

So I now adjusted the files again and added in matter.js ALL files to import the js ones (because also there was a mixup earlier). I also adjusted the tsconfigs so that it is detected durcing build and also development time if this is missing in matter.js files.
For this I needed to add "type: module" to packge.json which should from my tries still work with our esm/cjs mixed package (but is needed for ts to support the above). When we next deploy e need to verify that again, but I'm confident.

So, R4R again

PS: I also spent some time in trying to get karma to also do esm testing but failed finding a solution

turon:
First off thanks for adding `ts-mode` support!  It really helps with fast iteration during development, which is a key feature of javascript.

Yes, it can be messy to setup.  Between all the `js` vs `ts` file extensions and `dist` vs `src` path mixups it indeed can quickly become a very messy house of cards paradox.  When I was adding similar support to `matternode`, I compared `@swc-node/register` to `ts-mode` and noted `@swc-node/register` is significantly faster due to essential JIT support which bypasses compile-time type checks.  That said, `ts-mode` has very broad use and support and is easier to setup reliably.  I'm fine with the choice. But if app start-up times become too large when using `ts-mode` we have another option to consider.

Apollon77:
in fact @swc-node/register was not liking the static methods - but ok I also did not invested too much time to find the reason for now ...  So if you please approve :-)

Apollon77:
I found a solution with this the matter.js package.json stays "type module" as needed but the cjs dist dir is handled as commonjs.

tested and verified working for me. Idea from https://www.sensedeep.com/blog/posts/2021/how-to-create-single-source-npm-module.html

packages/matter-node.js/package.json
```diff
@@ -27,7 +27,9 @@
         "clean": "rm -rf dist && rm *.tsbuildinfo || true",
         "build": "tsc --project tsconfig.dist.json",
         "build-clean": "npm run clean && npm run build",
-        "test": "TZ=utc jest"
+        "test": "TZ=utc jest --maxWorkers=5",
```
mfucci, line 30: Maybe can be moved to jest.config?

packages/matter.js/package.json
```diff
@@ -25,7 +25,10 @@
   },
   "scripts": {
     "clean": "rm -rf dist && rm -rf out && rm *.tsbuildinfo || true",
-    "build": "tsc --project tsconfig.dist-es.json && tsc --project tsconfig.dist-cjs.json && tsc --project tsconfig.dist-dts.json",
+    "build-cjs": "tsc --project tsconfig.dist-cjs.json && cp package-cjs.json dist/cjs/package.json",
```
mfucci, line 28: Maybe the copy can be done by adding this file in the tsconfig.dist-cjs.json as part of the include ?

mfucci (approved):

packages/matter-node.js/package.json
Apollon77: yes works, adjusted - also to 50% to be more generic

packages/matter.js/package.json
Apollon77: Thoisis with JSON "can" work but only if you also add "resolveJsonModule" to true which I'm not sure we want (I would propose not). So the cp is the best solution for now

PR #82, opened 2023-04-09
Sync InteractionMessage definitions matter-node.js -> matter.js

Apollon77:
This PR is doing the following:
* It mainly syncs the InteractionMessage definitions that were in matter.js already in an outdated version with the completed ones from matter-node.js.
* It is just moving code and adjusting needed imports, no other changes included (beside following point)
* Because of the current "flat everything exports" from matter.js there was a name conflict and so I just renamed both "StatusCode" to be uniquely identifyable. We can revert that rename once we change the exports

One note: The matter-node.js version still contains some todos for "renames" of fields to stick better with specs, this is planned for later still because it touches several files and conflice with next PRs.

This should be easy to review and is a needed preparation before the main code move. Thank you


packages/matter-node.js/src/matter/interaction/InteractionClient.ts
```diff
@@ -7,14 +7,16 @@
...
+    Attribute, AttributeJsType, Attributes, Cluster, Command, Commands, TlvNoResponse, RequestType, ResponseType,
+    TlvSchema, TlvStream, InteractionProtocolStatusCode
+} from "@project-chip/matter.js";
 import { DataReport, IncomingInteractionClientMessenger, InteractionClientMessenger } from "./InteractionMessenger";
 import { ResultCode } from "../cluster/server/CommandServer";
 import { ClusterClient } from "../cluster/client/ClusterClient";
 import { ExchangeProvider } from "../common/ExchangeManager";
 import { INTERACTION_PROTOCOL_ID } from "./InteractionServer";
 import { ProtocolHandler } from "../common/ProtocolHandler";
-import { StatusCode } from "./InteractionMessages";
```
vves, line 17: Let's consider using `import {InteractionProtocolStatusCode as StatusCode} from "@project-chip/matter.js"` to minimize diff issues in the future.  As long as the Admin and Interaction status codes are not used in the same files this should be bueno.

vves:
@Apollon77 once you sort out the StatusCode rename I'll do another review.  Should go quickly.

vves (changes requested):
Please minimize inline renames by aliasing to StatusCode.

packages/matter-node.js/src/matter/interaction/InteractionClient.ts
Apollon77: I agree, good idea, especially because I want to rename back after the code merge :-)

Apollon77:
You can re review

packages/matter.js/src/protocol/InteractionProtocol.ts
```diff
@@ -45,132 +44,211 @@ export const enum StatusCode {
...
-        cluster: TlvField(2, TlvUInt32),
-        event: TlvField(3, TlvUInt32),
-        isUrgent: TlvField(4, TlvBoolean),
-    }))),
-    eventFilters: TlvOptionalField(5, TlvArray(TlvList({
-        node: TlvField(0, TlvNodeId),
-        eventMin: TlvField(1, TlvUInt64),
-    }))),
+    eventRequests: TlvOptionalField(4, TlvArray(TlvEventPath)),
+    eventFilters: TlvOptionalField(5, TlvArray(TlvEventFilter)),
```
vves, line 207: Was TlvEventFilter defined before or did you refactor?

vves (approved):

PR #83, opened 2023-04-09
Abstract Crypto Implementation

Apollon77:
As preparation for the code move this PR does the following:
* It abstracts Crypto class into a generic base class and uses it via singleton like Timer and Network already
* Becaue TS do not support "abstract static" class members I decided to use a normal class which becomes via the singleton again kind of "static"
* Additionally the places where "Buffer" was returned by Crypto methods and so also ended in storage the Buffer is always converted into a ByteArray type. Becauseof this also Buffer support is removed from Storage. IMPORTANT: This breaks existing storages potentially!!

The PR should be relatively straight forward to review

### DIFF
--- a/packages/matter-node.js/src/Controller.ts
+++ b/packages/matter-node.js/src/Controller.ts
@@ -17,6 +17,11 @@ import { NetworkNode } from "./net/node/NetworkNode";
 
 Network.get = singleton(() => new NetworkNode());
 
+import { Crypto } from "./crypto/Crypto";
+import { CryptoNode } from "./crypto/CryptoNode";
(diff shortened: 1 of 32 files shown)


Apollon77:
Yes exactly. Currently Crypto is a "core" function but implemented node.js specific and the goal is to allow platform specific implementations (platform aka Web, Nodejs, maybelow.js or whatever other JavaScript based implementations). Thats why all these topics are currently built in an abstractable way (e.g. Network, Time/Date), so Crypto is just "the same"

vves (approved):

PR #84, opened 2023-04-09
TlvNoArguments should be void

Apollon77:
TlvNoArguments was an alias for `TlvObject({})` until now which lead to the need to send an empty object if there are no request parameters which is strange. In fact according to Matter specs the arguments can be supressed completely when there are none, so I adjusted the tpe to be TlvVoid.
This PR adjusts this and the other relevant needed places for it that can now be "undefined" (aka void type)


packages/matter-node.js/src/matter/cluster/server/CommandServer.ts
```diff
@@ -23,11 +23,11 @@ export class CommandServer<RequestT, ResponseT> {
         readonly name: string,
         protected readonly requestSchema: TlvSchema<RequestT>,
         protected readonly responseSchema: TlvSchema<ResponseT>,
-        protected readonly handler: (request: RequestT, session: Session<MatterDevice>, message: Message) => Promise<ResponseT> | ResponseT,
+        protected readonly handler: (request: RequestT | undefined, session: Session<MatterDevice>, message: Message) => Promise<ResponseT> | ResponseT,
```
vves, line 26: can the request really be undefined per the spec?

packages/matter-node.js/src/matter/interaction/InteractionMessages.ts
vves, line 1: Tough to tell from the spec 10.5.12.1 CommandFields : "If there are no arguments in the Request or Response, this tag MAY be omitted entirely."  Which tag?   From what I can tell, the CommandData can be empty, but not optional.  Is that what you were sorting out here?

Apollon77: It is about the "args" tag in this case ...

packages/matter-node.js/src/matter/cluster/server/CommandServer.ts
Apollon77: The specs say that "if there are no parameters then the request can be omitted in the invoke structure". In fact here "RequestT" stands for "the parameters" ... but now that yousay it ... let me check something tomorrow

packages/matter-node.js/src/matter/interaction/InteractionMessages.ts
Apollon77: The commandData needs to be there but the args field is optional.

Apollon77:
I adjusted the PR the following way:
1.) "Receiving side": the "args" re optional because according to specs "MAY be omitted when no arguments", so the Field needs to get optional. Nevertheless we now internally detect this case and use the "TlvNoArguments" as value instead fort the invoke call, so the "| undefined" is not needed

2.) "Sending side": TlvVoid as value works fine that you can call the method without parameters, but would encode to not sending the args field (which is ok according to specs - it's that "MAY"). In fact no other controller sends it that way right now so I decided to detect tht case and send an empty object till now in the invoke request (so we convert TlvVoid -> TlVObject({}) befire sending). I know this is formally not needed but before we get issues with the "MAY" part it is the secure way.

packages/matter-node.js/src/matter/interaction/InteractionClient.ts
```diff
@@ -175,9 +175,12 @@ export class InteractionClient {
 
     async invoke<C extends Command<any, any>>(endpointId: number, clusterId: number, request: RequestType<C>, id: number, requestSchema: TlvSchema<RequestType<C>>, _responseId: number, responseSchema: TlvSchema<ResponseType<C>>, optional: boolean): Promise<ResponseType<C>> {
         return this.withMessenger<ResponseType<C>>(async messenger => {
+            // if the invoke call do not have arguments we want to send an empty tlv object to not suppress the args field (even if MAY be allowed by specs)
+            const args = (requestSchema === TlvNoArguments && request === undefined) ? TlvObject({}).encodeTlv({}) : requestSchema.encodeTlv(request);
```
mfucci, line 179: TlvSchema are made to encapsulate encoding / decoding so TlvNoArguments should define its own schema instead of adding complexity here:
```
const TlvEmptyObject = TlvObject({});

export class TlvNoArguments extends TlvSchema<void> {

    override encodeTlvInternal(writer: TlvWriter, value: void, tag: TlvTag = {}): void {
        if (value !== undefined) throw new Error("No value should be passed");
        TlvEmptyObject.encodeTlvInternal(writer, {}, tag);
    }

  [...]

}
```

packages/matter-node.js/src/matter/interaction/InteractionServer.ts
```diff
@@ -323,6 +323,9 @@ export class InteractionServer implements ProtocolHandler<MatterDevice> {
         await Promise.all(invokes.map(async ({ path, args }) => {
             const command = this.commands.get(commandPathToId(path));
             if (command === undefined) return;
+            if (args === undefined) { // If no arguments are provided in the invoke data, we use an empty TlvStream
```
mfucci, line 326: I am confused: from the modification above, this is never undefined but an empty object instead.
What am I missing?

mfucci (commented):
It would be good an unit test that captures the expected behavior: it is hard to understand the expected behavior and if this code changes doesn't break any tests, this means that unit tests are missing...

packages/matter-node.js/src/matter/interaction/InteractionServer.ts
Apollon77: If we have an "optionalField" and this is not provided ... then the value is "undefined", but I think with the above idea of  an own Schema that gets all simplified

packages/matter-node.js/src/matter/interaction/InteractionClient.ts
Apollon77: done

Apollon77:
Test also added for both variants to invoke a command (with empty args and "not provided" args aka Optional)

Ready for re-review

turon (approved):

PR #86, opened 2023-04-10
Controller Low Level API Enhancement

### LINKED ISSUE #48: Validate "incoming" (decoded) data to defined shema breaks with List chunking
While playing around with node-matter and pairing as controlelr with chip-tool lighting app I came across the protocol options for array/List chunking.

The chip app always send an empty array and then "add" elements to it in additional packages. This is in fact also needed if an array gets that big that it do not fit in one data package.
While this topic itself can not be handled on "pure Tlv level" (aka /src/tlv/TlvArray.ts) the current issue starts when we have a schema with e.g. "minimum 1 list entry". Because decoding the first tlv package gives an empty array and this will fail because we do validation.

The main question that arises is: If devices is sending a invalid value (aka "decode for the library), should that be an exception? In fact we can not do anything against such "Misbehaving devices". In fact yes, it should be still an exception if we want to encode (aka send ourself) such a value, so the shema validation should be used when setting such a value.

The alternative would be to allow a flag when decoding values to not validate the value and the caller ist then self-responsible. (so after processing all value parts we could call validate ourself)

Details for that "Change" (Chunking) handling is in Core Specs "10.5.4.3.1. Lists".

As an example this is the DataReport received from chip tool app for reading one value:

```
{
    "values": [
        {
            "value": {
                "version": 2206465283,
                "path": {
                    "endpointId": 0,
                    "clusterId": 29,
                    "attributeId": 0
                },
                "value": [
                    {
                        "typeLength": {
                            "type": 22
                        }
                    },
                    {
                        "typeLength": {
                            "type": 24
                        }
                    }
                ]
            }
        },
        {
            "value": {
                "version": 2206465283,
                "path": {
                    "endpointId": 0,
                    "clusterId": 29,
                    "attributeId": 0,
                    "listIndex": null
                },
                "value": [
                    {
                        "typeLength": {
                            "type": 21
                        }
                    },
                    {
                        "tag": {
                            "id": 0
                        },
                        "typeLength": {
                            "type": 4,
                            "length": 0
                        },
                        "value": 22
                    },
                    {
                        "tag": {
                            "id": 1
                        },
                        "typeLength": {
                            "type": 4,
                            "length": 0
                        },
                        "value": 1
                    },
                    {
                        "typeLength": {
                            "type": 24
                        }
                    }
                ]
            }
        }
    ],
    "suppressResponse": true,
    "interactionModelRevision": 1
}
```

The "Change" logic is within the listIndex field when getting attribute responses. In this case listIndex=null means "add these data to the array".

Apollon77:
### Attention!!
Beside this PR I decided to also open #106 which is bringing this content together with a big refactoring of that. Ideally we merge only #106 and close this one. To be decided

This PR adds the following controller "low level API" features that were missing so far:
* Allow to read not only single attributes, but also multiple up to */*/*. In the read data the oath is enhanced by the attributeName for convenient usage
* Adds generic Attribute decoding based on the existing Cluster definitions - all other data are ignored right now when read!
* Adds Array de-chunking during the value decoding
* Allow to subscribe not only to single attribites, but also to multiple up to */*/*
* Subscribed values are cached internally and a "get" call on these is using the cached value instead of requesting the value from the device, therefor subscription listener was made optional
* Allow to set/write values - single ones and also multiple
* Add support to send "TimedRequests"
* Add internal support to send requests with "suppressResponse=true" (write, invoke) to correctly expect response type
* Add Tooling class ClusterHelper and methods to lookup clusters and also cluster Attributes by Id
* Add info to Readme about "what the current controller actually is" - just a dev showcase and debugging/testing tool to make sure expectations are set

All this Low level API will ideally never used by a dev using the library! They should all use the upcoming high level API!

fixes #48


Apollon77:
superceeded by #106

PR #99, opened 2023-04-16
Color control work

### LINKED ISSUE #97: Add Color Control Cluster
Working this, but it will take a while with around 50 Attributes, nearly 20 Commands any many enums and bitmaps.

JimBuzbee:
Continued work to meet standard. Still need to double-check against standard, check formatting and define "TODOs" showing how dependencies affect Optional vs. Mandatory for Commands and Attributes. XML has "clusterExtension" section for extended commands and attributes, but standard does not seem to handle these differently except for them having their own range of IDs

fixes #97


Apollon77:
@JimBuzbee Please also add to https://github.com/project-chip/matter.js/blob/main/packages/matter.js/src/cluster/ClusterHelper.ts#L35

JimBuzbee:
Will generate new PR updated for new API

PR #105, opened 2023-04-20
Initial push of ModeSelectCluster

### LINKED ISSUE #104: Add Mode Select Cluster
Working this one. Interesting cluster that allows fairly arbitrary functionality to be presented to the controller with an overall "description" string and a list of "label" strings that a user would select from in order to tell the cluster to do an operation or switch "modes".

JimBuzbee:
Initial - still a few cleanups to do

fixes #104


Apollon77:
@JimBuzbee please also add it to https://github.com/project-chip/matter.js/blob/main/packages/matter.js/src/cluster/ClusterHelper.ts#L35

Apollon77:
@JimBuzbee I have updated it a bit here ... should be ok that way ... so if you like please also check and rebase/merge master into it

Apollon77:
@JimBuzbee Thank you very much for your contribution. I close the PR as we now got auto generated clusters within #192. I verified it and it is "the same". so that helped a lot.

PR #112, opened 2023-04-26
Codemove and Library exports introduction

### LINKED ISSUE #62: Complete node-matter->matter.js Initial File move concept
@mfucci @turon I will write together here how I propose to m,ove the stuff from node-matter which is not node.js dependent into matter.js. Please review and comment it, so that we can discuss this before I actually do it and I will then execute exactly as agreed to prevent double/triple work!

I show my move concept in https://gist.github.com/Apollon77/64224fa6c867958c90f2be2dfc495817 ... please check and comment there

Apollon77:
This PR is completing the code move of all "Non-node.js" functionality into the matter.js package.

Additionally it introduces an "topic based" export structure for the packages which allows to export the funcions still simplified, but not in a "one export for anything" structure.
Tests are also moved from the matter-node repo to matter.js best possible. Some tests needed to stay in matter-node because they use CryptoNode functionalities and we do not (yet) have a pure JS Crypto lib. Because of the fact that we always run all tests it is made sure that still anything is tested as before. We can move them later too.

The PR is not changing any code beside the move relevant import changes, adjustments to tsconfig, eslint and package.jsons.

I also adjusted the Readmes to describe the exports and also streamline the Readme's a bit.

I think this PR is near to impossible to review using GitHub comparism view, so maybe explore the final code in https://github.com/Apollon77/matter.js/tree/codemove/packages directly.

And check the Readme's at https://github.com/Apollon77/matter.js/blob/codemove/packages/matter.js/README.md and https://github.com/Apollon77/matter.js/blob/codemove/packages/matter-node.js/README.md

The PR also incorporates the PR #113 from @lauckhart

fixes #62


Controlhaus:
I was able to build this PR, run node `./dist/DeviceNode`, pair the device with the Alexa App and see on & off commands arrive in terminal. The on & off commands continued to work after restarting matter-node, so persistence is working as well.

One tiny grammar error in both readme's: the L in cluster should be lower case : ) ` Exports Matter CLuster definitions and Default Cluster-Handlers`

Apollon77:
Typo fixed

Apollon77:
PR enhanced by some tsconfig hints after feedback from @Controlhaus

Koenkk:
Looks good, nice refactor 😃

Apollon77:
Optimized tsconfig even more to allow local linking also for matter-node.js package

Controlhaus:
I was able to implement all the exports used in DeviceNode.ts in my own implementation successfully. Pairing and device control tested ok with Alexa.

adminy:
The code is still highly inter-dependent and monolithic looking, but easier to work with now.

vves:
Should be easier to work with.  Approving as all compiles, all tests run, and examples are functional.

vves (approved):
Let's do this.  It's a big change and all is probably not perfect, but this is a better structure.

lauckhart:
Makes sense from my admittedly limited perspective.  Thanks for the added documentation, that's helpful.

Here's some quick thoughts.  Probably best to accept this PR and iterate from there...  But these could change the public API so figured you might want to consider now.  LMK if it'd be better to open as separate issue(s) and/or PRs.

* Storage currently acts like a singleton but the pattern is different vs. crypto/time/net.  Makes sense if there will be different storage pools or methods for a single platform in the future...  But I wonder if "default" storage could be handled by a singleton so everything platform-specific is configured the same way.
* Much of the remaining code in `DeviceNode` and `ControllerNode` doesn't seem to be Node specific.  Could this be moved into base classes in matter.js or a new package?
* Or even better, maybe have Device/Controller classes that handle platform specific bits via composition?  So e.g. `QrCodePresenter` (eww, just an example) similar to `StorageManager`.  The benefit would be that if I create `BridgeDevice extends Device` it could be used in either Node.js or React Native.  `BridgeDevice extends DeviceNode` would tie the implementation to Node.
* R.e. duplicative singleton registration in `DeviceNode` and `ControllerNode`, maybe create `MatterNode.registerProviders()` or even just have `import '@project-chip/matter-node'` do registration on module initialization?  Would reduce redundancy but mostly relevant in light of previous point as the library user would use to initialize their platform.
* If I understand correctly, except for the hard-coded On/Off Light endpoint, the call to start(), and a few missing extension points, DeviceNode looks like a good base class for any device!  Is that generally where you're headed?

Apollon77:
Hi @lauckhart,

thank you very much for your review and your thoughts. This PR mainly focus on code structuring, so it do not touch existing functionality.

But I want to take the chance to answer your questions and give a bit background on the "history and future" ;-)

to 1.) We started as singleton for storage but then this broke in testing because there we needed a "controller" and a "device" storage in parallel and they have overlapping data to store. We will see if the planned high level API changes that, but for now it was the better and more flexible way

to 2.) Yes I also thought about that but that would have required several code changes in the "Device/ControllerNode class" but I wanted to do least possible "other code changes" becasue the PR gets big enough - additionally the new High level API (review coming soon after this got merged) will change all that dramatically - then exactly this happens :-)

to 3-5.) The High level API will be your friend ;-) lets discuss when I posted my proposal :-)

lauckhart:
👍 Thanks, makes complete sense.  I found your discussions r.e. high-level API in Discord after I posted.  I had missed the elegantly named #matterjs-nodejs-node-matter channel 😁

Apollon77:
Lol. Might get renamed soon ;-)

PR #113, opened 2023-04-27
Add declaration map for .d.ts files

lauckhart:
Kind of trivial...  Allows you to click a symbol imported from matter.js in VS Code and go to the source .ts instead of the .d.ts file.

### DIFF
--- a/packages/matter.js/tsconfig.dist-dts.json
+++ b/packages/matter.js/tsconfig.dist-dts.json
@@ -5,6 +5,7 @@
     "moduleResolution": "node", // Required to correctly resolve Babel imports
     "outDir": "dist/dts",
     "declaration": true,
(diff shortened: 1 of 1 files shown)


Apollon77:
@lauckhart I curremntly prepare a big code move (see #112 ) ... if ok I would incorpoate your change there (because that one removes that extra build step and combines it into the es/cjm ones

Apollon77:
@lauckhart I incorporated your PR in https://github.com/project-chip/matter.js/pull/112/commits/3b1b371ac29a5c81722cee598e81319d2c68ad2e into the mentioned PR, and extended it to be for both Repos.

If you like please have a look also on #112 which should be (after reviews) the future basis of code structure and also exports structure

PR #114, opened 2023-04-28
Bring back Crypto Static interface

Apollon77:
This PR adjusts the crypto abstraction back to an interface that can be used as static methods. Additionally it removed the kast "Buffer" usages from the code


vves (approved):
much prefer this pattern instead of the singleton() pattern. Cheers.

PR #115, opened 2023-04-28
Consolidate and Rename Interaction Protocol Types

Apollon77:
This PR is solving some planned renamings to consolidate the Interaction Protocol datatypes and Objects to have the same naming as in the specs and also renames "InteractionProtocolStaticCode" back to "StatucCode" now that the new export structure prevents conflicts here.


vves (approved):
thank you for clearing up some of the naming.  No slight to those who put this library in place as that was a lift, but it's very helpful to have names other than `value` or `path` when there are multiple values or paths.   The elevation of `TlvStatus` type is a nice addition.  Cheers All.

Issue #117, opened 2023-04-30
Clean fabric


lboue:
Hello,

Could you add the procedure to remove the pairing and reset the node?

Regards

lboue:
I noticed with that command it worked:
```
rm device-node/FabricManager.fabrics
```

Apollon77:
In fact you need ideally to kill the whole storage dir because also in other places (e.g. session manager) there could be data that should also be removed

Apollon77:
done differently via new parameter -cleastorage

lboue:
Thanks

PR #118, opened 2023-04-30
Update README.md to clear existing pairings

lboue:
Clear existing pairings.
Fix for #117


packages/matter-node.js/README.md
```diff
@@ -151,6 +151,12 @@ The following parameters are available:
 * -pin: the pin to use for pairing (default: 20202021)
 * -store: the storage location (directory) to use for storing the pairing information (default: controller-node). Delete the directory or provide an alternative name to reset the controller
 
+### Clear existing pairings
+To clear existing pairings:
+```
+rm device-node/FabricManager.fabrics
```
Apollon77, line 157: I would rater remove the whole storage directory and start over because also sessionmanager has infos and such. SO it would maybe better to "how to I clear the storage"?

Apollon77 (changes requested):

PR #121, opened 2023-04-30
Add Endpoint data to invoke parameters

Apollon77:
Some cluster logic requires details from the endpoint and later also access to other clusters on the endpoint. This PR introduces this in a minimalistic way - it will get refactored once again when new API is fully in place. But it also helps the use for GroupsCluster and allows in parallel Cluster logic development (e.g. Lighting control for onoff or scenes) - even with a refactoring later.


vves:
@Apollon77 I wonder if you can roll this, #123, #100, #131, and #134 into a single PR.  Each looks ok on their own, but would be more effective to test/review all as one fully functional PR to help avoid regressions during merge.

vves (approved):

PR #123, opened 2023-04-30
New High Level API

Apollon77:
This is the beginning of a bigger PR to introduce a new high level API that replaces the current low level one and should be the API to use for the future by developers. The PR is there to allow real development experience with the API to collect real-usage-feedback to iterate until it works for all cases.
Some special parts are also still WIP and will be enhanced by new commits in the near future.

It still internally bases on some internal main classes "from the past" where I decided to rebuold them in a separate PR later! So some functionalities that the API offers in general (meinly "multi node capabilities" will only exist theoretically on API level but can not be used for now!

This API also includes Bridge and Composed Device functionality that works with iOS and Google.

In order to "see" the changes I decided to create some additional MD files:
* https://github.com/Apollon77/matter.js/blob/newapi/packages/matter-node.js/README_NEW.md -  please read this to get a status overview of the PR
* https://github.com/Apollon77/matter.js/blob/newapi/packages/matter.js/API.md - please read this to get an overview

For feedback please use GitHub comments or review possibilities or join the Discord and use the channel "newapi-matterjs" (https://discord.com/channels/1071473433482891284/1102285806636367882 in Matter-Integrators Discord https://discord.gg/ujmRNrhDuW)

PS: Device side of API is basically tested, Controller not fully


Apollon77:
Some images of bridge and composed devices in iOS 

image

image

image

Apollon77:
Good Idea, will change it with next update later today

packages/matter.js/src/protocol/interaction/InteractionClient.ts
```diff
@@ -35,9 +36,78 @@ export interface AttributeStatus {
...
+                logger.warn("Unknown attribute id", path.attributeId);
+                return;
+            }
+            (this.attributes as any)[attributeName].update(value);
+        });
+    }
+}
+
+export function ClusterClientInstance<CommandT extends Commands, AttributeT extends Attributes>(interactionClient: InteractionClient, endpointId: number, clusterDef: Cluster<any, AttributeT, CommandT, any>): ClusterClientInstance<CommandT, AttributeT> {
     const result: any = {};
```
adminy, line 104:
```javaascript
const { id: clusterId, name, commands, attributes } = clusterDef;
const result: any = { id: clusterId, name, endpointId, attributes, commands };
```

Apollon77: Yes, this ClientInstance is currentl kind iof unused and will coming in soon again. WIll then consider your idea. thank you very much

adminy: Really looking forward to a Client Cluster for things like sensors.

Apollon77: on my plan for this weekend

Apollon77: PS: Ok CLusterClients on a device (and not controller) might be another topic ... I assume that then also events join the game and we do not have them till now implemented

adminy: I wonder if it can be done without all the events implemented just yet,
Like just display a static value, like an information dummy card for now.
I really wanna see what that looks like on the home app.

Apollon77: Ok are we talkling about the same? If you want to offer a TemperatureSensorDevice this work already because there a TemperatureServer is used and NOT a client. so what exactly you want to do?

adminy: I thought that a device client is needed in order to report information about a device to the controller.
Like a server is something that listens to commands, while a client is something that sends commands back
to the matter controller and does not accept commands, as that would make it a device server.
You're saying that a temperature sensor is a device server?

adminy: Another example of a device client would be a physical switch, since its a state setter, it does not react to anything, its the thing that other things react to it. (until you add a state indicator and it becomes both a server and a client...).

packages/matter.js/src/protocol/interaction/AttributeDataDecoder.ts
```diff
@@ -162,3 +162,17 @@ export function decodeChunkedArray<T>(schema: ArraySchema<T>, values: TypeFromSc
     schema.validate(result);
     return result;
 }
+
+export function structureReadDataToClusterObject(data: DecodedAttributeReportValue[]) {
+    const structure: { [key: number]: { [key: number]: { [key: string]: any } } } = {};
+    for (const { path: { endpointId, clusterId, attributeName }, value } of data) {
+        if (structure[endpointId] === undefined) {
```
adminy, line 169:
```javascript
Object.assign(structure, {[endpointId]: {[clusterId]: {[attributeName]: value}}})
```
saves from writing too many if statements ;)

Apollon77: good catch .. thx, done

packages/matter.js/src/protocol/interaction/InteractionClient.ts
Apollon77: >  You're saying that a temperature sensor is a device server?

yes https://github.com/project-chip/matter.js/blob/main/packages/matter.js/src/common/DeviceTypes.ts#L630-L643

For cluster Clients this is a good example:

https://github.com/project-chip/matter.js/blob/main/packages/matter.js/src/common/DeviceTypes.ts#L690-L714

For ClusterClients, when I understand the specs rught, you utilize "Binding" as feature which uses the Binding cluster (see 9.4 and 9.6 in core specs).

So as I understand that is on example of an onOff Device:
OnOff device can have an optional Occupancy sensor client. This menas that you can optionally bind the onoff device to an Occupancysensor device (which then have a Occupancy Cluster SERVER). The the light can e.g. subscribe to that occupancy sensor and turn on/off the light based on this sensor.

So if you want to have a "device that measures a teperature and provides that value to ios Home app" then you need a Temperature devcie with a TemperatureMeasurement ClusterServer.

I hope now it got more clearer?

packages/matter.js/src/protocol/interaction/AttributeDataDecoder.ts
Apollon77: Ahhh nooooo ... this do not work!! Object assign is not working recursive with objects ...

Apollon77: Yes but a depenency to add for that some lines code is it not worth ;-)

lauckhart (approved):
LGTM, kind of improves readability too

vves (approved):
After a couple of weeks working with this branch, I have to say it's better than main!  The  fix was the cherry on top.

Apollon77:
Thanks to all testers and feedback providers here and in Discord. I will merge it now.

It will be released as 0.4.0 in some days after some more finalizations. The Changelog is already updated also for the WIP changes

PR #124, opened 2023-04-30
Implements ManualPairingCodeSchema.decodeInternal()

digitaldan:
Implements the missing decode functionality for pairing codes and allows passing in a code to the `ControllerNode.ts` class.  Tested on a Philips Hue bridge, using the pairing code generated in their IOS app.

### DIFF
--- a/packages/matter-node.js/README.md
+++ b/packages/matter-node.js/README.md
@@ -149,7 +149,7 @@ The following parameters are available:
 * -ip: the IP address of the device to commission
 * -discriminator: the discriminator to use for pairing (default: 3840)
 * -pin: the pin to use for pairing (default: 20202021)
-* -code: code to use for pairing (-discriminator and -pin will be ignored)
+* -pairingcode: code to use for pairing (-discriminator and -pin will be ignored)
 * -store: the storage location (directory) to use for storing the pairing information (default: controller-node). Delete the directory or provide an alternative name to reset the controller
 
 ## Modifying the server (DeviceNode) behavior
--- a/packages/matter-node.js/README.md
+++ b/packages/matter-node.js/README.md
@@ -149,6 +149,7 @@ The following parameters are available:
 * -ip: the IP address of the device to commission
 * -discriminator: the discriminator to use for pairing (default: 3840)
 * -pin: the pin to use for pairing (default: 20202021)
+* -code: code to use for pairing (-discriminator and -pin will be ignored)
 * -store: the storage location (directory) to use for storing the pairing information (default: controller-node). Delete the directory or provide an alternative name to reset the controller
 
 ## Modifying the server (DeviceNode) behavior
--- a/packages/matter-node.js/README.md
+++ b/packages/matter-node.js/README.md
@@ -114,7 +114,7 @@ npm run matter -- -on "echo 255 > /sys/class/leds/led1/brightness" -off "echo 0 
 
 The following parameters are available:
 * -passcode: the passcode to use for pairing (default: 20202021)
-* -discriminator: the discriminator to use for pairing (default: 3840)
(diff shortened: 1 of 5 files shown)


packages/matter-node.js/README.md
```diff
@@ -149,6 +149,7 @@ The following parameters are available:
 * -ip: the IP address of the device to commission
 * -discriminator: the discriminator to use for pairing (default: 3840)
 * -pin: the pin to use for pairing (default: 20202021)
+* -code: code to use for pairing (-discriminator and -pin will be ignored)
```
Apollon77, line 152: Thats the same as "passcode" on DeviceNode right? SHould we then also name it that way so that it is easier mappable?

Apollon77 (changes requested):
Thank you very much for your contriubution, thats cool.

Could you please also add a test for decoding? maybe decode a just encoded pairingcode to show that it is the same or such? So adding a decode test in https://github.com/project-chip/matter.js/blob/main/packages/matter.js/test/schema/PairingCodeSchemaTest.ts#L50.

Thank you!

packages/matter-node.js/README.md
Apollon77: Hehe no it is not the same ... but maybe call it better "pairingcode" like also in other places, ok?

digitaldan: > Could you please also add a test for decoding? 

Yeah, it occurred to me after i submitted this that a test would have probably been good.   will add one shortly, thanks!

digitaldan:
So, writing the tests was useful , it made me look back at the spec more closely. Would like a second opinion on this.

According to section § 5.1.4.1 Table 38, the manual pairing code only uses the 4 MSB of the discriminator value, so their instructions to encode use the full discriminator and then use only the 4 MSB out of it.  

For reference, the current ManualPairingData type looks like:
```
export type ManualPairingData = {
    discriminator: number,
    passcode: number,
    vendorId?: number,
    productId?: number,
};
```
So when we take the encoded string and decode it back, we will have a different discriminator value from what was in the ManualPairingData data type, so something like:

```
export type ManualPairingDataDecoded = {
    discriminator: number, // now only the 4 MSB , not the original value.
    passcode: number,
    vendorId?: number,
    productId?: number,
};
```

Not sure if we should have two different data types ( one that goes in, one that goes out) ?

digitaldan:
and i guess looking at this again, ManualPairingData should really only contain the 4 bit discriminator code so its the same going in and out, and we should have some other helper method to create a ManualPairingData  type from a more generic PairingData type ?

Apollon77:
Hm, ok I checked the code also where we use discriminator:
* The matter message to open a new commissioning window gets a "long discriminator" but limits value to 4095 (which is 12bits) and this long value is used for announcements
* The "Long" discriminatpr is then contained in _L<discriminator> record
* The Short one is used in _S record

Maybe then really decode to "shortDiscriminator"? It can still be used for the lookup we want to do with it next step.

PS: With this being said: Could you please add also a check to DeviceNode and ControllerNode if discriminator is >4095 then error and stop? ;-)

digitaldan:
> Maybe then really decode to "shortDiscriminator"? It can still be used for the lookup we want to do with it next step.

>PS: With this being said: Could you please add also a check to DeviceNode and ControllerNode if discriminator is >4095 then error and stop? ;-)

Sounds good, thanks!

Apollon77:
@digitaldan When you have udpdated the PR just poke me or re-request review

digitaldan:
PR is updated, let me know if this is what you were thinking.  Thanks!

Apollon77 (approved):

Apollon77:
Perfect that way, thank you! I just added the value info to readme too ..

PR #125, opened 2023-04-30
Correct TCP advertisement because not supported

Apollon77:
We do ot support TCP (and even when noone seems to use it at all), so we should also not advertise it.

PS: tested it with Apple afterwards ... no change in behavior


vves (changes requested):
Per 4.3.1 Commissionable Node Discovery and 4.3.4 Common TXT KV Pairs:

T should be 1 since the project default implementations are TCP based nodes.

Apollon77:
But we do not support TCP but UDP only ...  (and also TCP, even if announced, seems not to be used at all)

vves (approved):
Per our verbal discussion - Approved.

PR #129, opened 2023-05-03
Logging upgrades

lauckhart:
Streamlines and colorizes logs.  Hopefully not controversial.

Not enabled by default; to enable:

```typescript
import { Logger, Format } from "@project-chip/matter.js/log";
logger.format = Format.ANSI;
```

Ends up looking as follows:

image

Also includes HTML formatting, boosts logging test coverage to 100% and includes a minor housekeeping commit.


packages/matter.js/src/log/Logger.ts
```diff
@@ -16,58 +16,220 @@ export enum Level {
...
+        const entries: string[] = [];
+        this.entries.forEach((v, k) => entries.push(`${keyFormatter(k)}${formatValue(v, keyFormatter, valueFormatter)}`));
+        return entries.join(" ");
+    }
+
+    public toString() {
+        return this.serialize();
+    }
+
+    private entries: Map<string, any> = new Map();
```
Apollon77, line 169: nit:
```suggestion
    private entries = new Map<string, any>();
```

packages/matter.js/src/log/Logger.ts
```diff
@@ -16,58 +16,220 @@ export enum Level {
...
+ * Pass this dictionary type as a logging parameter to improve formatting of
+ * log output.  See Logger.dict() for maximal convenience.
+ */
+export class DiagnosticDictionary {
+    /**
+     * Create a new dictionary with optional entry values.
+     * 
+     * @param entries the entries as [ "KEY", value ] tuples
+     */
+    public constructor(entries: { [KEY: string]: any } = {}) {
```
Apollon77, line 132: nit: I think no need for "Public" before constructor

packages/matter.js/src/log/Logger.ts
Apollon77: And please ove up. We alwys have "class variables" directly at the start above constructor.

packages/matter-node.js/tsconfig.json
```diff
@@ -12,6 +12,7 @@
         "noUnusedParameters": true,
         "noUnusedLocals": true,
         "strict": true,
+        "resolveJsonModule": true,
```
Apollon77, line 15: Why you added this? Which json files are resolved?

Apollon77 (changes requested):
Thank you very much, this improves Logging a log. Please see my comments. What do you think, maybe by default use the ANSI Format in DeviceNode.ts and ControllerNode.ts because they are utilized as cli scripts

packages/matter-node.js/tsconfig.json
lauckhart: Didn't mean to leave that in.  I needed it for the following on newapi branch...  Not sure why it worked for you but not me, maybe node version (I'm on 20)?  Regardless, I can pull it out if you want.

```typescript
import packageJson from "../package.json"
```

Apollon77: Yes I will kill those again ... no idea why they came in again :-)

lauckhart:
Thanks, glad you like it.  Yeah, probably makes sense as CLI default.  Should probably be contingent on TTY check.  Eventually maybe something along the lines of:

```typescript
Logger.format = process.env.MATTER_LOG_FORMAT
    || config.log.format
    || (process.stdout.isTTY && Format.ANSI)
    || Format.PLAIN
```

I was interested earlier in factoring boilerplate out of DeviceNode/ControllerNode (and now NewDeviceNode etc.) in part so there'd be somewhere to stash stuff like this.

Apollon77:
@lauckhart yes that makes sense.

And for "factoring boilerplate out of DeviceNode/ControllerNode" ... I see these as "reference implementations" kind of and in a newxt PR I plan to move them into an "examples" folder, maybe it is easier to just "structure the code" with comment blocks to tell what does what so that devs can take out themself the relevant parts based on that "structuring"? That would allow not to putt too much effort in these things ;-)

if we once really want to have a "cli"-tooling or official scripts then I would more see a @project-chip/matter-cli.js package and structure it there so as needed for the goal of that cli thing

Apollon77:
@lauckhart DO you want tzo add the latter idea to DeviceNode and Controller? Or separate PR?

lauckhart:
@Apollon77 I'll do a followup PR

PR #134, opened 2023-05-04
Implement Scenes Handler Logic, Revamp Groups Handler

Apollon77:
This PR got a bit bigger then planned because it turned out that more is touched, but here we go.

This PR is basing (and because of this including) #121 and #131, so "Ignore" then for review; as soon as they are merged I can rebase/merge main in.

This PR adds the following:
* Implement a generic Scenes Cluster Server Handler (insluding storing and recalling a scene in/from storage and interactions with GroupsCluster)
* Revamp (and fix) some things from the Groups Cluster Server Handler (including interaction with IdentifyCluster and ScenesCluster)
* Implement for Scenes and Clusters an generic manager class that can then also be used based on the Fabric scoped storage to allow cluster interactions as defined by specs and can be used for cluster interactions
* Add a generic way to do Cluster Handler testing and implement tests for most cases for Groups and Scenes
* Rework the "dynamic getter" handling to have an data interface like the other ClusterCommandHandlers (object and not single parameters) and make attributeservers and endpoint data also available for them
* add a scene property for attribute definitions which is then used to select the scene relevant datapoints
* Add storage support for all matter.js datatypes because now with stored dynamic Attribute data they could be part of the data to store (and add testing)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,8 @@
+# Changelog for matter.js
+
+## 0.3.0 (2023-05-03)
+* Initial release of matter.js and matter-node.js packages after the code merge
+* From now on we will add a changelog for each release
+
+## < 0.3.0
+* Releases of matter.js with initial Logic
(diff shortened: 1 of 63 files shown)


vves (approved):
Looking forward to exercising scenes.  Approving first pass of Scenes in order to dig deeper as it's an optional cluster.

PR #135, opened 2023-05-05
Upgrade to Typescript 5 and use multiextend

Apollon77:
This PR upgrades the dependencies, mainly Typescript to v5 which enables to use multiextend. Said this (on ifea if @mfucci  the ts dev config is cutted out in one central file to have the "same" config best possible in one place.

I also added sourcemaps to be generated.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,22 +10,22 @@
                 "packages/matter-node.js"
             ],
             "devDependencies": {
-                "@typescript-eslint/eslint-plugin": "^5.54.0",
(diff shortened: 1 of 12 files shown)


Apollon77:
Jest do not support it right now ... see https://github.com/kulshekhar/ts-jest/issues/4122

Lets wait somem days if jest supports it soon ... else we can add a file for jest only

lauckhart:
Closed as we're on ts5 now

PR #136, opened 2023-05-05
Fix the array maximum size according to specs

Apollon77:
... because we have the first users running against it with big bridges


vves (approved):
Tested against Home with large bridge.

PR #137, opened 2023-05-06
Standardize environmental logic

lauckhart:
This is a proposal relating to the conversation from #129.  Goals:

- Streamline usage of the public API
- Create a central place for recommended initialization logic
- Do not limit flexibility (library users can still initialize themselves if they want)

Note that I implemented a new `StorageManager.get()` but StorageManager isn't a singleton.  The `get()` instance just allows for  default argument to the classes that utilize StorageManager.

I have a more ambitious branch that standardizes configuration but probably that's best to left to a separate conversation/PR.

### DIFF
--- a/packages/matter-node.js/src/ControllerNode.ts
+++ b/packages/matter-node.js/src/ControllerNode.ts
@@ -5,24 +5,8 @@
  * Copyright 2022-2023 Project CHIP Authors
  * SPDX-License-Identifier: Apache-2.0
  */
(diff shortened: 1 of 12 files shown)


Apollon77:
Hi @lauckhart, thank you very much for your contribution, lets discuss in Discord.

lauckhart:
Will open new PR against newapi if it makes sense after next round of updates

PR #143, opened 2023-05-13
Add generated API documentation

Apollon77:
Now with having an more or ess official API we should also provide API Docs to the users.

The idea would be to generate that and commit in GitHub with eas new release on npm.

It can be brwsed for this PR via https://github.com/Apollon77/matter.js/blob/typedoc/docs/README.md


turon:
I like it.  But should the docs be generated into dist and possibly hosted at GitHub.io, but not go into the main source repo?

Apollon77:
You mean generate to gitHub pages ... Then I need to check how to do that ....

turon:
Looks like even with GitHub pages typically the docs take space in the repo anyway in a site branch.  Approving.  We can shuffle things around later if desired.

turon (approved):

PR #144, opened 2023-05-14
Feedback request: Quick prototype of automatic device class implementation

lauckhart:
TL/DR a couple of places to look are `packages/matter.js/cluster/interface/ScenesInterface.ts` (auto-generated) and `packages/matter.js/cluster/interface/ClusterIfaceImpl.ts` (stub for automagical implementation... I put little effort into naming).

This isn't intended for merge, I just want to share an approach I've been tossing around r.e. implementing classes based on `device/Device`.  I'm still getting familiar with the API so I'm likely off base in some of my assumptions.  I apologize in advance for confusion on my part but hopefully there's some useful ideas here.

My main Q is whether this is worth pursuing in some form and whether it makes sense as part of the current API or some other abstraction.

This is related to but different from @turon's #53 as I've focused on the the high-level API.  I use the matter.js schema as "source of truth" which is somewhat tangential to how that schema is created.

The general idea is:

1. Using the cluster schema, generate a detailed high-level TypeScript interface for each cluster either at build or dev time,
2. Offer mixin classes that implement each interface automatically at runtime, and
3. Device classes then use automatic implementation for the boring parts and developers focus on adding convenience

I'm not 100% confident this approach will work but after implementing the prototype it seems likely.


Apollon77:
Hi,

thank you for sharing this approach. I must say i like the idea. That especially could simplify the generation a lot and especially simplifies adding all the methods automatically.

Here some comments/thorughs (some alredy go in detail but I came around that while looking at it and some are already formal details ;-):
* Formal: the "Generate" stuff should be put into a tools directory and such and ideally write the "generate.sh" as npm script(s) then it is platform agnostic.
* formally you do not need to "generate" all these types that you also generate, you can use TypeFromSchema<typeof TlvWhateverSchemaThingy>, so these types are all automagically mapped already. If you want better named types then a "export type Bla = TypeFromSchema<typeof TlvBla>" should also do the trick. This should make the generation maybe much more simple! We might just need to make the relevant ones being "exported" if not yet. But that then also could simplify development because types are then adjusted without the need to re-generate.
* When we generate it that detailed we should also enhance our cluster schemas by info if an attribute is "fixed" - aka static config .. then we disallow sets afterwards and also do not need Listeners. That should be easy to do and remove a lot of unneeded stuff from the generation results
* Maybe we really need tp generate a client and a server variant of the cluster. commands are only relevant when client. For clients we only need "set" if the attribute is writable ... and such
* The global Attributes (featuremap, attribute and such) we should also exclude from generation
* Formal: TlvNoArguments should translate to "nothing" instead request: any
* There are some Clusters (like Groups, Scenes, Binding and many more on root level) that I would not expose at all. They are completely internal and can organize all needed stuff automagically. No need to have the user "see" them, so we should have an exclude listfor the mixins to use

So yes I wopuld like to see how that behaves when really using it in a class and which kind of API would be the result.
If you like a next step then I propose start with the "server" part because I'm still unsure what "clusterClients" in a device will really are/be. In fact they would somehow relate to the Binding Cluster to get an assignment for endpoint and such ... but yes later. Clusters would be first

lauckhart:
One other point I forgot but I think is important...

An advantage of generating the interfaces is that we can add proper documentation for developers.  If developers use that as their primary API then it probably makes sense to transfer the rich documentation in the *Cluster classes.  The lowest friction approach may just be to parse the *Cluster.ts files and extract the comments from the AST.  The TS API makes that pretty easy.

Not sure how feasible this is if the interfaces aren't generated, will have to research that.

lauckhart:
@Apollon77 please see what you think now.  This is the cleanest API I could come up with.  Haven't fully automated cluster implementation yet.

I ended up hating the mixin syntax too much so I came up with a prettier approach with equivalent functionality.

```js
import { WindowCoveringDevice } from "@project-chip/device/standard";

class MyWindowCovering extends WindowCoveringDevice {
    // A window covering with interface for standard clusters
    // Will flesh out as example once clusters are functional
}

class MyWindowCovering extends WindowCoveringDevice.with(WindowCoveringDevice.options.Scenes) {
    // A window covering with interface for standard clusters + scenes cluster
}
```

I'll move on to finishing cluster implementation but LMK if anything seems too off to you.

lauckhart:
As discussed, here's my notes as I work through adjusting the interface and wiring up.

```js
// We agreed that server interface is the priority...  Here's a more detailed
// proposal on the API.
//
// Note 1: This is a conceptual interface that models what the current PR is
// generating (or will generate) for a single cluster.  I pulled it together
// into a single place for discussion purposes.
//
// Question 1. We just call this "Device" but it's really a "ServerDevice".
// In a controller we could potentially also have a "ClientDevice" that looks
// very similar but proxies over the fabric to a remote endpoint.  So should
// we consider making space in the API for a future Client/Server bifurcation
// or am I thinking about this wrong?
//
// Note 2. The concept of a "ClientDevice" wouldn't be just for controllers, it
// would also be for client clusters of a "ServerDevice".  I think.
export interface PumpDevice extends Device {
    // ATTRIBUTES
    //
    // My thinking: This is how the server conveys property values to
    // matter.js.  They represent "last known state".
    //
    // Note 3: We could instead do getOnOff() or get(Property.OnOff) etc.
    // Implementation would be the same.
    //
    // Question 2: Right now OnOffLightDevices "pushes" changes into the
    // ClusterServer.  This model would be more of a "pull" that matter.js
    // would use whenever it needs to read current attribute state.  My
    // thinking here is that matter.js is not actually the "source of truth".
    // It just needs an adapter layer it can use to find current values.  We
    // can then automate change detection and otherwise update ClusterServer
    // in the same way we do now.  Is this approach worth pursuing or should I
    // try to align more closely with current approach?
    readonly currentLevel: number | undefined;
    readonly minLevel?: number;
    readonly maxLevel?: number;
    readonly onOffTransitionTime?: number;
    readonly onLevel: number | undefined;
    readonly onTransitionTime?: number | undefined;
    readonly offTransitionTime?: number | undefined;
    readonly defaultMoveRate?: number | undefined;
    readonly options: number;
    readonly remainingTime?: number;
    readonly startUpCurrentLevel?: number | undefined;

    // This is how the device implementation tells matter.js that attribute
    // values are different.  Internal implementation would then read all
    // attribute values and push any changes into the appropriate
    // ClusterServer.
    //
    // Question 3: One reason I like this approach is I worry that notifying
    // matter.js for every change would get too chatty.  Right now changes
    // are coalesced in SubscriptionHandler before notifying subscribers so
    // I don't think this is an issue from a network perspective.  But there
    // are a lot of internal function calls for every attribute on every change
    // I don't think there's any immediate need for change but is it worth
    // considering for the shape of the public API for when there are hundreds
    // of endpoints and thousands of attributes on a low-power device?
    //
    // Note 4: The alternative approach would be to make the properties above
    // read/write.  Then this function is unnecessary.  Regardless, we still
    // free the user from worrying about which cluster an attribute is
    // associated with.
    triggerChange(): void;

    // This is how matter.js conveys a request to change values from the
    // fabric.  It would be wired to AttributeServers via InteractionServer
    // internally.
    //
    // Question 4: We discussed whether this should be asynchronous or not.
    // I'm thinking of use cases like bridges or even I/O in the case of a
    // local device implementation.  Right now we get the attribute into
    // internal storage and give a thumbs up.  But should this be asynchronous
    // so an attribute write fails if the "source of truth" rejects the change?
    //
    // Question 5: I worry about chattiness here too.  If we don't shape the
    // API like this, are we potentially forcing API consumers to handle
    // coalescing of values on their side?  For example, if a controller says
    // "turn this light to 70%, red, with a 5 second delay" am I correct in
    // surmising that a Hue bridge implementation would want to take receipt of
    // all three attribute changes simultaneously?
    set(values: { [ name: string ]: any } ): Promise<void>;

    // COMMANDS
    //
    // These are callbacks.  The internal device implementation wires these to
    // a CommandServer via CommandHandler.  This means a device implementer
    // just needs to override a function to implement the functionality.
    //
    // Note 5: We can offer default implementations for these functions that
    // try to make the user's life easier.  But until we do that we can just
    // install a default implementation that throws a "not supported" error.
    // Then the fancier implementation won't be a breaking API change, it'll
    // just be a smarter default.
    invokeMoveToLevel(request: MoveToLevelRequest): Promise<void>;
    invokeMove(request: MoveRequest): Promise<void>;
    invokeStep(request: StepRequest): Promise<void>;
    invokeStop(request: StopRequest): Promise<void>;
    invokeMoveToLevelWithOnOff(request: MoveToLevelWithOnOffRequest): Promise<void>;
    invokeMoveWithOnOff(request: MoveWithOnOffRequest): Promise<void>;
    invokeStepWithOnOff(request: StepWithOnOffRequest): Promise<void>;
    invokeStopWithOnOff(request: StopWithOnOffRequest): Promise<void>;

    // EVENTS
    //
    // These are invoked by the API consumer to notify matter.js of an event.
    //
    // Question 6: I haven't figured where to wire these yet.  Is this bit of
    // interaction server still in flight or did I just miss it?
    //
    // Note 6: I'm assuming these are "fire and forget" so no need for
    // async.  Probably good to validate that assumption.
    //
    // Note 7: Pumps levels don't apparently have events so I had to snag these
    // from a different cluster for illustration purposes.
    triggerWiredFaultChange(): void;
    triggerBatFaultChange(): void;
    triggerBatChargeFaultChange(): void;
}
```

packages/matter.js/src/cluster/interface/LevelControlInterface.ts
```diff
@@ -0,0 +1,46 @@
...
+    readonly onOffTransitionTime?: number;
+    readonly onLevel: number | undefined;
+    readonly onTransitionTime?: number | undefined;
+    readonly offTransitionTime?: number | undefined;
+    readonly defaultMoveRate?: number | undefined;
+    readonly options: number;
+    readonly remainingTime?: number;
+    readonly startUpCurrentLevel?: number | undefined;
+
+    invokeMoveToLevel(request: MoveToLevelRequest): Promise<void>;
```
Apollon77, line 34: The "invoke" should only fo into a "Client" object and correstponding "commandhandling callbacks" into Server

packages/matter.js/src/cluster/interface/OnOffLightingInterface.ts
```diff
@@ -0,0 +1,32 @@
...
+
+    invokeOff(): Promise<void>;
+    invokeOn(): Promise<void>;
+    invokeToggle(): Promise<void>;
+    invokeOffWithEffect(request: OffWithEffectRequest): Promise<void>;
+    invokeOnWithRecallGlobalScene(): Promise<void>;
+    invokeOnWithTimedOff(request: OnWithTimedOffRequest): Promise<void>;
+}
+
+export const OnOffLighting: ClusterInterface<Common, Common> = {
```
Apollon77, line 30: In fact the "Normal" OnOffDevice is missing here?!

packages/matter.js/src/crypto/Spake2p.ts
```diff
@@ -5,11 +5,17 @@
  */
 
 import BN from "bn.js";
-import { ec } from "elliptic";
 import { Crypto } from "./Crypto.js";
 import { ByteArray, Endian } from "../util/ByteArray.js";
 import { DataWriter } from "../util/DataWriter.js";
 
+// Lint complains about the next line but ts-node complains about importing
```
Apollon77, line 12: Ääähmm ... my eslint do not complain!

Apollon77 (commented):
@lauckhart I added some comments and tried to asswer your questions. I think main "todo" is still the attribute get/set secion :-)

packages/matter.js/InterfaceNotes.ts
```diff
@@ -0,0 +1,121 @@
...
+//
+// Note 1: This is a conceptual interface that models what the current PR is
+// generating (or will generate) for a single cluster.  I pulled it together
+// into a single place for discussion purposes.
+//
+// Question 1. We just call this "Device" but it's really a "ServerDevice".
+// In a controller we could potentially also have a "ClientDevice" that looks
+// very similar but proxies over the fabric to a remote endpoint.  So should
+// we consider making space in the API for a future Client/Server bifurcation
+// or am I thinking about this wrong?
```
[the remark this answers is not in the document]
lauckhart: Right, so my thinking was that developers would extend Device to implement server clusters and would see a separate device for bound clusters.  I know matter sticks them both on the same endpoint but I was thinking it'd be confusing if my light switch device implementation talks to itself like it's a light bulb. 🤔

Anyway makes sense they'd share a base class.  I only brought it up now because with the approach above we might want to make separate subclasses for local vs bound/controlled devices.

packages/matter.js/InterfaceNotes.ts
```diff
@@ -0,0 +1,121 @@
...
+    // considering for the shape of the public API for when there are hundreds
+    // of endpoints and thousands of attributes on a low-power device?
+    //
+    // Note 4: (no longer relevant)
+    set(changes: Partial<State>): void;
+
+    // This is a convenience form of set for updating a single attribute.
+    //
+    // Note 8: We could also generate a setter for every individual attribute.
+    // However, I worry that would encourage inefficient implementation.
```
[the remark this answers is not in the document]
lauckhart: If there are a lot of attribute changes then we'll trigger a *lot* of callbacks.  I'm always nervous about that as it can add up and slam you with latency.  A lot of my home IOT stuff takes forever to start up which is annoying.  I've got 60+ zwave switches plus various sensors and shades.

Don't want to optimize prematurely, I just didn't want to code into a corner with public API.  But I think I can implement to current approach and keep more efficient path open for the future if it becomes necessary.

I just finished wiring for the generated attributes.  I made state attributes writable and have each read/write percolate down into the cluster server.  I'll remove the write portion for fixed attributes once your PR with appropriate metadata gets merged.

packages/matter.js/InterfaceNotes.ts
```diff
@@ -0,0 +1,121 @@
...
+    //
+    // Note 9: Not sure how to validate value type in the interface so might
+    // need to do that programmatically.
+    set(name: keyof (State), value: any): void;
+
+    // This is how matter.js conveys a request to change values from the
+    // fabric.  It would be wired to AttributeServers via InteractionServer
+    // internally.
```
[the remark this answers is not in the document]
lauckhart: OK.  I'll see how that affects wiring.  Maybe will want separate interfaces for "configuration" vs "session" attributes.

Also, r.e. earlier discussion about the need to propagate and/or veto writes before commit...  That only applies to configuration attributes.  So might be another benefit to separate interfaces.

packages/matter.js/src/cluster/interface/BasicAdminCommissioningInterface.ts
```diff
@@ -0,0 +1,27 @@
...
+
+/*** THIS FILE IS GENERATED, DO NOT EDIT ***/
+
+import { BasicAdminCommissioningCluster, ClusterInterface } from "../index.js";
+import { TypeFromSchema } from "../../tlv/TlvSchema.js";
+
+type OpenCommissioningWindowRequest = TypeFromSchema<typeof BasicAdminCommissioningCluster.commands.openCommissioningWindow.requestSchema>;
+type OpenBasicCommissioningWindowRequest = TypeFromSchema<typeof BasicAdminCommissioningCluster.commands.openBasicCommissioningWindow.requestSchema>;
+
+export interface Common {
```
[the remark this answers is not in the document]
lauckhart: Yeah I didn't actually mean to export those so was hoping it didn't matter.  But then vs code showed types like AutoDevice<Server, Server, Server, Server>.  So I stuck them into a module, now it looks more reasonable.

packages/matter.js/src/cluster/interface/LevelControlInterface.ts
lauckhart: Resolved on discord, now generated as "onCommand" for server and "sendCommand" for client.  Went back to "send" for client, less matterish but saves two characters.

packages/matter.js/src/cluster/interface/OnOffLightingInterface.ts
lauckhart: Yeah I merged in the ClusterExtend stuff after I started.  It caused problems because I was keying off the cluster ID.  I think I resolved it but we may want some additional metadata...  Right now Pump offers OnOffLighting as an option which I assume is not appropriate.

packages/matter.js/src/crypto/Spake2p.ts
lauckhart: So in order to run the code gen without precompiling the project I wanted to run via ts-node.  Works pretty well after some fiddling but with elliptic in ES6 module mode (which is necessary for other reasons) ts-node complains that there is no named export in the library.  So I changed to a default import.  But then lint failed insisting that I should be using a named import.  Finally gave up and resorted to disabling lint.  I think I just pulled a muscle rolling my eyes.

packages/matter.js/InterfaceNotes.ts
```diff
@@ -0,0 +1,121 @@
...
+    set(changes: Partial<State>): void;
+
+    // This is a convenience form of set for updating a single attribute.
+    //
+    // Note 8: We could also generate a setter for every individual attribute.
+    // However, I worry that would encourage inefficient implementation.
+    //
+    // Note 9: Not sure how to validate value type in the interface so might
+    // need to do that programmatically.
+    set(name: keyof (State), value: any): void;
```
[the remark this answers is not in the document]
lauckhart: I went ahead and pulled this out per earlier comment.

packages/matter.js/InterfaceNotes.ts
```diff
@@ -0,0 +1,121 @@
...
+    // ClusterServer.
+    //
+    // Question 3: One reason I like this approach is I worry that notifying
+    // matter.js for every change would get too chatty.  Right now changes
+    // are coalesced in SubscriptionHandler before notifying subscribers so
+    // I don't think this is an issue from a network perspective.  But there
+    // are a lot of internal function calls for every attribute on every change
+    // I don't think there's any immediate need for change but is it worth
+    // considering for the shape of the public API for when there are hundreds
+    // of endpoints and thousands of attributes on a low-power device?
```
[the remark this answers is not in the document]
lauckhart: Yeah, seems like there are a bunch of classes of attributes.  To test my understanding, here's the buckets I see:

# Session (volatile, writable by operator client)
# Configuration (non-volatile, writable by manager or administrator client)
# Reporting (writable by device)
# Fixed (writable at startup)

Fortunately all pretty well formalized in the spec.

packages/matter.js/InterfaceNotes.ts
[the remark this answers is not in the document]
Apollon77: > separate subclasses for local vs bound/controlled devices.

You will know that only on runtime, so separate classes might not be possible

packages/matter.js/InterfaceNotes.ts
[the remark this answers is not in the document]
Apollon77: In fact to translate that to matter.js because we use other terminology and in some cases do not care if config or operational
* Attributes - can be get/set locally, remote only get, may be persisted
* WriteableAttribute - get/set locally AND from remote, may be persisted
* FixedAttributes - defined in default, local and remote get-only, never persisted
* FabricScopedWritableAttribute - get/set locally when knowing fabric AND from remote (fabric via session), always persisted

And yes somewhen we also add ACL checking and then we limit who can write what

packages/matter.js/InterfaceNotes.ts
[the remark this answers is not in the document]
Apollon77: But in fact it only affects wiring locally by adding afabric parametrr unless someone finds a better idea. The question is also if it will be really an "local writing topic" for these. will wait for use cases where devs have issues with that approach :) But there will also be a way to get Fabrics - then it is "just choosing" if ther eis more then one

packages/matter.js/InterfaceNotes.ts
```diff
@@ -0,0 +1,121 @@
...
+    set(name: keyof (State), value: any): void;
+
+    // This is how matter.js conveys a request to change values from the
+    // fabric.  It would be wired to AttributeServers via InteractionServer
+    // internally.
+    //
+    // Question 4: We discussed whether this should be asynchronous or not.
+    // I'm thinking of use cases like bridges or even I/O in the case of a
+    // local device implementation.  But should this be asynchronous so an
+    // attribute write fails if the "source of truth" rejects the change?
```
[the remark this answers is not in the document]
Apollon77: for completeness: discussed in discord

packages/matter.js/InterfaceNotes.ts
```diff
@@ -0,0 +1,121 @@
...
+    //
+    // Question 5: I worry about chattiness here too.  If we don't shape the
+    // API like this, are we potentially forcing API consumers to handle
+    // coalescing of values on their side?  For example, if a controller says
+    // "turn this light to 70%, red, with a 5 second delay" am I correct in
+    // surmising that a Hue bridge implementation would want to take receipt of
+    // all three attribute changes simultaneously?
```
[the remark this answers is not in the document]
Apollon77: discussed in discord

packages/matter.js/InterfaceNotes.ts
```diff
@@ -0,0 +1,121 @@
...
+    //
+    // These are callbacks.  The internal device implementation wires these to
+    // a CommandServer via CommandHandler.  This means a device implementer
+    // just needs to override a function to implement the functionality.
+    //
+    // Note 5: We can offer default implementations for these functions that
+    // try to make the user's life easier.  But until we do that we can just
+    // install a default implementation that throws a "not supported" error.
+    // Then the fancier implementation won't be a breaking API change, it'll
+    // just be a smarter default.
```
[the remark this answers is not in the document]
Apollon77: Discussed in Discord.
alternative is also named handler and a "addCOmmandHandler(name, handler)" interface

packages/matter.js/InterfaceNotes.ts
```diff
@@ -0,0 +1,121 @@
...
+    invokeMoveWithOnOff(request: MoveWithOnOffRequest): Promise<void>;
+    invokeStepWithOnOff(request: StepWithOnOffRequest): Promise<void>;
+    invokeStopWithOnOff(request: StopWithOnOffRequest): Promise<void>;
+
+    // EVENTS
+    //
+    // These are invoked by the API consumer to notify matter.js of an event.
+    //
+    // Question 6: I haven't figured where to wire these yet.  Is this bit of
+    // interaction server still in flight or did I just miss it?
```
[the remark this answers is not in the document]
Apollon77: Ok, I had a look into events ... will be interesting. 

My current thinking is that on "new ClusterServer" we enhance it by a "supportedCommands" object where you define which events you support aka "your code can generate". then clusterServer instance (?) will have a "triggerEventXY" method or we have an EventManager or such internally.

On Client side events will be part or read, so there will be "getXYEvent" methods I assume on clusterclient and an option to add it to a subscribe.
Events alwys might return a list of events happened since last "get" and events of same name could come in that list ...

so for API: TriggerEvent methods to generate an event and for clients "eventListeners"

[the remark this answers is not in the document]
Apollon77: I will add "put in top region" for after vacation because will have bigger impact on API so we should do soon

packages/matter.js/InterfaceNotes.ts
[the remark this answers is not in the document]
Apollon77: On the other hand I thought about the idea of "devs extend generated classes to implement logic" ... When thinking about bridge builder I do not think that this is realistic because in fact:
You have an e.g, zwave devce, so based on it's featurees you search fr the correct of our classes ... but these can be maaaany ... so maybe it is better to always hae a handler callback approach?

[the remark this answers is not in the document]
lauckhart: Hmm right, I've been keeping that use case in back of mind but haven't focused too much on it.  Note that there's not (I don't think) any functional difference between creating a JS class and creating an instance with handlers.  So the current stuff would work fine...  But we'd maybe want a couple of utility methods to make usage more natural.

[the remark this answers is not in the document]
Apollon77: PS: With explicit handler registration it also migth be more easy to detect if an "own handler" was implemented ... or how can we get that info else?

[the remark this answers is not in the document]
lauckhart: Would maybe depend on context...  Are you thinking to populate FeatureMap during discovery or to determine whether to use default implementation?  Or other use cases?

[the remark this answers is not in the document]
Apollon77: Wether to use default implementation or a custom command handler.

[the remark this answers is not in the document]
lauckhart: OK.  So I'm actually wiring ClusterServer in right now.  Not done yet but this is roughly where I'm headed.  WDYT?

- For each (server) cluster:
    - Search cluster exports for default handlers
    - For each attribute:
        - Create `onAttributeChange` in the device class that does nothing by default
        - Select an initial value:
            - Search `state.attrName` in user-supplied initial state
            - Fall back to TLV schema for default value
        - If there's a `getAttribute` in default handlers, add it to handler list
    - For each command:
        - Create `onCommand` in the derived class
            - If we found a handler above, `onCommand` calls that
            - If not, `onCommand` throws `Error("Not implemented")`
        - Add handler `CommandServer` that invokes `onCommand`
    - For each event:
        - Create `triggerEvent` in the derived class
            - `triggerEvent` has a TODO and throws `Error("Not implemented")`
    - Create a `ClusterServer`
        - `clusterDef` = the Cluster
        - `attributeInitialValues` = defaults collected above
        - `handlers` = handlers created above

So to answer your question, default command handlers are called unless the user overrides `onCommand` and doesn't call `super()`.  There's not a way to override attribute getters, I'm not sure that's necessary.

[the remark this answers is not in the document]
Apollon77: Re onAttributeChange
ok, and constructire adds these methods then automatically as listeners? thats ok. Then user can still add another one OR overwrite

Re onCommand:
> If we found a handler above, onCommand calls that
it needs to work differently. handlers are used deeply internal where devcie classes are not known. Unless we want to change that (what I would not think is a good idea). In my eyes the default handlers need to know if there is a "more high level provided implementation" and then not execut his logic but the other. Thats why I did that "namedHandler approach and idea was top simply (other then now) pass the named handler into the "getDefaultHandlers" methods because then we would have the logic here like

`handler.hashandlerFor("identify") -> call that, else do my onwn minimalistic stuff.

Or how you planned to wire it that the "handlers existing on ClusterServer" (and those are used internally to handle commands) know when what to do?

So the more I think the more I would go with the "addCommandHandler("name", () func)" approach) to "Override command handlers" and then auto generate the allowed name to methiod signature list for that named handler

Re: triggerEvent:
ok, makes sense. will wire then later to a eventServer - already have it in my mind :-)

Re: "Create a clusterServer"
Here we need to think. I planned it now with this "createDefaultClusterrServer" to have anything for a meaning default - and in fact users can override defaults. I would love to stay with this.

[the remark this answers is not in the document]
Apollon77: Re "initial attribute value": simly get it from attribute.get() ?

packages/matter.js/src/cluster/Cluster.ts
```diff
@@ -176,5 +180,6 @@ export const ClusterExtend =
             attributes: Merge(attributes, attributesExtend),
             commands: Merge(commands, commandsExtend),
             events: Merge(events, eventsExtend),
+            extension: true,
```
Apollon77, line 183: why this makes a difference? Alao normal "Cluster" can be used nor ally. I would really just use if it is exported it is externally available as a cluster variant": irrelevant if CLuster or ClusterExtend

packages/matter.js/src/cluster/ClusterInterface.ts
```diff
@@ -0,0 +1,13 @@
+import { Cluster, Attributes, Commands, Events } from "./Cluster.js";
+
+/**
+ * Instances of this class tie together the public interface types and the
+ * definition for a specific cluster.  They are generated in the cluster
+ * interface file and consumed in standard device implementations.
+ */
+export type ClusterInterface<State, Client, Server> = {
```
Apollon77, line 8: should we use the same notation like in other places with 

ST extends State, CL extends Client, SRV extends Server

and then use these below?

packages/matter.js/src/crypto/Spake2p.ts
Apollon77: Provde "the"/"a" tsconfig? Because there is one tsconfig option for exactly that topic ... maybe we just need to use it

packages/matter.js/src/device/AutoDevice.ts
```diff
@@ -0,0 +1,361 @@
...
+
+        // Getter
+        const descriptor = <PropertyDescriptor>{
+            get(this: BaseState) {
+                return this[DEVICE].attributeFor(cluster, attr).get();
+            }
+        };
+
+        // Setter
+        if (attr.fixed) {
```
Apollon77, line 148: should be "not fixed" or? :-)

packages/matter.js/src/device/AutoDevice.ts
```diff
@@ -0,0 +1,361 @@
...
+                return true;
+            }
+        } else {
+            descriptor.writable = false;
+        }
+
+        // Add the property
+        Object.defineProperty(this.stateProto, attr.name, descriptor);
+
+        // If there's a default "get" handler, install it
```
Apollon77, line 160: please let the attribute server being "the main place"  ... simply always use attribjute.getLocal ... I would not interfere that ddeep here because else it will be issues with changes later (and not only I might kill getters soon)

packages/matter.js/src/device/AutoDevice.ts
```diff
@@ -0,0 +1,361 @@
...
+    }
+
+    // Implement the state interface
+    private wireAttributeState(attr: AttributeModel) {
+        const cluster = this.cluster;
+
+        // Getter
+        const descriptor = <PropertyDescriptor>{
+            get(this: BaseState) {
+                return this[DEVICE].attributeFor(cluster, attr).get();
```
Apollon77, line 143: getLocal()

packages/matter.js/src/device/AutoDevice.ts
```diff
@@ -0,0 +1,361 @@
...
+            });
+        });
+    }
+
+    // Implement device commands
+    private wireCommand(command: CommandModel) {
+        this.deviceProto[command.invoker] = ignore;
+        this.serverHandlers[command.handler] =
+            async (request: any, _session, _message, endpoint: any) =>
+                await endpoint[command.handler](request);
```
Apollon77, line 183: see above - this ill not work that way unless we change the whole InteractionServer logic. Commands are triggered from remote, so the Device class normally plays no role into it. The only thing is to make sure that the registered ClusterServer handlers know that there is a better implementation "upstairs" :-)

packages/matter.js/InterfaceNotes.ts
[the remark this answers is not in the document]
Apollon77: Re commands: AHhhh ... read below (wireCLusterServer)  first ....

packages/matter.js/src/device/AutoDevice.ts
```diff
@@ -0,0 +1,361 @@
...
+
+    // Wire cluster server installation
+    private wireClusterServer() {
+        const cluster = this.cluster;
+        const handlers = this.serverHandlers;
+        this.initializers.push(function(this: AutoDevice, options: AutoDeviceOptions<BaseState>) {
+            this.addClusterServer(ClusterServer(
+                cluster.definition,
+                Object.assign({}, cluster.defaultState, options.state || {}),
+                handlers));
```
Apollon77, line 200: Ok now I think I see your approach ...

my current approach was that for many commands we can do a default implementation and it is then very likely that devs do not need to do anything. best example is onWithTimedOff or such on OnOffLighing cluster. and the whole Scenes, Groups and such. So my idea was that we have a defualt implementation that "can check" if there is a better one from the dev, else we use the default. my question here is ... how we could realize this with your approach? Mainly: Where should we put these default and how to wire them? Else we would nee a "getDefualtHandlerFor ClusterrXY" and then the logic here?

packages/matter.js/src/device/AutoDevice.ts
```diff
@@ -0,0 +1,361 @@
...
+        this.set(changes);
+    }
+
+    static implement<Interfaces extends ClusterInterface<any, any, any>[]>(definition: DeviceTypeDefinition, ...interfaces: Interfaces) {
+        return this.extendWithDefinition(this, definition, ...interfaces);
+    }
+
+    serverFor(cluster: ClusterModel) {
+        const server = this.getClusterServer(cluster.definition);
+        if (server === undefined) {
```
Apollon77, line 261: why not return undefined?

packages/matter.js/src/device/CodeModel.ts
```diff
@@ -0,0 +1,232 @@
...
+
+export type UntypedHandlers = { [key: string]: (...any: any[]) => any };
+type UntypedHandlerFactory = () => UntypedHandlers;
+
+function findDefaultHandlers(cluster: ClusterModel): UntypedHandlers {
+    let handlers = <UntypedHandlerFactory>(<any>clusterExports)[`${cluster.name}ClusterHandler`];
+    if (handlers && typeof handlers == "function") return handlers();
+
+    // Some of the handlers don't have "Cluster" in their name
+    handlers = <UntypedHandlerFactory>(<any>clusterExports)[`${cluster.name}Handler`];
```
Apollon77, line 122: then we should streamline this and name them correctly

packages/matter.js/src/device/AutoDevice.ts
Apollon77: PS: Ok in this.defaultServerHandlers you know the default ones, but they are not considered here

packages/matter.js/src/device/CodeModel.ts
```diff
@@ -0,0 +1,232 @@
...
+
+    return {};
+}
+
+function defaultState(cluster: ClusterModel) {
+    const result = <{ [key: string]: any }>{};
+
+    cluster.attributes.forEach((attr) => {
+        if (attr.default !== undefined) {
+            result[attr.name] = attr.default;
```
Apollon77, line 133: see above ... ideally do not use that ... just use "getLocal" ... Reason: There is also maybe storage involved and so the initial value and default value are not relevant at all

packages/matter.js/src/crypto/Spake2p.ts
lauckhart: I can revisit.  I did try the various tsconfigs.  If I remember correctly the ones that put it into CJS mode didn't work for other reasons.  But you do have other places using ts-node?

packages/matter.js/src/cluster/Cluster.ts
lauckhart: I needed this because in DeviceTypes the cluster is only referenced by ID but the ID is shared across the extensions.  So we need to know what is the "required" cluster and what is optional optional extension.  But this does cause other problems, see my comment r.e. pumps offering "OnOffLighting" as an option.

Maybe we could make DeviceTypes reference the cluster definitions directly rather than the cluster ID?

packages/matter.js/src/cluster/ClusterInterface.ts
lauckhart: Sure, will change.  I know cryptic generic parameters are pretty standard.  As a relative newcomer to the library I feel obliged to mention that the numerous appearances of "F, SF, A, C, E" seemed opaque and took a bit to understand. 😉  I'll try to ameliorate using little comments.

packages/matter.js/src/device/AutoDevice.ts
lauckhart: Yeah of course.  I was planning to start on tests today.  Waited longer than normal because things were changing so fast.

packages/matter.js/src/device/AutoDevice.ts
lauckhart: OK cool, I wasn't sure what the get handlers were for anyway. I just added it because there were some defined in some of the cluster servers.

packages/matter.js/src/device/AutoDevice.ts
lauckhart: Right, mistake

packages/matter.js/src/device/AutoDevice.ts
lauckhart: D'oh I totally botched that implementation.

packages/matter.js/InterfaceNotes.ts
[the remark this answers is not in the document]
lauckhart: Right sorry I threw that out quickly but I know I'm kind of violating encapsulation with how I did handlers.  I should have added more explanation.  I'll respond to subsequent comments.

packages/matter.js/src/device/AutoDevice.ts
lauckhart: I agree with your idea.  It would've been clearer if I had finished command wiring but ran out of wifely patience IRL.

If we can leave aside the implementation for a second, here's the shape of the command API that I think would be lowest-friction for the user.  Note that I added one additional feature - the ability for a user handler to fall back to the default handler.

```js
class MyLight extends OnOffLightDevice {
    onOff() {
        if (isNighttimeHours())
            throw new Error("Leave the lights on it's scary out there");
        super();
    }

    // onOn not implemented - just default handler is fine

    onToggle() {
        if (this.state.onOff) {
            // Default implementation doesn't call onOn/onOff
            this.onOff();
            return;
        }
        super();
    }
}
```

I like this because it builds on native language constructs so the syntax is concise and hopefully intuitive.  But it's also powerful in that it gives a lot of flexibility with extension points.

WDYT?  There's a number of ways to implement but probably makes sense to agree if this is the goal first.

...

Now, r.e. implementation...  We could have the device inspect the prototype chain and only add handlers for commands that have been overridden.  That was actually my original plan.  It would make the implementation more complicated (OK for the right API) but the real downside is that invoking super() wouldn't work.

So the solution I came up with was to "proxy" to the real handlers through the device class.  I'm not sure this would work but I looked pretty carefully through the existing handlers and though it might have a chance.

Another idea would be to add a "pre-invoke" handler into CommandServer that could tell the CommandServer whether to invoke the default handler.

Note 1. I had originally taken out scenes and groups from the public API and we could do that again.  I just added back to see what it looks like with them implemented this way.

Note 2. I lean toward making "too many" extension points available rather than too few, so long as they're invisible unless the user wants them.  So I might add interfaces to the device API even if they're completely handled by matter.js.  Can't count the number of times I've had to monkey patch some library because I needed something the author had decided was internal.

packages/matter.js/src/device/AutoDevice.ts
lauckhart: This is just a type safe utility to access the cluster server when it should already be defined.

packages/matter.js/src/device/CodeModel.ts
lauckhart: Cool, will do.  I've been trying to avoid making breaking changes in the public API.

packages/matter.js/src/device/CodeModel.ts
lauckhart: I did this because the handlers have createDefaultFoo with default values in them.  Other places only have the bare handlers.  In order to make it work everywhere I targeted the least-common denominator.

Maybe there could be a generic `createDefaultClusterServer(cluster, attributeInitialValues, handlers)` that does the right thing?  Might be able to slim down the server files if we set the defaults from the schema.

I looked at a handful of the createDefaultFoo, maybe they could all be replaced by this?  And then use FooClusterHandler for further specialization since it's already a factory?

packages/matter.js/InterfaceNotes.ts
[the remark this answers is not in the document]
lauckhart: Going to resolve this conversation as it's split out in separate comments around the code now which is nice.

packages/matter.js/src/device/AutoDevice.ts
lauckhart: OK I pushed the implementation that I intended.  (I think, still untested.)  See the new wireCommand

I also forgot to add the empty callback for attribute change events.  That's in wireAttributeState

packages/matter.js/src/device/AutoDevice.ts
lauckhart: Another point - the approach I took allows us to support any cluster even if the server isn't implemented for it yet.  If there was a generic `createDefaultClusterServer` it could do the same thing.

It could also add "Not implemented" handlers for commands that don't have a handler.  Then I could take that logic out of "AutoDevice".

Which reminds me, maybe need a better name for "AutoDevice".

packages/matter.js/src/cluster/Cluster.ts
Apollon77: This do not work because it just says "it hast cluster XY",  but not neededly "which feature variant of it". In fact even more bad (see device library specs 5.3.5.2 ... 
But in fact the Pump I would more say it is OnOffWithLevelControl, but I would not add the Lighing one. And when we are rellay correctwe also need special OnOff Handlers ... no idea how we can manage this compexity :-(

packages/matter.js/src/crypto/Spake2p.ts
Apollon77: I think we only have tsnode on matter-node level ... and in testing (jest/karma). could it be you need 

"esModuleInterop": true

??

packages/matter.js/src/device/AutoDevice.ts
Apollon77: These were for "attributes where values could be more easy generated dynamically then statically" ... but as written in discord this has big issues in subscriptions. I currently work on killing them. PR will come soon,

packages/matter.js/src/device/AutoDevice.ts
Apollon77: Oookk, Re concept:
I love that idea in general and for a developer that develops "a special device" this is the most convenient way and I like it.

But when I think that I'm a bridge builder that wants to "bind" several different devicetypes to an other kind of structure (like zwave, zigbee, or myself with iobroker) the idea of inheritence will not help me and so this idea (alone) very much limiting these kind of devs.
The reality right now is that more devs work on/plan bridges is giving this topic a big prio.
For them we would more need a handler approach that I can implement my binding logic once and just "add the handler" to the device class.

With reading the implementation option then they are all not that nice or very complicated or limiting ... Is it that worth "just" for "the most ideal interface?... i'm unsure
A simple "handler approach" (do nmot get me wring, I'm not fighting for it, but it wopuld be a more simple alternative) could still allow devs to implement own methods on the classes and bind them as handlers in the constructor.

lauckhart: Also, for completeness:  What do you think about this builder API for bridges?  It would work exactly the same except there would be static methods for installing handlers.

This is really just syntactical sugar for installing handlers.

```js
let device = AutoDevice.implement(chooseDeviceType(zwaveNode));
if (zwaveNode.commandClasses.SwitchBinary) {
    device = device.with(OnOff)
        .onOnOffChange((value) => { /* update z-wave device */ });
    // Map events from zwave -> OnOffInterface
}
if (zwaveDevice.commandClasses.SwitchMultilevel) {
    device = device.with(LevelControl)
        .onLevelChange(/* ... */);
    // ...
}
// Etc.
```

packages/matter.js/src/device/CodeModel.ts
Apollon77: I also had such an idea already

packages/matter.js/src/device/AutoDevice.ts
```diff
@@ -87,7 +87,10 @@ export interface StateManager<State extends BaseState> {
 }
 
 type Constructor = new (...args: any[]) => any;
-type UntypedClusterServer = <() => >;
+
+// TODO - should this be properly typed and added to the public API?  I'm
```
Apollon77, line 91: Yes should be typed ... but also "request" missing here or?

packages/matter.js/src/device/AutoDevice.ts
```diff
@@ -177,10 +185,24 @@ class WiringContext {
 
     // Implement device commands
     private wireCommand(command: CommandModel) {
-        this.deviceProto[command.invoker] = ignore;
+        // Install the default handler
+        const defaultHandler = this.cluster.defaultServerHandlers[command.handler];
+        if (defaultHandler) {
+            // TODO - this may not work or may otherwise change pending PR
+            // conversations
+            this.deviceProto[command.invoker] = function(request: any, { session, message, endpoint }: CallContext) {
```
Apollon77, line 193: and we need to get the typing right ... Re "any"

lauckhart: This is implementing the interface method...  The type is correct in the public signature.

packages/matter.js/src/device/AutoDevice.ts
Apollon77: Ok a bit abstract. It i think I get the idea. Could work. I think we need that in an example and in the end would throw against @vves in first place  ;-))

packages/matter.js/src/cluster/Cluster.ts
lauckhart: OK, it's only a problem now in the sense that unnecessary features are in the "options" object.  We don't add them by default because we don't know if they're necessary.  So this just puts it on the user.  They presumably know what features they need so I think it's all good.

packages/matter.js/src/device/AutoDevice.ts
lauckhart: 👍  I think there's 3 API shapes we're considering.  I'll call them "big constructor", "big body" and "builder".  We can default to "big constructor" if the other two don't work out.  They aren't mutually exclusive so we can always start with easiest and get fancier later.

packages/matter.js/src/device/AutoDevice.ts
lauckhart: The request body I had as a separate argument.  This is just internal implementation detail at the moment, there's proper typing in the public API.  I'll move development to a new branch so I don't slam you with updates every time I push a WIP. 😉

lauckhart:
Documenting three possible API shapes in top-level comment.

**Big body**

```js
class MyLight extends OnOffLightDevice.with(OnOffLightDevice.options.LevelControl) {
    onOnOffChange(value) {
        console.log(`On: ${value}`);
    },

    onToggle(...args) {
        console.log("Toggled!");
        super();
    }
}
```

**Big constructor**

```js
class MyLight extends OnOffLightDevice.with(OnOffLightDevice.options.LevelControl) {
    constructor() {
        this.addHandlers({
            onOnOffChange: (value) =>
                console.log(`On: ${value}`),

            onToggle: (request, context) => {
                console.log("Toggled!");
                context.super(); // Just speculating here
            }
        });
    }
}
```

**Builder**

```js
const MyLight = OnOffLightDevice.with(OnOffLightDevice.options.LevelControl)
    .onOnOffChange((value) =>
        console.log(`On: ${value}`);
    ).onToggle((request, context) => {
        console.log("Toggled!");
        context.super();
    }
}
```

Note 1. Builder alternative could also have `.addHandlers()` form
Note 2. Question of `getOnOff()` vs `state.onOff` still open but either/both pretty easy

packages/matter.js/src/cluster/Cluster.ts
Apollon77: In fact users need to know which featureset they want - for the "device builders"

lauckhart:
Closing this, will open new draft PR soon.

PR #146, opened 2023-05-14
Remove the exposed legacy API

Apollon77:
The PR removes the exposed parts of the legacy API. Some classes still exist internally but will be replaced in next time


vves (commented):
Thoughts on deprecating first?  I don't believe there are many if any devs using this library that can't deal with this breaking change, but wanted to bring it up.

Apollon77:
I started having issues with upcoming changes where I always needed to tweak stuff for this legacy classes in a special way or need to start refactoring it to also e.g. use endpoint class as new basis. Because of that I think it is easier to do it now and clean. We have released 0.3.0 before newapi merge. I could also release a 0.4.0 which is already breaking in some ways now and we deprecate it with this. Then we release soon afterwards a 0.5.0 and merge this for this version. 

Otherwise we simply remove the legacy already with 0.4.0 . As you say I assume that just some
Devs use that old api and need to migrate anyway.

Removing it from officially exported allows me to start refactoring it internally.

vves (approved):

PR #152, opened 2023-05-18
Rework/Enhance Attribute Typings and Logic

Apollon77:
This PR cleans up and reworks a lot of things around Attribute types and AttributeServer:

* Matter-Core functionality:
  * Breaking: Added support to allow to clearly model some more attribute types with internally different behaviour:
    * Added types for WritableFabricScopedAttribute and OptionalWritableFabricScopedAttribute
    * Added types for FixedAttribute and OptionalFixedAttribute
    * Added FixedAttributeServer which only allows to "get" the value but not to set it
    * Added FabricScopedAttributeServer which gets and sets the value based on the provided fabric
    * Updated ClusterServerObj and ClusterClientObj typings to respect these Attribute types
    * Updated all Cluster definitions that use such attribute types
  * Enhance: Splitted up and corrected PowerSource Cluster based on Matter 1.1 Specs
  * Fix Typing of Commands in ClusterClient if no commands were present
  * Fix: Fix equality checks in Attribute servers to check deeper then just === (and introduce new util method isDeepEqual)
* matter.js API:
  * Breaking: Remove the Legacy API classes from examples and also do not export them anymore. use the new API please!
  * Breaking: Adjusted some constructors of the new API and remove the option to pass an array of clusters to be added initially because this was no longer compatible to the strong typing in some places. Use addClusterServer and addClusterClient methods
* Misc:
  * Added Specification links for Matter Specifications 1.1


packages/matter.js/src/device/Device.ts
```diff
@@ -33,7 +33,15 @@ export class PairedDevice extends Endpoint {
         clusters: (ClusterServerObj<Attributes, Commands> | ClusterClientObj<Attributes, Commands>)[] = [],
         endpointId: number
     ) {
-        super(definition, clusters, endpointId);
+        super(definition, endpointId);
+        clusters.forEach(cluster => {
+            if (isClusterServer(cluster)) {
+                this.addClusterServer(cluster);
+            } else if (isClusterClient(cluster)) {
```
lauckhart, line 40: I think the second "if" here is unnecessary, TS knows it's a client if it's not a server since there are only two options

packages/matter.js/src/device/Device.ts
```diff
@@ -33,7 +33,15 @@ export class PairedDevice extends Endpoint {
         clusters: (ClusterServerObj<Attributes, Commands> | ClusterClientObj<Attributes, Commands>)[] = [],
         endpointId: number
     ) {
-        super(definition, clusters, endpointId);
+        super(definition, endpointId);
+        clusters.forEach(cluster => {
+            if (isClusterServer(cluster)) {
```
lauckhart, line 38: Just an aside but might be cleaner to have "isClient = true; isServer = false;" on the client and inverse on server.  Then TS will use as a discriminator so no need to import is functions

lauckhart (approved):
LGTM.  Made a couple of trivial comments

packages/matter.js/src/device/Device.ts
Apollon77: Yes I know ... knows because of typing .. the security guy in me added it ;-)

packages/matter.js/src/device/Device.ts
Apollon77: In fact the functions are type hints, but yes additional such attributes would also have worked

vves (approved):

PR #154, opened 2023-05-19
Enhance Storage system to allow to create subcontext stores

Apollon77:
This PR enhances the storage system by allowing sub contexts ending in an hierarchical storage key system.

This is the preparation to alalow multi node structures in storage and will also be used to simplify the Fabrics object again because the stored attribute data and such can be stored in a subcontext.


packages/matter-node.js/src/storage/StorageBackendDisk.ts
```diff
@@ -27,19 +27,21 @@ export class StorageBackendDisk implements Storage {
         // nothing to do
     }
 
-    buildStorageKey(context: string, key: string): string {
-        return context + "." + key;
+    buildStorageKey(contexts: string[], key: string): string {
+        const contextKey = contexts.join('.');
+        if (!key.length || !contextKey.length || contextKey.includes("..") || contextKey.startsWith(".") || contextKey.endsWith(".")) throw new Error("Context must not be an empty string!");
```
lauckhart, line 32: Would be good if error message matched test.  Bonus if it calls out the specific element that fails.

Also not sure about having "!" at the end of message unless that's a formal convention?

lauckhart (approved):
Note one comment but don't think it needs to block merge.

packages/matter-node.js/src/storage/StorageBackendDisk.ts
Apollon77: In fact it ends  up in "one of the context keys was an empty string" when that matches

PR #158, opened 2023-05-23
Cluster API enhancements

lauckhart:
Background:  While designing device-level API, a few small gaps in the cluster-level API were causing friction with implementation.  This is an attempt to address.

@Apollon77 - I *think* I covered all your asks here, see summary before you look at the code.  Definitely fixed the initial naive implementation once I actually ran it.  I didn't squash so you can see the differences.

Might be helpful to scan `ClusterServerFactoryTest.ts` and `CommandServerTest.ts` before looking at implementation.

Quick summary of scope:

---

**Problem:** It would be good to give the user a convenient mechanism for creating any cluster along the lines of `createDefaultScenesCluster()`.  Currently convenience is limited to clusters with hand-crafted server implementations.  This also made it difficult to automatically implement the clusters on the device.

**Proposal:** Implement generalized `ClusterServerFactory.create` that creates a `ClusterServer` for any cluster, regardless of whether we have better default handlers available.

**Side benefits:** The cluster server files can be a bit smaller as `ClusterServerFactory` handles more automatically.  Imports are smaller since there's only one factory method and it already knows about default handlers.  The default attribute values only use the schema as source-of-truth in this version.

---

**Problem:** Cluster definitions weren't available centrally.  `AllClustersMap` contained most clusters but it was keyed by ID so couldn't contain extended clusters which share an ID with their parent.  Previously I was scanning exports.  Worked but a bit icky.

**Proposal:** Create a central registry for clusters and extended clusters.  Populate the registry automatically when the cluster is defined.

**Side benefits:** Devs don't need to remember to update `AllClustersMap`.  Previously some handler exports accidentally didn't follow conventions; won't be an issue now.

---

**Problem:** It would be good to offer a mechanism for API users to invoke matter.js's "smart" handlers from within their own handlers.  There was a bit of an inversion of control issue here as `InteractionHandler` looks for the presence of a handler to understand if it should respond with `StatusCode.UnsupportedCommand`.

**Proposal:** Create an alternate path for handlers to convey status information to `CommandServer`.  In addition to the current approach (return value) create a specialized exception class that conveys a status code.

**Side benefits:** This pattern may make implementation a bit simpler because a handler can bail at any time with an exception.  Some of the current handlers threw exceptions that were effectively internal errors, now they'll just result in `StatusCode.UnsupportedCommand`.  Handlers could also use this to send other status codes though I'm not sure if that's desirable.  Finally, `CommandServer` has unit tests now. :)

### DIFF
--- a/packages/matter.js/src/cluster/ClusterIdentifier.ts
+++ b/packages/matter.js/src/cluster/ClusterIdentifier.ts
@@ -1,0 +1,17 @@
+/**
+ * @license
+ * Copyright 2022-2023 Project CHIP Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+// There will be an enum here.  But should ClusterIdentifier be this enum?
+// During development of a new type of cluster you would not want to be
+// constrained by current IDs.
+
+// If we want to use "ClusterId" we should probably rename datatype/ClusterId.
+
+/**
+ * A numeric code from the Matter Application Cluster specification that
+ * uniquely identifies a cluster.
+ */
+export type ClusterIdentifier = number;
--- a/packages/matter.js/src/cluster/ClusterServerFactory.ts
+++ b/packages/matter.js/src/cluster/ClusterServerFactory.ts
@@ -4,20 +4,14 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+import { MatterError } from "../common/MatterError.js";
 import { ClusterServer } from "../protocol/interaction/index.js";
 import { BitSchema, TypeFromBitSchema } from "../schema/BitmapSchema.js";
-import { Attributes, Cluster, Commands, Events, GlobalAttributes, MandatoryLocalAttributeNames, OptionalAttributeNames } from "./Cluster.js";
+import { Attributes, Cluster, Commands, Events, GlobalAttributes, MandatoryLocalAttributeNames } from "./Cluster.js";
 import { AttributeInitialValues, ClusterServerHandlers } from "./server/ClusterServer.js";
-import { UnsupportedCommandError } from "./server/CommandServer.js";
 
 type AnyCluster = Cluster<any, any, any, any, any>;
 const GLOBAL_ATTRIBUTES = new Set(Object.keys(GlobalAttributes({})));
-
-export function notImplemented() {
-    // Command server will fail gracefully with the appropriate status code
-    // in response to this error
-    throw new UnsupportedCommandError();
-}
 
 function mandatoryLocalAttributeNames<A extends Attributes>(attributes: A) {
     return Object.entries(attributes)
@@ -26,29 +20,13 @@ function mandatoryLocalAttributeNames<A extends Attributes>(attributes: A) {
         .map(([name]) => name as MandatoryLocalAttributeNames<A>);
 }
 
-function optionalAttributeNames<A extends Attributes>(attributes: A) {
-    return Object.entries(attributes)
-        .filter(([_name, attr]) => attr.optional)
-        .map(([name]) => name as OptionalAttributeNames<A>)
-}
-
 // Shortcut for cluster server implementers - create initial attribute map from
-// schema if not provided.  Could move into ClusterServer but then
-// attributeInitialValues argument would need to be optional.  Not sure if
-// that's desirable.  Also, this file is shorter than InteractionServer ;)
+// schema if not provided
 function createDefaultAttributeInitialValues<F extends BitSchema, SF extends TypeFromBitSchema<F>, A extends Attributes, C extends Commands, E extends Events>(cluster: Cluster<F, SF, A, C, E>): AttributeInitialValues<A> {
     const result = {} as AttributeInitialValues<A>;
 
     mandatoryLocalAttributeNames(cluster.attributes).forEach((name) =>
         (result as any)[name] = cluster.attributes[name].default);
-
-    // TODO - caller can override but should we make the set of optional
-    // attributes configurable?
-    optionalAttributeNames(cluster.attributes).forEach((name) => {
-        if (cluster.attributes[name] !== undefined) {
-            (result as any)[name] = cluster.attributes[name].default
-        }
-    });
 
     return result;
 }
@@ -63,12 +41,16 @@ export class ClusterServerFactory {
     /**
      * Retrieve default values for attributes based on cluster schema.
      */
-    public static getDefaultAttributeInitialValues<F extends BitSchema, SF extends TypeFromBitSchema<F>, A extends Attributes, C extends Commands, E extends Events>(cluster: Cluster<F, SF, A, C, E>) {
-        const aiv = this.defaultAttributeInitialValues.get(cluster) as AttributeInitialValues<A>;
+    static getDefaultAttributeInitialValues<F extends BitSchema, SF extends TypeFromBitSchema<F>, A extends Attributes, C extends Commands, E extends Events>(
+        cluster: Cluster<F, SF, A, C, E>
+    ) {
+        let aiv = this.defaultAttributeInitialValues.get(cluster);
         if (aiv == undefined) {
-            return createDefaultAttributeInitialValues(cluster);
+            this.defaultAttributeInitialValues.set(cluster, aiv = createDefaultAttributeInitialValues(cluster));
         }
-        return aiv;
+
+        // Do not pass by reference so the user can't overwrite
+        return Object.assign({}, aiv) as AttributeInitialValues<A>;
     }
 
     /**
@@ -78,15 +60,13 @@ export class ClusterServerFactory {
      * @param defaultHandlerProvider a function that create the cluster handlers
      * @param defaultAttributeInitialValues optional overrides for default attribute values
      */
-    public static registerClusterDefaults<F extends BitSchema, SF extends TypeFromBitSchema<F>, A extends Attributes, C extends Commands, E extends Events>(
-        cluster: Cluster<F, SF, A, C, E>, defaultHandlerProvider: () => ClusterServerHandlers<Cluster<F, SF, A, C, E>>,
+    static registerClusterDefaults<F extends BitSchema, SF extends TypeFromBitSchema<F>, A extends Attributes, C extends Commands, E extends Events>(
+        cluster: Cluster<F, SF, A, C, E>,
+        defaultHandlerProvider: () => ClusterServerHandlers<Cluster<F, SF, A, C, E>>,
         defaultAttributeInitialValues: AttributeInitialValues<A> = createDefaultAttributeInitialValues(cluster)
     ) {
         this.defaultHandlerProviders.set(cluster, defaultHandlerProvider);
         this.defaultAttributeInitialValues.set(cluster, defaultAttributeInitialValues);
-
-        const x = this.defaultAttributeInitialValues.get(cluster);
-        if (x.sceneCount) console.log(x.sceneCount);
     }
 
     /**
@@ -97,7 +77,11 @@ export class ClusterServerFactory {
      * @param handlers optional overrides for cluster handlers
      * @returns 
      */
-    public static create<F extends BitSchema, SF extends TypeFromBitSchema<F>, A extends Attributes, C extends Commands, E extends Events>(cluster: Cluster<F, SF, A, C, E>, attributeInitialValues?: AttributeInitialValues<A>, handlers?: ClusterServerHandlers<Cluster<F, SF, A, C, E>>) {
+    static create<F extends BitSchema, SF extends TypeFromBitSchema<F>, A extends Attributes, C extends Commands, E extends Events>(
+        cluster: Cluster<F, SF, A, C, E>,
+        attributeInitialValues?: AttributeInitialValues<A>,
+        handlers?: ClusterServerHandlers<Cluster<F, SF, A, C, E>>
+    ) {
         handlers = Object.assign({}, handlers);
 
         const provider = this.defaultHandlerProviders.get(cluster);
@@ -105,18 +89,20 @@ export class ClusterServerFactory {
             Object.assign(handlers, provider());
         }
 
-        for (const command in cluster.commands) {
-            if (!(handlers as any)[command]) {
-                (handlers as any)[command] = notImplemented;
-            }
-        }
+        const aiv = Object.assign(
+            this.getDefaultAttributeInitialValues(cluster),
+            attributeInitialValues);
 
-        // TODO - should we throw an error here if mandatory attributes do not
-        // have a value yet?
+        // Sanity check - ensure all mandatory attributes are present
+        const missing = mandatoryLocalAttributeNames(cluster.attributes)
+            .filter((name) => (<any>aiv)[name] === undefined);
+        if (missing.length) {
+            throw new MatterError(`Cannot create ${cluster.name} server, missing mandatory attributes: ${missing.join(", ")}`);
+        }
 
         return ClusterServer(
             cluster,
-            Object.assign({}, this.getDefaultAttributeInitialValues(cluster), attributeInitialValues),
+            aiv,
             handlers);
     }
 }
--- a/packages/matter.js/src/cluster/ClusterServerFactory.ts
+++ b/packages/matter.js/src/cluster/ClusterServerFactory.ts
@@ -1,0 +1,48 @@
+/**
+ * @license
+ * Copyright 2022-2023 Project CHIP Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { ClusterServer } from "../protocol/interaction/index.js";
+import { BitSchema, TypeFromBitSchema } from "../schema/BitmapSchema.js";
+import { AttributeInitialValues, Attributes, Cluster, ClusterServerHandlers, Commands, Events } from "./index.js";
+
+type AnyCluster = Cluster<any, any, any, any, any>;
+
+function notImplemented() {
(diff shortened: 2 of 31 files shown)


packages/matter.js/src/cluster/ClusterServerFactory.ts
```diff
@@ -0,0 +1,122 @@
...
+// that's desirable.  Also, this file is shorter than InteractionServer ;)
+function createDefaultAttributeInitialValues<F extends BitSchema, SF extends TypeFromBitSchema<F>, A extends Attributes, C extends Commands, E extends Events>(cluster: Cluster<F, SF, A, C, E>): AttributeInitialValues<A> {
+    const result = {} as AttributeInitialValues<A>;
+
+    mandatoryLocalAttributeNames(cluster.attributes).forEach((name) =>
+        (result as any)[name] = cluster.attributes[name].default);
+
+    // TODO - caller can override but should we make the set of optional
+    // attributes configurable?
+    optionalAttributeNames(cluster.attributes).forEach((name) => {
```
Apollon77, line 47: Idea for optional is that the dev decides if he supports them. So: if he sets a value he states support of them. If not then not. Means: we should not populate default values for optionales at all. 

Ps: we also can not be sure that all mandatory attributes have a default value ... just saying.

packages/matter.js/src/cluster/ClusterServerFactory.ts
```diff
@@ -0,0 +1,122 @@
...
+     * @param defaultAttributeInitialValues optional overrides for default attribute values
+     */
+    public static registerClusterDefaults<F extends BitSchema, SF extends TypeFromBitSchema<F>, A extends Attributes, C extends Commands, E extends Events>(
+        cluster: Cluster<F, SF, A, C, E>, defaultHandlerProvider: () => ClusterServerHandlers<Cluster<F, SF, A, C, E>>,
+        defaultAttributeInitialValues: AttributeInitialValues<A> = createDefaultAttributeInitialValues(cluster)
+    ) {
+        this.defaultHandlerProviders.set(cluster, defaultHandlerProvider);
+        this.defaultAttributeInitialValues.set(cluster, defaultAttributeInitialValues);
+
+        const x = this.defaultAttributeInitialValues.get(cluster);
```
Apollon77, line 88: Thats test code, right?!

packages/matter.js/src/cluster/ClusterServerFactory.ts
```diff
@@ -0,0 +1,122 @@
...
+     */
+    public static create<F extends BitSchema, SF extends TypeFromBitSchema<F>, A extends Attributes, C extends Commands, E extends Events>(cluster: Cluster<F, SF, A, C, E>, attributeInitialValues?: AttributeInitialValues<A>, handlers?: ClusterServerHandlers<Cluster<F, SF, A, C, E>>) {
+        handlers = Object.assign({}, handlers);
+
+        const provider = this.defaultHandlerProviders.get(cluster);
+        if (provider) {
+            Object.assign(handlers, provider());
+        }
+
+        for (const command in cluster.commands) {
```
Apollon77, line 108: Formally here the same with optional commands. We should not define them. In General er need a way later to identify implemented/supported commands vs not implemented ones - or need to populate global attributes here already.

packages/matter.js/src/cluster/ClusterServerFactory.ts
```diff
@@ -0,0 +1,122 @@
...
+            if (!(handlers as any)[command]) {
+                (handlers as any)[command] = notImplemented;
+            }
+        }
+
+        // TODO - should we throw an error here if mandatory attributes do not
```
Apollon77, line 114: Formally yes ... that's why my default clusters always had a full attribute initial values because we can not make sure all mandatory have default values.

packages/matter.js/src/cluster/server/ClusterServer.ts
```diff
@@ -36,7 +36,7 @@ import { Endpoint } from "../../device/Endpoint.js";
 export type AttributeServers<A extends Attributes> = Merge<Omit<{ [P in MandatoryAttributeNames<A>]: AttributeServer<AttributeJsType<A[P]>> }, keyof GlobalAttributes<any>>, { [P in OptionalAttributeNames<A>]?: AttributeServer<AttributeJsType<A[P]>> }>;
 
 /** Initial values for the cluster attribute */
-export type AttributeInitialValues<A extends Attributes> = Merge<Omit<{ [P in MandatoryAttributeNames<A>]: AttributeJsType<A[P]> }, keyof GlobalAttributes<any>>, { [P in OptionalAttributeNames<A>]?: AttributeJsType<A[P]> }>;
+export type AttributeInitialValues<A extends Attributes> = Merge<{ [P in MandatoryLocalAttributeNames<A>]: AttributeJsType<A[P]> }, { [P in OptionalAttributeNames<A>]?: AttributeJsType<A[P]> }>;
```
Apollon77, line 39: We cannot change this because this  else destroys typing! And why change it at all? The global attributes are populated internally. No user will ever populate them. 

If you need a Type including them we should use a new type.

packages/matter.js/src/cluster/server/CommandServer.ts
```diff
@@ -14,6 +14,32 @@ import { Endpoint } from "../../device/Endpoint.js";
 
 const logger = Logger.get("CommandServer");
 
+/**
+ * This error type allows a command handler to communicate the appropriate
+ * status code to CommandServer by throwing an exception.
+ */
+export class CommandError extends Error {
```
Apollon77, line 21: Please check MatterError class and especially StatusResponseError which is kind of what you do here.  If you do not want to use them pleaser inherit from that

packages/matter.js/src/cluster/server/CommandServer.ts
```diff
@@ -27,8 +53,19 @@ export class CommandServer<RequestT, ResponseT> {
     async invoke(session: Session<MatterDevice>, args: TlvStream, message: Message, endpoint: Endpoint): Promise<{ code: StatusCode, responseId: number, response: TlvStream }> {
         const request = this.requestSchema.decodeTlv(args);
         logger.debug(`Invoke ${this.name} with data ${Logger.toJSON(request)}`);
-        const response = await this.handler(request, session, message, endpoint);
-        logger.debug(`Invoke ${this.name} response : ${Logger.toJSON(response)}`);
-        return { code: StatusCode.Success, responseId: this.responseId, response: this.responseSchema.encodeTlv(response) };
+        try {
```
Apollon77, line 56: Not needed when using StatusResponseError because handling for this is already existing "upstream" in code

packages/matter.js/src/cluster/server/GroupsServer.ts
```diff
@@ -71,7 +71,7 @@ export class GroupsManager {
     }
 }
 
-export const GroupsClusterHandler: () => ClusterServerHandlers<typeof GroupsCluster> = () => {
+ClusterServerFactory.registerClusterDefaults(GroupsCluster, () => {
```
Apollon77, line 74: Maybe we could here already set the default attribute initial values that enhance the defined defaults - see the name support thingy.

packages/matter.js/test/device/MatterNodeStructureTest.ts
```diff
@@ -296,7 +296,7 @@ describe("Endpoint Structures", () => {
             assert.deepEqual(rootPartsListAttribute?.getLocal(), [new EndpointNumber(1)]);
 
             assert.equal(attributePaths.length, 146);
-            assert.equal(commandPaths.length, 38);
+            assert.equal(commandPaths.length, 39);
```
Apollon77, line 299: And that's the problem. There should be. I increase here. Only happens because we now have "not implemented" commands reported as supported commands. See above

Apollon77 (changes requested):
Good idea. I have some comments and needed adjustments. I might also come up with one topic in discord.

packages/matter.js/src/cluster/ClusterServerFactory.ts
lauckhart: OK, that's how I had it originally but the only optional value with a default was the lastConfiguredBy: null in ScenesCluster so I wasn't sure if default for options was intended to designate they're supported by default.

If you look at registerClusterDefaults, I added the defaultAttributeInitialValues there so the server implementation could override.  Do you think it would be OK to make it a partial so they don't have to copy all of the default values and could just change the ones that diverge from the schema?

packages/matter.js/src/cluster/ClusterServerFactory.ts
lauckhart: Yeah

packages/matter.js/src/cluster/ClusterServerFactory.ts
lauckhart: Sorry, forgot to take that out.

packages/matter.js/src/cluster/ClusterServerFactory.ts
lauckhart: Yeah I was thinking this was a safety check on the server implementation.

packages/matter.js/src/cluster/server/ClusterServer.ts
lauckhart: Wait, is it different semantically?  My intention was just to do the omit and merge in two steps, I was having a hard time parsing.

packages/matter.js/src/cluster/server/CommandServer.ts
lauckhart: I'll use StatusResponseError, looks like it's the same.

packages/matter.js/src/cluster/server/CommandServer.ts
lauckhart: I could be reading it wrong but it looks like the behavior will be different in handleInvokeResponse:

* If you throw it will go into the Promise.all which will then throw from handleInvokeRequest
* If you return a code then it will go into the invokeResponses array so other invocations could potentially succeed

But larger question is whether the approach is valid at all.  Last time we discussed I got the impression we needed to return the appropriate status code.  This solves for that, but it doesn't solve for any type of introspection.  Sounds like we need to be able to report on which commands are supported separately from invocation?

packages/matter.js/src/cluster/server/GroupsServer.ts
lauckhart: Yeah, that seems right to me.  I added the defaultInitialAttributeValues override to the registration function because you asked for that but I wasn't sure about making it a partial because you'd lose some of the type safety.  But if we check manually farther down then that would mitigate that issue.

packages/matter.js/test/device/MatterNodeStructureTest.ts
lauckhart: My bad.  When I auto-registered the clusters a bunch of numbers changed, I think I got clusters and commands confused.

packages/matter.js/src/cluster/ClusterServerFactory.ts
Apollon77: When we do Partial we need to throw if a value is missing... On the other hand - also because some values depend on each other it might make smore sense to say" you want to define, define all" because else dev also need to check the cluster def where we have a defautl and where mot and such ... will in the end become more complex and effort for the dev with an Partial.

So my conclusion would be: lets do with "full or nothing" (kind of) ... We can change to Partial any time later if we want - and we need to throw also anyweay ... sooo

packages/matter.js/src/cluster/ClusterServerFactory.ts
Apollon77: With the approach to use default values to populate default we need to check and throw

packages/matter.js/src/cluster/server/ClusterServer.ts
Apollon77: Yes but this is also used e.g. as parameters in constructor which would then pot requitre to also set global attributes that should be explicitely ignored here because populated internally

packages/matter.js/src/cluster/server/CommandServer.ts
Apollon77: The currentl code will catch that even one level higher.
I have one todo on my plate to make sure we answer all sorts of reponses correctly in all cases - like that case.
So this try catch that you added there I would move to https://github.com/project-chip/matter.js/blob/main/packages/matter.js/src/protocol/interaction/InteractionServer.ts#L431 here ... and then do generically ... when "StatusResponseError" then get the code out of the error and use this as response code and add reponse like that

```
                invokeResponses.push({ status: { commandPath, status: { status: code } } });
```

If you like do that already ... then I check the rest cases later.

The "not implemented" cases we have in code right now were throwing an "neutral exception" so that we see if someone is causing them because in fact it is more like "TODO needs to be implemented" instead of a valid "ok to not have them", thats whey they also till now were not really implemented as real errors.
So question is if we leave it that way or "clean it up" like in your approach.
Ideally they should all be implemented correctly by us (core logic) or "the dev".

packages/matter.js/src/cluster/server/GroupsServer.ts
Apollon77: And as said above: Even with the risk of "duplicate code" (aka defaults also listed in attribute initial values) I have no issue with this because then you see all attributes used in one place because sometimes they depend on each other

packages/matter.js/test/device/MatterNodeStructureTest.ts
Apollon77: I more assume it comes because of the "implementing optional commands with "no implementation"

packages/matter.js/src/cluster/ClusterServerFactory.ts
lauckhart: So the first use case is ScenesClusterServer.  Either has to copy 5 default values out of the schema or just add `lastConfiguredBy: null`.  Not sure if that's a representative use case.

Other 3 are fine though with just mandatory defaults.

packages/matter.js/src/cluster/ClusterServerFactory.ts
lauckhart: I'll probably add that anyway anyway just as a safety assertion.

packages/matter.js/src/cluster/server/ClusterServer.ts
lauckhart: If you look at the definition of MandatoryLocalAttributeNames it's (mandatory - global) so AttributeInitialValues are (mandatory - global + optional).  NBD to do the (mandatory - global) in multiple places though if you think it's safer, I'll maybe just leave a comment that they need to match.

If you ever want spread onto multiple lines with comments I'm happy to volunteer next time I'm digging through. :)

packages/matter.js/src/cluster/ClusterServerFactory.ts
Apollon77: ScenesCluster is easy to fix ... see cluster def: "Null" is default - and in the end is optional ... so we can also simply skip it until we support it)

packages/matter.js/src/cluster/server/CommandServer.ts
lauckhart: I'd probably make it explicit with IncompleteServerError extends StatusResponseError or somesuch but LMK what you'd prefer, I'll change the ones in the handlers to that.

packages/matter.js/src/cluster/server/ClusterServer.ts
Apollon77: Alll ok you added a second one ... then also fine - we just need to check the places where the "former only type" is used and adjust to the new named one! 8at least devcie constructir .. if it stays that way) and also in cluster defs and such

packages/matter.js/test/device/MatterNodeStructureTest.ts
lauckhart: Yeah that's what caused it

packages/matter.js/src/cluster/server/CommandServer.ts
Apollon77: subclasses are fine, so we can still catch StatusResponseError generically ... thats what I would like to keep as option because a command can throw also other statusResponse errors nd so the catch is generic

lauckhart:
@Apollon77 FYI, I haven't forgotten about this guy...  Waiting for some variant of #162 and #165 to land, will finish modifications to cluster registration then do the 😱 merge.

Apollon77:
I agree. Lets do the current cluster topic first. Then Generation of feature clusters and this. In my eyes this used generated ones and Turnschuhs ones as "all in" clusters.

packages/matter.js/src/cluster/ClusterIdentifier.ts
```diff
@@ -0,0 +1,17 @@
+/**
+ * @license
+ * Copyright 2022-2023 Project CHIP Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+// There will be an enum here.  But should ClusterIdentifier be this enum?
```
Apollon77, line 7: I would do this as enum with all „cluster ids defined by specs“ irrelevant if we have them already implemented or not. With this they are not limiting in any way - at least not more limiting then the specs itself ;-) or what do you mean? 
i would not rename ClusterId because that’s official from spec type „cluster-id“ like the others too. In theory

lauckhart:
Closing this as the generated clusters make the factory largely irrelevant. Will resurrect other bits as necessary as part of device generation.

PR #159, opened 2023-05-23
Allows Node.js Buffer objects to be persisted to storage as a Uint8Arrays that they subclass

### LINKED ISSUE #138: Resuming a Session the second time fails
I'm using the ControllerNode class to resume a session to an already paired node (Hue bridge ).  The pairing works correctly, and the 1st time we resume, this works correctly as well.  Unfortunately, after resuming the first time, the CaseClient receives the `sigma2resume` message in a incompatible format here: 
https://github.com/project-chip/matter.js/blob/f68c2a1760485c4da3f6c897fd56ceb4e08a5797/packages/matter.js/src/session/case/CaseClient.ts#L46 
 if i dump the `sigma2resume` message, it looks something like:
```json
{"resumptionId":{"type":"Buffer","data":[161,2,101,132,250,185,175,4,185,69,179,75,60,143,34,190]},"resumeMic":{"type":"Buffer","data":[134,115,226,224,208,16,250,27,80,52,38,173,21,48,233,95]},"sessionId":18290}
```
I think the values should be Uint8Arrays, instead they look like raw Node.js Buffer objects.  The client then persists these to storage, and the next time it goes to use them (so the second time we resume) we get a range error here when doing the ByteArray  concat call:
https://github.com/project-chip/matter.js/blob/f68c2a1760485c4da3f6c897fd56ceb4e08a5797/packages/matter.js/src/session/case/CaseClient.ts#L38

I tried hunting this down for a bit, but I can't figure out why these raw buffer objects are returned, where everywhere else it seems to be fine?

digitaldan:
Fixes #138

### DIFF
--- a/packages/matter.js/src/storage/StringifyTools.ts
+++ b/packages/matter.js/src/storage/StringifyTools.ts
@@ -57,7 +57,8 @@ export function toJson(object: SupportedStorageTypes): string {
         if (value instanceof Uint8Array) {
             return `{"${JSON_SPECIAL_KEY_TYPE}":"Uint8Array","${JSON_SPECIAL_KEY_VALUE}":"${value.toHex()}"}`;
         }
-        if (value != null && typeof value === 'object' && value.type === 'Buffer' && value.data instanceof Array) {
+        //Node.js can sometimes pass in a native Buffer object in place of a Uint8Array, of which it is a subclass of, the Buffer class implements its own toJSON method which breaks our serialization.
+        if (value != null && typeof value === 'object' && value.type === 'Buffer' && value.data) {
             return `{"${JSON_SPECIAL_KEY_TYPE}":"Uint8Array","${JSON_SPECIAL_KEY_VALUE}":"${Uint8Array.from(value.data).toHex()}"}`;
         }
         if (value instanceof Map) {
--- a/packages/matter.js/src/storage/StringifyTools.ts
+++ b/packages/matter.js/src/storage/StringifyTools.ts
@@ -58,7 +58,7 @@ export function toJson(object: SupportedStorageTypes): string {
             return `{"${JSON_SPECIAL_KEY_TYPE}":"Uint8Array","${JSON_SPECIAL_KEY_VALUE}":"${value.toHex()}"}`;
         }
         //Node.js can sometimes pass in a native Buffer object in place of a Uint8Array, of which it is a subclass of, the Buffer class implements its own toJSON method which breaks our serialization.
-        if (value != null && typeof value === 'object' && value.type === 'Buffer' && value.data) {
+        if (value != null && typeof value === 'object' && value.type === 'Buffer' && Array.isArray(value.data)) {
             return `{"${JSON_SPECIAL_KEY_TYPE}":"Uint8Array","${JSON_SPECIAL_KEY_VALUE}":"${Uint8Array.from(value.data).toHex()}"}`;
         }
         if (value instanceof Map) {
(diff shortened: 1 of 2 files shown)


digitaldan:
So after much digging around, i have come to the conclusion that Node.js can sometimes use native Buffer objects interchangeably with Uint8Arrays, which they subclass, and this seems to work except for serialization.  I don't know why that is. 

The Buffer class provides a custom `toJSON` method which is used for serialization, you can see this here:
https://nodejs.org/api/buffer.html#buftojson
as well as:
https://github.com/nodejs/node/blob/c55aa01e4873152ea6a0dbff98d55d403b6b97bb/lib/buffer.js#L1135 

Hence the issue with storage.

I have this fix,  and can now resume my session again and again.  There does not seem to be any other issues so far as i guess as Buffer objects subclass Uint8Arrays and are largely interchangeable ?

This seems kinda kludgy , but i also still don't understand why sometimes Uint8Arrays are used vs Buffer objects.

packages/matter.js/test/util/storage/StringifyToolsTest.ts
```diff
@@ -68,6 +68,19 @@ describe("JsonConverter", () => {
             assert.ok(decodedObj instanceof Uint8Array);
         });
 
+        it("encode/decode Node.js Buffer", () => {
+
+            const obj = { type: 'Buffer', data: [1, 2, 3] };
```
Apollon77, line 73: But are you sure this already is a jsonified Buffer? I would have expected that a normal Buffer would come in here, does your code work then still?

packages/matter.js/src/storage/StringifyTools.ts
```diff
@@ -57,6 +57,9 @@ export function toJson(object: SupportedStorageTypes): string {
         if (value instanceof Uint8Array) {
             return `{"${JSON_SPECIAL_KEY_TYPE}":"Uint8Array","${JSON_SPECIAL_KEY_VALUE}":"${value.toHex()}"}`;
         }
+        if (value != null && typeof value === 'object' && value.type === 'Buffer' && value.data instanceof Array) {
```
Apollon77, line 60: is that really already json-ified value in this place?

In any case please add a comment here why we do this because we would add logic for Buffers (which should not be part of matter.js) but needed here ...

Apollon77 (changes requested):
Iiinteresting .. but I still do not understand WHY that happens

packages/matter.js/test/util/storage/StringifyToolsTest.ts
digitaldan: yeah, sorry, this was a left over of me using the test method for trying out various inputs, I'll have added back in the Buffer object that i intended to have here.

packages/matter.js/src/storage/StringifyTools.ts
digitaldan: > is that really already json-ified value in this place?

So, this was the part that confused me a bit. Its not JSON per se , the `JSON.stringify` replacer function calls `toJson` if that function exists on a object its iterating on , so the `value` here is not an actual Buffer  but an object.  Whats more confusing is that `toJSON` does not actually return JSON, but rather a object that is intended to be JSON serialized, its a badly named function imho.  You can see that at https://github.com/nodejs/node/blob/c55aa01e4873152ea6a0dbff98d55d403b6b97bb/lib/buffer.js#L1135

Apollon77 (changes requested):
I propose this one last change

packages/matter.js/src/storage/StringifyTools.ts
```diff
@@ -57,6 +57,10 @@ export function toJson(object: SupportedStorageTypes): string {
         if (value instanceof Uint8Array) {
             return `{"${JSON_SPECIAL_KEY_TYPE}":"Uint8Array","${JSON_SPECIAL_KEY_VALUE}":"${value.toHex()}"}`;
         }
+        //Node.js can sometimes pass in a native Buffer object in place of a Uint8Array, of which it is a subclass of, the Buffer class implements its own toJSON method which breaks our serialization.
+        if (value != null && typeof value === 'object' && value.type === 'Buffer' && value.data) {
```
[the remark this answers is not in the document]
digitaldan: agreed

Apollon77 (approved):

PR #160, opened 2023-05-24
Add Interface for Events

Apollon77:
This PR adds the first step for Event support. This means the basic ClusterSeverr/ClusterClient interface and typing is added.

I decided for now that for CLusterServer the full list of Events needs to be provided, which means that also the mandatory ones are needed to be provided - mainly that the Dev on this level "knows" that he needs to support this. We could discuss if we only want to have the optional ones being specified, but so it matches for me to the current level of string typing in ClusterServer.

The user will interact with EventServer and EventClient instances - in fact the same concept as alreayd used for Attributes. And we have convenient named methods to trigger an event (Server) or read/subscribe for events (Client).

Internally because of the way Events needs to be managed according to specs I added an EventManager class which will hold all events in a list based on priority and holds the events up to a certain number (se to 10k for now). The final implemntation shere and all logic to read, subscribe to events will be added in a later PR because it is internally without big interface changes.


lauckhart (approved):
LGTM.

PR #161, opened 2023-05-25
Dynamic attribute checking

vves:
WIP for discussion


packages/matter.js/src/cluster/Cluster.ts
```diff
@@ -23,29 +23,188 @@ export const enum AccessLevel {
 }
 
 /* Interfaces and helper methods to define a cluster attribute */
-export interface Attribute<T> { id: number, schema: TlvSchema<T>, optional: boolean, readAcl: AccessLevel, writable: boolean, scene: boolean, persistent: boolean, omitChanges: boolean, writeAcl?: AccessLevel, default?: T }
+export interface Attribute<T> {
+    id: number, schema: TlvSchema<T>,
+    optional: boolean,
+    optionalIf: number[],
```
Apollon77, line 29: I would really put these just into "*Optional*" attributes -or better as own "Conditional" types. I would also define that a conditional attribute is (typing wise for this cluster) always Optional ...

And in fact "number" is the wring typing, or? I would not store the encoded values here, just the feature data

packages/matter.js/src/protocol/interaction/InteractionMessenger.ts
```diff
@@ -160,10 +160,7 @@ export class InteractionServerMessenger extends InteractionMessenger<MatterDevic
 
     async sendDataReport(dataReport: DataReport) {
         const messageBytes = TlvDataReport.encode(dataReport);
-        if (!Array.isArray(dataReport.attributeReports)) {
```
Apollon77, line 163: thats a merge issue it seems ... this code is no longer in main!!

packages/matter.js/src/protocol/interaction/InteractionMessenger.ts
```diff
@@ -160,10 +160,7 @@ export class InteractionServerMessenger extends InteractionMessenger<MatterDevic
...
-            throw new Error(`DataReport.values must be an array, got: ${dataReport.attributeReports}`);
-        }
-        if (messageBytes.length > MAX_SPDU_LENGTH) {
+        if(dataReport.attributeReports !== undefined && messageBytes.length > MAX_SPDU_LENGTH){
```
Apollon77, line 163:
```suggestion
        if (dataReport.attributeReports !== undefined && messageBytes.length > MAX_SPDU_LENGTH) {
```

packages/matter.js/src/protocol/interaction/InteractionServer.ts
```diff
@@ -46,6 +46,20 @@ export const INTERACTION_PROTOCOL_ID = 0x0001;
 
 const logger = Logger.get("InteractionProtocol");
 
+function getEnabledBits(bitmask: number): number[] {
```
Apollon77, line 49: if you do not encoe the feature data above you do not need to do this here :-)

packages/matter.js/src/protocol/interaction/InteractionServer.ts
```diff
@@ -132,11 +146,26 @@ export function ClusterServer<F extends BitSchema, SF extends TypeFromBitSchema<
...
     for (const attributeName in attributeDef) {
         const capitalizedAttributeName = capitalize(attributeName);
+        // logger.info(`check this for REQUIRED Attributes ${JSON.stringify(attributeName)}`)
+
+        const { requiredIf, optionalIf } = attributeDef[attributeName]
+        const requiredOk = validateInitialAttributesExist<F, SF, A, C, E>(requiredIf, clusterDef, attributesInitialValues, attributeName, supportedFeatures);
+        if(!requiredOk){
+            logger.warn(`*** InitialAttributeValue for "${clusterDef.name}/${attributeName}" is REQUIRED by supportedFeatures:${JSON.stringify(supportedFeatures)} but is not set`);
+        }
+        const optionalOk = validateInitialAttributesExist<F, SF, A, C, E>(optionalIf, clusterDef, attributesInitialValues, attributeName, supportedFeatures);
```
Apollon77, line 160: why we need this? optionalis optional, so alles if not there

Apollon77 (changes requested):
see my comments, partially already noted in Discord. And I think we should not compare features as numbers, why the not the named objects?

packages/matter.js/src/cluster/Cluster.ts
lauckhart: Given that optionality is conditional, maybe optionality should be more of a query?  Something like:

```js
    isRequired(context: Cluster<lotsOfAnys>)
```

or `isOptional` works too, that just feels like more of a double negative to me for some reason.

Also maybe some of this stuff could move into an `Element` base interface/class shared by Attribute, Command and Event.  That would reduce repetition.

I do love seeing it spread out like that.  Now just needs to go into a separate file. 😉

packages/matter.js/src/protocol/interaction/InteractionServer.ts
vves: I read the spec as optional being recommended for best support across all consumers.  So noting it for developers using the library is helpful.

packages/matter.js/src/protocol/interaction/InteractionServer.ts
vves: If you have a way of encoding something more basic than number[] I am all ears.  The Attributes don't have scope of the Features and I didn't think it appropriate to make the dependency.  Dig deeper if you think there is a better solution.

packages/matter.js/src/cluster/Cluster.ts
```diff
@@ -23,29 +23,188 @@ export const enum AccessLevel {
 }
 
 /* Interfaces and helper methods to define a cluster attribute */
-export interface Attribute<T> { id: number, schema: TlvSchema<T>, optional: boolean, readAcl: AccessLevel, writable: boolean, scene: boolean, persistent: boolean, omitChanges: boolean, writeAcl?: AccessLevel, default?: T }
+export interface Attribute<T> {
+    id: number, schema: TlvSchema<T>,
```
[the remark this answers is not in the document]
vves: formatting oversight.  let's fix this kind of thing on the merge.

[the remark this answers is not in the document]
vves: actually - latest prettier had fixed after this review.

vves:
> 

Readability.  Having the 'Code' values from the pdf spec makes coding the Clusters less error prone.

{lift:true, tilt:false, liftPositionAware:true, tiltPositionAware:false, absolutePosition:false} vs LF_PALF

packages/matter.js/src/cluster/Cluster.ts
vves: @lauckhart let me see if Element makes sense.

lauckhart: Could do it later, just wanted to mention

vves: @lauckhart that didn't take long.  Adding a base Element class is a significant undertaking outside the scope of this PR.  Let's discuss this on Discord.

vves:
@Apollon77 what do you think about accepting this as is since it is an extension without breaking changes for anyone else.  This allows us to move forward with some of the other ideas and for you and @lauckhart to use it as well as better understand why number[] was my first choice.  Additionally - the WCCluster PR is dependent on this PR.

lauckhart:
I think that's probably the right thing to do. Just a bit of git history if we decide to change.  Would be good to get everything in.

Apollon77:
This a bit overlapped with my approach where I used this to create #165 as proposal with a bit better typing and also based on adjustments discussed in Discord in parallel ...

Apollon77:
Incorporated in #165, thank you very much!

PR #162, opened 2023-05-25
Window-covering-cluster

vves:
Window Covering Cluster.  Dependent on dynamic-attribute-checking


Apollon77:
Incorporated in #165, thank you very much!

PR #165, opened 2023-05-26
Conditional ClusterSchemas and WindowCoveringCluster

Apollon77:
This PR is my propsal for the Conditional idea with runtime check and also as basis for typing. The PR includes:

* Enhance BitMap typing and Schemas to allow "Partially" provided Bitmaps by suprerssing the "unset" bits. this should make is easier and less code
* Clusters use this Partial type for Schemas
* Intoduce new types for Attributes, Commands and Events as "Conditional" type to be used in ClusterSchemas (new idea of "the all in one Cluster used as basis for generation and anything). These types allow to specify mandatoryId and optionalIf attributes as array of corresponding cluster types. All this is typed correctly. Conditional* is inherited from Optional, so are all optional by definition
* On ClusterServer instance creation - if Conditional Attrinbutes are used - there is a type checking which logs
* I created (as discussed as idea) new dir cluster/schema with these type of clusters
* I also added then the Window Cluster for the other Extend variants in normal dir (would be replaced by generated one)

Additionally I fixed the command return types because they are TlvNoResponse. And I added some first "scene" flags.  I will do another fixing run later.

Happy to get your feedback

### DIFF
--- a/packages/matter.js/src/cluster/Cluster.ts
+++ b/packages/matter.js/src/cluster/Cluster.ts
@@ -5,7 +5,7 @@
  */
 
 import { Merge } from "../util/Type.js";
-import { BitSchema, TypeFromBitSchema } from "../schema/BitmapSchema.js";
+import { BitSchema, TypeFromPartialBitSchema } from "../schema/BitmapSchema.js";
 import { TlvSchema } from "../tlv/TlvSchema.js";
 import { TlvVoid } from "../tlv/TlvVoid.js";
 import { TlvFields, TlvObject, TypeFromFields } from "../tlv/TlvObject.js";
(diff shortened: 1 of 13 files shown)


packages/matter.js/src/protocol/interaction/InteractionServer.ts
```diff
@@ -48,7 +50,34 @@ export const INTERACTION_PROTOCOL_ID = 0x0001;
...
+function isConditionalEvent(event: Event<any>) {
+    return (event as any).mandatoryIf !== undefined || (event as any).optionalIf !== undefined;
+}
+
+function validateConditionalEntity<F extends BitSchema, SF extends TypeFromBitSchema<F>>(
+    featureSets: ConditionalFeatureList<F>,
+    entityValue: any,
+    supportedFeatures: SF): boolean {
+    for (const features of featureSets) {
+        if (Object.keys(supportedFeatures).every(feature => !!features[feature] === !!supportedFeatures[feature])) {
```
lauckhart, line 69: I was hoping if we used bit schema there's be proper set operations/testing.  Can this become supportedFeatures.hasAll(features)?

packages/matter.js/src/protocol/interaction/InteractionServer.ts
```diff
@@ -48,7 +50,34 @@ export const INTERACTION_PROTOCOL_ID = 0x0001;
...
+            console.log(`check value ${entityValue}`);
```
lauckhart: Dev remnant?

packages/matter.js/src/protocol/interaction/InteractionServer.ts
```diff
@@ -223,6 +274,23 @@ export function ClusterServer<F extends BitSchema, SF extends TypeFromBitSchema<
     const generatedCommandList = new Array<number>();
     for (const name in commandDef) {
         const handler = (handlers as any)[name];
+
+        if (isConditionalCommand(commandDef[name])) {
+            const { mandatoryIf, optionalIf } = commandDef[name] as any;
```
lauckhart, line 285: If above test was commandDef[name].isConditional then it would act as a type assertion and the cast to any would not be nessary.  If not that, then isConditionalCommand could assert type to make cast unnecessary.

packages/matter.js/src/cluster/Cluster.ts
```diff
@@ -22,58 +22,487 @@ export const enum AccessLevel {
...
+    mandatoryIf?: ConditionalFeatureList<F>,
+    optionalIf?: ConditionalFeatureList<F>
+}
+
 export type ResponseType<T extends Command<any, any>> = T extends OptionalCommand<any, infer ResponseT> ? ResponseT : (T extends Command<any, infer ResponseT> ? ResponseT : never);
 export type RequestType<T extends Command<any, any>> = T extends OptionalCommand<infer RequestT, any> ? RequestT : (T extends Command<infer RequestT, any> ? RequestT : never);
 
-export const Command = <RequestT, ResponseT>(requestId: number, requestSchema: TlvSchema<RequestT>, responseId: number, responseSchema: TlvSchema<ResponseT>): Command<RequestT, ResponseT> => ({ optional: false, requestId, requestSchema, responseId, responseSchema });
-export const OptionalCommand = <RequestT, ResponseT>(requestId: number, requestSchema: TlvSchema<RequestT>, responseId: number, responseSchema: TlvSchema<ResponseT>): OptionalCommand<RequestT, ResponseT> => ({ optional: true, requestId, requestSchema, responseId, responseSchema });
+interface ConditionalCommandOptions<F extends BitSchema> {
```
lauckhart, line 387: Why did you define a different interface for every element type?  Vs. just ElementOptions

packages/matter.js/src/cluster/Cluster.ts
```diff
@@ -22,58 +22,487 @@ export const enum AccessLevel {
...
+interface ConditionalEventOptions<F extends BitSchema> {
+    optionalIf?: ConditionalFeatureList<F>,
+    mandatoryIf?: ConditionalFeatureList<F>
+}
+
+export interface OptionalEvent<T> extends Event<T> {
+    optional: true
+}
+
+export interface ConditionalEvent<T, F extends BitSchema> extends OptionalEvent<T> {
```
lauckhart, line 457: I think this can just extend OptionalEvents<T> & ConditionalEventOptions<T> (or ElementOptions<T>) and have an empty body?

lauckhart (changes requested):
I looked closer and I think this looks good.  The partial bitmap is a nice innovation.  I did add some suggestions.

If we merge this PR should probably get the full commit history in.

packages/matter.js/src/cluster/WindowCoveringCluster.ts
```diff
@@ -0,0 +1,177 @@
...
+import { Attribute, ClusterExtend, Command, OptionalCommand, TlvNoResponse } from "./Cluster.js";
+import { ClusterServerObjForCluster } from "./server/ClusterServer.js";
+import { WindowCoveringClusterSchema } from "./schema/WindowCoveringCluster.js";
+import { MatterApplicationClusterSpecificationV1_0 } from "../spec/Specifications.js";
+
+/** alias for percentages expressed as 0 to 100.00 with 2 decimals */
+const WCPercent100ths = TlvUInt16.bound({ max: 10000 });
+
+/** @see {@link MatterApplicationClusterSpecificationV1_0} § 5.3.5.1 */
+export enum WindowCoveringType {
```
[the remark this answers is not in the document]
Apollon77: Yes normally yes, but because we stiill have some CLusterExtdn here (to not block vves until generation is there) they are needed, but I do not wanted to export them in the other one because they should not be needed to be exported. And because we would throw that file away anyway when generation is there I decided for the "copy" way.

packages/matter.js/src/cluster/schema/WindowCoveringCluster.ts
```diff
@@ -0,0 +1,478 @@
+/**
```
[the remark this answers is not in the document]
Apollon77: i rename

packages/matter.js/src/protocol/interaction/InteractionServer.ts
Apollon77: supported Features in this place is "just a data object that has a defined set of keys with a defined valuetype". not more. so ... sorry but no ... We could do that but then on the Bitschea class whcih would be also a lot of overhead because then we always need to build the BitmapSchema instance ... would be bad performance wise. :-(

packages/matter.js/src/protocol/interaction/InteractionServer.ts
Apollon77: already out with a following commit

packages/matter.js/src/cluster/Cluster.ts
Apollon77: In attribute it is "special" anyway. For commands and events yes could be reused .. decided to kep separated for later extensibility (else we need split up later) ...

packages/matter.js/src/cluster/WindowCoveringCluster.ts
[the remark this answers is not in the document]
lauckhart: Cool I just need to avoid accidentally deleting.

packages/matter.js/src/protocol/interaction/InteractionServer.ts
```diff
@@ -48,7 +50,34 @@ export const INTERACTION_PROTOCOL_ID = 0x0001;
 
 const logger = Logger.get("InteractionProtocol");
 
-export function ClusterServer<F extends BitSchema, SF extends TypeFromBitSchema<F>, A extends Attributes, C extends Commands, E extends Events>(
+function isConditionalAttribute(attribute: Attribute<any>) {
```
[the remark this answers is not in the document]
Apollon77: makes sense

packages/matter.js/src/protocol/interaction/InteractionServer.ts
lauckhart: Maybe a utility export then?  Would be nice to have this short and explicit, will probably need stuff like this for code gen too.  I can add later if you don't want to do now.

packages/matter.js/src/cluster/Cluster.ts
lauckhart: No reason you can't specialize the base interface if necessary.  So might be ElementOptions -> AttributeOptions -> ConditionalAttribute but only ElementOptions -> ConditionalEvent.  NBD as is but expect me to squawk any time lines are duplicated in four places. 🙂

packages/matter.js/src/protocol/interaction/InteractionServer.ts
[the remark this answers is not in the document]
Apollon77: done

packages/matter.js/src/protocol/interaction/InteractionServer.ts
Apollon77: done

lauckhart (commented):
LGTM, will defer to @vves for final approval since this encompasses WC

packages/matter.js/src/schema/BitmapSchema.ts
```diff
@@ -52,6 +53,14 @@ export class BitmapSchemaInternal<T extends BitSchema> extends Schema<TypeFromBi
         // TODO: validate that bitSchemas is coherent
     }
 
+    /**
+     * Allow to use a filly defined Bitmap schema as input, but also allow one where only the entries of bits set are
```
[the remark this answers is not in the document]
Apollon77: lol, fixed in next commit

Apollon77:
Further WC rework is prepared in https://github.com/Apollon77/matter.js/pull/2/files#diff-7a1cae91ba88568f50b6f19025c6031c6d748c9f83c26ac28205593e62bada77 ...

Additionalyl I openend one specs check issue: https://github.com/project-chip/connectedhomeip/issues/26905

Apollon77:
> If we merge this PR should get the full commit history in. Git blame generally isn't very useful in the repository right now.

I will try to remember when I merge :-)

But in fact because it is basing on main and not the other PRs it's mainly "ingo was it"

PR #169, opened 2023-05-30
Hvac related clusters and devices & some fixes on existing clusters

mow:
-  extended bridged device node example with new clusters (testing purposes)
-  bridged device node mandatory parameter unid added (unique id)
-  Thermostat Cluster added
-  Thermostat device added
-  Fan Control Cluster added
-  Fan Control device added
-  HVAC device added (no client, WIP)
- Level Control Cluster modified


packages/matter-node.js/src/examples/BridgedDevicesNode.ts
```diff
@@ -72,7 +72,7 @@ class BridgedDevice {
         // product name / id and vendor id should match what is in the device certificate
         const vendorId = new VendorId(getIntParameter("vendorid") ?? deviceStorage.get("vendorid", 0xFFF1));
         const productName = `node-matter OnOff-Bridge`;
-        const productId = getIntParameter("productid") ?? deviceStorage.get("productid", 0x8000);
+        const productId = getIntParameter("productid") ?? deviceStorage.get("productid", 0x1000);
```
Apollon77, line 75: Please do not change default here because this id is officially not allowed for dev devices.

packages/matter-node.js/src/examples/BridgedDevicesNode.ts
```diff
@@ -97,7 +97,9 @@ class BridgedDevice {
 
         const matterServer = new MatterServer(storageManager, netAnnounceInterface);
 
-        const uniqueId = Time.nowMs(); // TODO Store it!
+        //console.log("UNID: " + getIntParameter("unid"));
+
+        const uniqueId = getIntParameter("unid"); //Time.nowMs(); // TODO Store it!
```
Apollon77, line 102: Good idea. Please also name parameter uniqueid and yes ideally put into storage and use like done with other parameters

packages/matter-node.js/src/examples/BridgedDevicesNode.ts
```diff
@@ -144,6 +147,39 @@ class BridgedDevice {
             });
         }
 
+        const thermostatDevice = new ThermostatDevice();
```
Apollon77, line 150: Please remove that again. You can still have this just for you locally. We will think about how having examples for more deviceTypes later. Please not mix too much

packages/matter.js/src/cluster/FanControlCluster.ts
```diff
@@ -0,0 +1,152 @@
...
+export const enum FanModeSequence {
+    OffLowMedHigh = 0,
+    OffLowHigh = 1,
+    OffLowMedHighAuto = 2,
+    OffLowHighAuto = 3,
+    OffOnAuto = 4,
+    OffOn = 5,
+}
+
+const rockBitmap = TlvBitmap(TlvUInt8, {
```
Apollon77, line 39: Please name all types with Tlv types also Tlv at the beginning.

Apollon77 (changes requested):
Hi, I could just do a rough review on my mobile with some commands. So I did not checked against specs yet. 

Why commands are missing for that one cluster?
In fact on the specs we need Saeco d pass but lets wait till 165 has landed as already said in the issue. Especially because it seems you only focused on certain feature combinations. Alternatively define anything which is not "always mandatory" as optional.

packages/matter.js/src/device/DeviceTypes.ts
```diff
@@ -150,7 +150,7 @@ export const DeviceTypes: { [key: string]: DeviceTypeDefinition } = {
         deviceClass: DeviceClasses.Node, // ???
         revision: 1,
         requiredServerClusters: [
-            MatterClusters.WiredPowerSourceCluster.id // any cluster features allowed
```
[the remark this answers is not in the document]
mow: I assume by mistake, I'll check it

packages/matter.js/src/cluster/FanControlCluster.ts
```diff
@@ -0,0 +1,152 @@
...
+    supportedFeatures: {
+        autoMode: true,
+        multiSpeed: true,
+        rocking: true,
+        windEmulation: true
+    },
+    attributes: {
+        ...commonAttributes,
+    },
+    // commands: {
```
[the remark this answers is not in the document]
mow: Fan Control Cluster doesn't have any commands according to specs, so I removed

packages/matter.js/src/cluster/FanControlCluster.ts
```diff
@@ -0,0 +1,152 @@
...
+    rockSetting: OptionalWritableAttribute(0x0008, rockBitmap, { default: { rockLeftRight: false, rockUpDown: false, rockRound: false } }),
+
+    // Support for wind feature
+    windSupport: OptionalAttribute(0x0009, windBitmap, { default: { sleepWind: true, naturalWind: true } }),
+
+    // Wind setting
+    windSetting: OptionalWritableAttribute(0x000A, windBitmap, { default: { sleepWind: false, naturalWind: false } }),
+
+    // Feature Map global attribute
+    featureMap: WritableAttribute(0xFFFC, featureMapBitmap, {
```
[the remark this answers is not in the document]
mow: sure, this was just a trial to understand why most of the fan control cluster attributes not working

packages/matter-node.js/src/examples/BridgedDevicesNode.ts
mow: sure

packages/matter-node.js/src/examples/BridgedDevicesNode.ts
mow: it was for testing google project, I'll do it.

packages/matter.js/src/cluster/FanControlCluster.ts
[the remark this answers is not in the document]
Apollon77: All good but then also no need to have something commented out here ;-)

Apollon77:
I will do review in a week once back from vacation. Need my laptop for that ;) sorry for the delay.

Apollon77:
@mow Thank you for your contribution and sorry that a final feedback took that long.
We were woring hard on auto generation for all clusters whcih just landed in #192. This means that now the clusters are available. We will do a PR soonnto switch the used clusters from the current "manual written ones" to these soon.

After that you could rebase/rework your PR and we add the device classes - in fact generating device classes is kind of next on the list ... 

Let me know what you thin we shoukd do with your PR

mow:
@Apollon77 you can do whatever you want, perhaps closing this PR and reusing needed parts with c/p is going to be better.

Apollon77:
Now we have anything needed in 0.8 API to buold this correctly, cloding here. Sorry for the delay

PR #184, opened 2023-06-19
Add device type list to readme

Apollon77:
I collected from best knowledge what I found compatible ...

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,15 +7,18 @@ Implementation of Matter protocol in Typescript with no native dependencies (and
 Matter is a new secure / reliable / local / standard protocol for smart devices launched at the end of 2022.
 To know more about Matter: https://csa-iot.org/all-solutions/matter/
 
+## Compatibility
 matter.js/matter-node.js is compatible with:
 - **Apple iOS (iPhone or iPad) and tvOS 16 (Apple TV) - "Home" app by Apple**: fully working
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -103,18 +106,64 @@ Completed implementations:
...
+## Device types supported by Ecosystems
+
+This list is a compilation of published information and own community tests. Many updates in the ecosystems happen without big announcements, so the information here can be a bit outdated and more device types are supported. If you find errors or added device types please open a PR or issue to report them. Especially for SmartThings and Tuya no information is officially published. 
+
+Table Legend:
+* "X" means supported
+* "-" means not supported from current knowledge
+* "?" means unknown
+
+| Gerätetyp                         | Apple | Google        | Amazon | SmartThings | Tuya |
```
lauckhart, line 122: Does "Gerätetyp" = "feature"?

lauckhart (approved):
I think this is fine for now but it would be cool to get this type of information into a machine-readable format.  We could then generate documentation but also use programmatically to create warnings and/or tailor published devices for specific clients.

JimBuzbee (approved):
Did a bit of testing with SmartThings. As usual, it's a bit of a pain to get SmartThings to pair, but I was able to get the following Matter devices working: Contact, Light, Temperature, Humidity, and Flow

README.md
Apollon77: Devcie type ... (sorry german leftover, fixed)

PR #186, opened 2023-06-21
Fix Subscription Timer duplication and Optimizations

### LINKED ISSUE #185: Duplicate retry timers
Per log review with @Apollon77 

"I see the issue, because of the one change there are three attributes changed. I assume because of the subscription wait times where apple usually have "0s" as minimum time the subscription handler goes a bit crazy and fires three subscription notifies "in parallel" ... this alone would not be the issue ... the issue is that also afterwards the "empty" subscriptions are fired that way ... so we have duplicated timers. As I last time looked at that code I had such a suspicion but no time to look into it."

[HDBridge.txt](https://github.com/project-chip/matter.js/files/11807326/HDBridge.txt)

Apollon77:
This PR should fix the current issue with Subscription timer duplication that could happen if multiple data updates came at "the same timepoint".
The main fix would have been to move one line but I decided to also add some optimizations. I now added a 50ms delay on sending data updates to allow more data changes to be collected and submitted together - else only the first change would have been sent out, but the others (assuming when we have changes then multiple attributes can change "at once") would have been queued for next run - or in case of "0" still beeng ttiggered in parallel increasing comnmunication load to the controller.

I think this 50ms delay, even not specified in specs is a reasonable addition.

Additionally the Timer interface and implementations got enhanced by adding a "isRunning" propoerty to find out if a Timer is currently running/active or not.

fixes #185

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ All Changes without a GitHub Username in brackets are from the core team: @Apoll
   * Fix: Fix equality checks in Attribute servers to check deeper than just === (and introduce new util method isDeepEqual)
   * Fix: Make sure an error received from sending subscription seed data reports is not bubbling up and activate subscription after successful seeding
   * Fix: Allows Node.js Buffer objects to be persisted to storage as a Uint8Arrays that they subclass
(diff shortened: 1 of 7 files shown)


lauckhart (approved):
LGTM.  Do you think batching or transaction management will be warranted at some point?

PR #187, opened 2023-06-22
Restructure Endpoint generation to allow dynamic and stired structures

Apollon77:
This PR  is mainly updating endpoint structuring. Additionally some smaller things go also added that were not that easy to cut out again. I tried to separete these things in own commits, so the commits can be used for review.

All details:
* [Remove some console logs](https://github.com/project-chip/matter.js/commit/c68c989924d10257c23909ecfece3dc68b513a61)
* [Rename eventId property to id like on other servers](https://github.com/project-chip/matter.js/commit/c83b5e5e18259b82cf85a3fa962f3093bf10ecc4)
* [Steamline property naming for Event related structures](https://github.com/project-chip/matter.js/commit/57c6eebb02ba826683760798f4d0e4b086cda1a9)
* [Allow to define discoveryCapabilities structure when getting pairing code](https://github.com/project-chip/matter.js/commit/23880a6d1ef39e765f4b62f9e2899c2bd6ce906e)
* [Enhance Array type](https://github.com/project-chip/matter.js/commit/dd812bcf2c373973cf06d0192d038d04a810f3be)
* [Rework Endpoint structuring](https://github.com/project-chip/matter.js/commit/a74dc0e3fa0a481fe8ab52a8b01161930fc8dcf2)
  * Composed devices should only be used on an Aggregator
  * introduce a custom uniqueId when adding endpoints to be used for endpoint storage
  * add removeBridgedDevice to remove a device from a bridge
  * Restructure endpoint structuring to split responsibilities to where they belong
  * Structure changes are reported back via a structure-changed-callback to commissioning server to update structures on changes
  * update server/client-lists-attributes when adding/removing clusters
  * determine a unique endpoint id key from custom-unique id or serialnumber or unique id
  * leave logic to update partslist on endpoint
  * Add EndpointStructure class on InteractionServer to hold and handle all relevant endpoint structure data and move logic from endpoint to this
  * Add Event relevant methods and structures here too already (used later)
  * Endpoint structure is updated on structure changes
* [Update examples and add unique id parameter](https://github.com/project-chip/matter.js/commit/8ba5c002b0c1838d8964e7eef8b6d99aa9b194e8)
* [Changelog](https://github.com/project-chip/matter.js/commit/e90b00dcca168aaac81ce6cd98599ea3438cfa99)

closes https://github.com/project-chip/matter.js/discussions/173

### DIFF
--- a/packages/matter-node.js/README_EXAMPLES.md
+++ b/packages/matter-node.js/README_EXAMPLES.md
@@ -56,6 +56,7 @@ The following parameters are available:
 * -discriminator: the discriminator to use for pairing (default: 3840, value between 0 and 4095)
 * -vendorid: the vendor ID as number to use for pairing (default: 65521 (0xFFF1))
 * -productid: the product ID as number to use for pairing (default: 32768 (0x8000))
+* -uniqueid: a unique ID for this device to be used in naming and to stire structure information (default: ms time now)
 * -type: the device type to use for pairing (default: light, alternative value is "socket")
 * -announceinterface: limit mdns announcements to the provided network interface, e.g. "en0" (default: all interfaces available)
 * -port: the port to listen on for the device (default: 5540)
--- a/packages/matter-node.js/README_EXAMPLES.md
+++ b/packages/matter-node.js/README_EXAMPLES.md
@@ -56,7 +56,7 @@ The following parameters are available:
 * -discriminator: the discriminator to use for pairing (default: 3840, value between 0 and 4095)
 * -vendorid: the vendor ID as number to use for pairing (default: 65521 (0xFFF1))
 * -productid: the product ID as number to use for pairing (default: 32768 (0x8000))
-* -uniqueid: a unique ID for this device to be used in naming and to stire structure information (default: ms time now)
+* -uniqueid: a unique ID for this device to be used in naming and to store structure information (default: ms time now)
 * -type: the device type to use for pairing (default: light, alternative value is "socket")
 * -announceinterface: limit mdns announcements to the provided network interface, e.g. "en0" (default: all interfaces available)
 * -port: the port to listen on for the device (default: 5540)
--- a/packages/matter.js/src/CommissioningServer.ts
+++ b/packages/matter.js/src/CommissioningServer.ts
@@ -38,11 +38,17 @@ import { EthernetNetworkCommissioningCluster, NetworkCommissioningStatus } from 
 import { AdminCommissioningCluster, CommissioningWindowStatus } from "./cluster/AdminCommissioningCluster.js";
 import { GroupKeyManagementClusterHandler } from "./cluster/server/GroupKeyManagementServer.js";
 import { QrCode } from "./schema/QrCodeSchema.js";
-import { ComposedDevice } from "./device/ComposedDevice.js";
 import { Device } from "./device/Device.js";
 import { ByteArray } from "./util/ByteArray.js";
 import { NamedHandler } from "./util/NamedHandler.js";
 import { Attributes, Commands, Events } from "./cluster/Cluster.js";
+import { Logger } from "./log/Logger.js";
+import { Aggregator } from "./device/Aggregator.js";
+import { TypeFromBitSchema } from "./schema/BitmapSchema.js";
+import { Endpoint } from "./device/Endpoint.js";
+import { StorageContext } from "./storage/StorageContext.js";
+
+const logger = Logger.get("CommissioningServer");
 
 /**
  * Represents device pairing information.
@@ -111,6 +117,7 @@ export class CommissioningServer extends MatterNode {
     private readonly flowType: CommissionningFlowType;
 
     private storageManager?: StorageManager;
+    private endpointStructureStorage?: StorageContext;
     private mdnsScanner?: MdnsScanner;
     private mdnsBroadcaster?: MdnsBroadcaster;
 
@@ -319,7 +326,12 @@ export class CommissioningServer extends MatterNode {
      * Advertise the node via mDNS and start the commissioning process
      */
     async advertise() {
-        if (this.mdnsBroadcaster === undefined || this.mdnsScanner === undefined || this.storageManager === undefined) {
+        if (
+            this.mdnsBroadcaster === undefined ||
+            this.mdnsScanner === undefined ||
+            this.storageManager === undefined ||
+            this.endpointStructureStorage === undefined
+        ) {
             throw new Error("Add the node to the Matter instance before!");
         }
 
@@ -351,10 +363,13 @@ export class CommissioningServer extends MatterNode {
         const productId = basicInformation.attributes.productId.getLocal();
 
         this.interactionServer = new InteractionServer(this.storageManager);
-        const nextEndpointId = Math.max(this.rootEndpoint.findHighestEndpointId() + 1, this.nextEndpointId);
-        const lastEndpointId = this.rootEndpoint.ensureEndpointIds(nextEndpointId);
-        console.log(`Last endpoint id: ${lastEndpointId}`);
-        this.interactionServer.setRootEndpoint(this.rootEndpoint);
+
+        this.nextEndpointId = this.endpointStructureStorage.get("nextEndpointId", this.nextEndpointId);
+
+        this.ensureEndpointIds(); // Make sure to have unique endpoint ids
+        this.rootEndpoint.updatePartsList(); // initialize parts list of all Endpoint objects with final IDs
+        this.rootEndpoint.setStructureChangedCallback(() => this.updateStructure()); // Make sure we get structure changes
+        this.interactionServer.setRootEndpoint(this.rootEndpoint); // Initialize the interaction server with the root endpoint
 
         // TODO adjust later and refactor MatterDevice
         this.deviceInstance = new MatterDevice(this.deviceName, this.deviceType, vendorId, productId, this.discriminator, this.storageManager)
@@ -368,6 +383,81 @@ export class CommissioningServer extends MatterNode {
         }
 
         await this.deviceInstance.start();
+    }
+
+    updateStructure() {
+        logger.debug("Endpoint structure got updated ...");
+        this.ensureEndpointIds(); // Make sure to have unique endpoint ids
+        this.interactionServer?.setRootEndpoint(this.rootEndpoint); // Reinitilize the interaction server structure
+    }
+
+    getNextEndpointId(increase = true) {
+        if (increase) {
+            this.nextEndpointId++;
+        }
+        return this.nextEndpointId;
+    }
+
+    ensureEndpointIds() {
+        const rootUniqueIdPrefix = this.rootEndpoint.determineUniqueID();
+        this.initializeEndpointIdsFromStorage(this.rootEndpoint, rootUniqueIdPrefix);
+        this.fillAndStoreEndpointIds(this.rootEndpoint, rootUniqueIdPrefix);
+        this.endpointStructureStorage?.set("nextEndpointId", this.nextEndpointId);
+    }
+
+    private initializeEndpointIdsFromStorage(endpoint: Endpoint, parentUniquePrefix = "") {
+        if (this.endpointStructureStorage === undefined) {
+            throw new Error("endpointStructureStorage not set!");
+        }
+        const endpoints = endpoint.getChildEndpoints();
+        for (let endpointIndex = 0; endpointIndex < endpoints.length; endpointIndex++) {
+            let endpointUniquePrefix = parentUniquePrefix;
+            const endpoint = endpoints[endpointIndex];
+            const thisUniqueId = endpoint.determineUniqueID();
+            if (thisUniqueId === "") {
+                if (endpoint.id === undefined) {
+                    logger.debug(`No unique id found for endpoint on index ${endpointIndex} / device ${endpoint.name} - using index as unique identifier!`);
+                }
+                endpointUniquePrefix += `${endpointUniquePrefix === "" ? "" : "-"}index_${endpointIndex}`;
+            } else {
+                endpointUniquePrefix += `${endpointUniquePrefix === "" ? "" : "-"}${thisUniqueId}`;
+            }
+
+            if (endpoint.id === undefined) {
+                if (this.endpointStructureStorage.has(endpointUniquePrefix)) {
+                    endpoint.id = this.endpointStructureStorage.get<number>(endpointUniquePrefix);
+                    logger.debug(`Restored endpoint id ${endpoint.id} for endpoint with ${endpointUniquePrefix} / device ${endpoint.name} from storage`);
+                }
+            }
+            if (endpoint.id !== undefined && endpoint.id > this.nextEndpointId) {
+                this.nextEndpointId = endpoint.id + 1;
+            }
+            this.initializeEndpointIdsFromStorage(endpoint, endpointUniquePrefix);
+        }
+    }
+
+    private fillAndStoreEndpointIds(endpoint: Endpoint, parentUniquePrefix = "") {
+        if (this.endpointStructureStorage === undefined) {
+            throw new Error("endpointStructureStorage not set!");
+        }
+        const endpoints = endpoint.getChildEndpoints();
+        for (let endpointIndex = 0; endpointIndex < endpoints.length; endpointIndex++) {
+            let endpointUniquePrefix = parentUniquePrefix;
+            endpoint = endpoints[endpointIndex];
+            const thisUniqueId = endpoint.determineUniqueID();
+            if (thisUniqueId === "") {
+                endpointUniquePrefix += `${endpointUniquePrefix === "" ? "" : "-"}index_${endpointIndex}`;
+            } else {
+                endpointUniquePrefix += `${endpointUniquePrefix === "" ? "" : "-"}${thisUniqueId}`;
+            }
+
+            if (endpoint.id === undefined) {
+                endpoint.id = this.nextEndpointId++;
+                this.endpointStructureStorage.set(endpointUniquePrefix, endpoint.id);
+                logger.debug(`Assigned endpoint id ${endpoint.id} for endpoint with ${endpointUniquePrefix} / device ${endpoint.name} and stored it`);
+            }
+            this.fillAndStoreEndpointIds(endpoint, endpointUniquePrefix);
+        }
     }
 
     /**
@@ -437,14 +527,15 @@ export class CommissioningServer extends MatterNode {
      */
     setStorageManager(storageManager: StorageManager) {
         this.storageManager = storageManager;
+        this.endpointStructureStorage = this.storageManager.createContext("EndpointStructure")
     }
 
     /**
      * Add a new device to the node
      *
-     * @param device Device or ComposedDevice instance to add
+     * @param device Device or Aggregator instance to add
      */
-    addDevice(device: Device | ComposedDevice) {
+    addDevice(device: Device | Aggregator) {
         this.addEndpoint(device);
     }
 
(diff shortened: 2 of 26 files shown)


Apollon77:
Topics where I wouod be interested in feedback and potential alternativ ideas:
* method namings - I'm not too happy with some of the new method names -especially when it comes to that endpoint structure stuff ... so if anyone has better names ... tell :-)
* I thought about cutting out the "EndpointStructure" class into an own file but in the end it is specific to InteractionServer, so I ended in staying there.
* The unique IDs that are compiled get a bit long potentially ... see e.g. https://github.com/project-chip/matter.js/pull/187/files#diff-7bc184520022fa8b621504a4aea223c78eab76837920a5a88da0554e517809cbR1566 ... On one side this should not be an issue ... we could also shorted the "id source" ... ?!
* A big adjustment here was to separate the parts of it that belongs to endpoint vs the structure topics that are just relevant for the commissioingserver ... In the end it is a "revert a former change" where I moved many things to endpoint. In the end too much.

packages/matter.js/src/device/Endpoint.ts
```diff
@@ -169,33 +203,39 @@ export class Endpoint {
...
+                return `basicInfoUniqueId_${uniqueId}`;
+            }
+            const serialNumber = basicInformationCluster.getSerialNumberAttribute?.();
+            if (serialNumber !== undefined) {
+                return `basicInfoSerialNumber_${serialNumber}`;
+            }
         }
-        return nextEndpointId;
+        // Else we have no unique id
+        return "";
```
lauckhart, line 235: Why use `""` rather than `undefined` to indicate absence of an ID?

packages/matter.js/src/device/Endpoint.ts
```diff
@@ -21,28 +17,41 @@ import { FixedLabelCluster, UserLabelCluster } from "../cluster/LabelCluster.js"
 import { ClusterClientObj } from "../cluster/client/ClusterClient.js";
 import { ClusterServerObj, ClusterServerObjForCluster } from "../cluster/server/ClusterServer.js";
 import { InteractionClient } from "../protocol/interaction/InteractionClient.js";
-import { AllClustersMap } from "../cluster/index.js";
+import { AllClustersMap, BridgedDeviceBasicInformationCluster } from "../cluster/index.js";
+import { BasicInformationCluster } from "../cluster/BasicInformationCluster.js";
+
+export interface EndpointOptions {
+    endpointId?: number;
+    uniqueId?: string;
```
lauckhart, line 25: endpointId is also unique I think?  What about something like "persistentId" or "deviceIdentificationKey"?

packages/matter.js/src/device/Endpoint.ts
```diff
@@ -21,28 +17,41 @@ import { FixedLabelCluster, UserLabelCluster } from "../cluster/LabelCluster.js"
...
+    /**
+     * Create a new Endpoint instance.
+     *
+     * @param deviceTypes One or multiple DeviceTypeDefinitions of the endpoint
+     * @param options Options for the endpoint
+     */
     constructor(
         protected deviceTypes: AtLeastOne<DeviceTypeDefinition>,
-        endpointId?: number
+        options: EndpointOptions = {}
```
lauckhart, line 47: Could this be `endpoint: number | EndpointOptions`?  Then other changes converting `4` to `{ endpointId: 4}` would be unnecessary.  Might be clearer if we had an `type EndpointId = number` so it could be `endpoint: EndpointId | EndpointOptions`.

packages/matter.js/src/device/Endpoint.ts
```diff
@@ -53,9 +62,22 @@ export class Endpoint {
...
-            this.id = endpointId;
+        if (options.endpointId !== undefined) {
+            this.id = options.endpointId;
         }
+        if (options.uniqueId !== undefined) {
+            this.uniqueId = options.uniqueId;
+        }
+    }
+
+    setStructureChangedCallback(callback: () => void) {
```
lauckhart, line 73: Do we have a standard way of handling events?  E.g. `onStructureChanged()` or `structureChanged.on`?  Or is this just used internally?

packages/matter.js/src/device/Endpoint.ts
```diff
@@ -53,9 +62,22 @@ export class Endpoint {
...
+        this.structureChangedCallback = callback;
+        this.childEndpoints.forEach(endpoint => endpoint.setStructureChangedCallback(callback));
```
lauckhart, line 75: Could these events bubble up from child endpoints to their parent automatically?  Might simplify state management.  Not sure there's a huge advantage but that's you wouldn't need the forEach here

packages/matter.js/src/protocol/interaction/InteractionServer.ts
```diff
@@ -371,17 +388,248 @@ export function attributePathToId({ endpointId, clusterId, attributeId }: TypeFr
...
-    return value === undefined ? "*" : `0x${value.toString(16)}`;
+export function eventPathToId({ endpointId, clusterId, eventId }: TypeFromSchema<typeof TlvEventPath>) {
+    return `${endpointId}/${clusterId}/${eventId}`;
 }
 
-export class InteractionServer implements ProtocolHandler<MatterDevice> {
+export class EndpointStructure {
+    endpoints = new Map<number, Endpoint>();
+    attributes = new Map<string, (AttributeServer<any> | FabricScopedAttributeServer<any> | FixedAttributeServer<any>)>();
+    attributePaths = new Array<AttributePath>();
```
lauckhart, line 398: I'm wondering if attributes could be saved as AttributeWithPath?  Would make state management simpler and the search code wouldn't have to do the secondary lookup from attributePaths -> attributes[attributePathToId(path)]

packages/matter.js/src/protocol/interaction/InteractionServer.ts
```diff
@@ -371,17 +388,248 @@ export function attributePathToId({ endpointId, clusterId, attributeId }: TypeFr
...
+        if (endpoint.id === undefined) {
+            throw new Error(`Endpoint ID is undefined. It needs to be initialized first!`);
+        }
+
+        endpoint.verifyRequiredClusters();
+
+        for (const cluster of endpoint.getAllClusterServers()) {
+            const { id: clusterId, attributes: clusterAttributes, _events: clusterEvents, _commands: clusterCommands } = cluster;
+            // Add attributes
+            for (const name in clusterAttributes) {
```
lauckhart, line 439: Could do this and following two loops in a generic function.

packages/matter.js/src/protocol/interaction/InteractionServer.ts
```diff
@@ -371,17 +388,248 @@ export function attributePathToId({ endpointId, clusterId, attributeId }: TypeFr
...
+        if (this.endpoints.has(endpoint.id)) throw new Error(`Endpoint ID ${endpoint.id} exists twice`);
+
+        this.endpoints.set(endpoint.id, endpoint);
+    }
+
+    toHex(value: number | undefined) {
+        return value === undefined ? "*" : `0x${value.toString(16)}`;
+    }
+
+    resolveAttributeName({ endpointId, clusterId, attributeId }: TypeFromSchema<typeof TlvAttributePath>) {
```
lauckhart, line 472: OK now there's really getting to be a lot of duplicate code.  Ideally maybe move to a generic "ElementPath" type where the path becomes `{ elementType, endpointId, clusterId, attributeId }`?

If that's too big of a change, maybe at least pull endpoint/cluster lookup into a this.resolveCluster(endpointId, clusterId) function.

packages/matter.js/src/protocol/interaction/InteractionServer.ts
```diff
@@ -371,17 +388,248 @@ export function attributePathToId({ endpointId, clusterId, attributeId }: TypeFr
...
+
+        if (commandId === undefined) {
+            return `${endpointName}/${clusterName}/*`;
+        }
+        const command = this.commands.get(commandPathToId({ endpointId, clusterId, commandId }));
+        const attributeName = `${command?.name ?? "unknown"}(${this.toHex(commandId)})`;
+        return `${endpointName}/${clusterName}/${attributeName}`;
+    }
+
+    getAttributes(filters: TypeFromSchema<typeof TlvAttributePath>[], onlyWritable = false): AttributeWithPath[] {
```
lauckhart, line 553: Similar comment to above although I know it's a bit more tricky with the onlyWritable specialization for attributes.

packages/matter.js/src/device/Aggregator.ts
```diff
@@ -17,26 +18,19 @@ import { ClusterServer } from "../protocol/interaction/InteractionServer.js";
...
  * and PowerSourceCluster) to the device!
  */
-export class Aggregator extends ComposedDevice {
+export class Aggregator extends Endpoint {
     /**
      * Create a new Aggregator instance and optionally directly add devices to it. If this is used the devices must
      * already have the BridgedDeviceBasicInformationCluster added!
      * @param devices Array of devices to add
-     * @param endpointId Optional endpoint ID to use. If not provided will automatically be assigned
+     * @param options Optional Endpoint options
```
lauckhart, line 26: Optional Endpoint options seems a bit redundant ;)

packages/matter-node.js/README_EXAMPLES.md
```diff
@@ -56,6 +56,7 @@ The following parameters are available:
 * -discriminator: the discriminator to use for pairing (default: 3840, value between 0 and 4095)
 * -vendorid: the vendor ID as number to use for pairing (default: 65521 (0xFFF1))
 * -productid: the product ID as number to use for pairing (default: 32768 (0x8000))
+* -uniqueid: a unique ID for this device to be used in naming and to stire structure information (default: ms time now)
```
lauckhart, line 59: I'm not familiar with "stire", is this a typo for "store"?

lauckhart (commented):
Looks good in general.  I made a bunch of comments but not anything very big.

packages/matter.js/src/protocol/interaction/InteractionServer.ts
```diff
@@ -371,17 +388,248 @@ export function attributePathToId({ endpointId, clusterId, attributeId }: TypeFr
...
 }
 
-function toHex(value: number | undefined) {
-    return value === undefined ? "*" : `0x${value.toString(16)}`;
+export function eventPathToId({ endpointId, clusterId, eventId }: TypeFromSchema<typeof TlvEventPath>) {
+    return `${endpointId}/${clusterId}/${eventId}`;
 }
 
-export class InteractionServer implements ProtocolHandler<MatterDevice> {
+export class EndpointStructure {
```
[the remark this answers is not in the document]
Apollon77: Renamed to InteractionEndpointStructure and extracted

packages/matter.js/src/CommissioningServer.ts
```diff
@@ -370,6 +385,81 @@ export class CommissioningServer extends MatterNode {
...
+    ensureEndpointIds() {
+        const rootUniqueIdPrefix = this.rootEndpoint.determineUniqueID();
+        this.initializeEndpointIdsFromStorage(this.rootEndpoint, rootUniqueIdPrefix);
+        this.fillAndStoreEndpointIds(this.rootEndpoint, rootUniqueIdPrefix);
+        this.endpointStructureStorage?.set("nextEndpointId", this.nextEndpointId);
+    }
+
+    private initializeEndpointIdsFromStorage(endpoint: Endpoint, parentUniquePrefix = "") {
+        if (this.endpointStructureStorage === undefined) {
+            throw new Error("endpointStructureStorage not set!");
```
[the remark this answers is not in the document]
Apollon77: Changed to "Storage manager must be initialized to enable initialization from storage." ... because the endpointStructureStorage is automatically initialized with the storage manager

packages/matter.js/src/device/Endpoint.ts
Apollon77: ok

packages/matter.js/src/device/Endpoint.ts
Apollon77: In fact with this you explicitely set the endpoint-ID (aka id attrobute of the endpoint). If you provide duplicates there is no correction or such  and you get an error, so it really is the endpoint ID. if you provide this this is also not stored in any way. So name intended

lauckhart: Oh sorry wasn't clear, I meant instead of "uniqueID"

packages/matter.js/src/device/Endpoint.ts
Apollon77: The main point here was to change from "sigle parameters that alone are optional to a options structure, so it is more easy to extend later if needed. because else it would end in stzarnge calls like "(..., undefined, "uniqe-id") just because you do not set the endpoint, id. I think using an object is common in javascript and also our lib already

packages/matter.js/src/device/Endpoint.ts
Apollon77: Currently we use callback style ... also eg to store Fabrics or also to inform upper layers that eg a connection is closed to cleanup. So right now no events ....

packages/matter.js/src/device/Endpoint.ts
lauckhart: Yeah makes sense, I was just thinking `Endpoint(deviceTypes, 1)` would be a shortcut for `Endpoint(deviceTypes, { endpointId: 1 })`

packages/matter.js/src/device/Endpoint.ts
Apollon77: For now we need to update too many data (partsLists and full list of endpoints and attributes), so thats why I decided to just inform the Commissioning server whoch then does it "down the structure" again. So bubbling through the structure would bring no benefit here.

packages/matter.js/src/protocol/interaction/InteractionServer.ts
Apollon77: AttributeWIthPath would bring no benefit becauseit would be in the data. with that "string key that contains "endpointId/clusterId/attributeId" is a direct match ... else you would need to search any time

lauckhart: Right, I didn't mean to imply get rid of the string key, just have it point to AttributeWithPath instead of Attribute

packages/matter.js/src/protocol/interaction/InteractionServer.ts
Apollon77: you mean because attrbtes/events/commands are basically "samie-ish". Yes but in fact the "toId" functions and both structures are all different, so wouod all be configurable ... Fund it easier to have "smeish" code :-)

Apollon77: But yes good Idea will think tomorrow about it

packages/matter.js/src/device/Aggregator.ts
Apollon77: :-) yes but name, and comment :-)

packages/matter-node.js/README_EXAMPLES.md
Apollon77: fixed

packages/matter.js/src/protocol/interaction/InteractionServer.ts
Apollon77: exactly thi was my main topic ... but we could post process. WIll think

packages/matter.js/src/device/Aggregator.ts
lauckhart: Yeah just meant the comment, could be "Optional endpoint settings" or "Optional endpoint parameters".  Anyway pretty trivial, going to mark resolved

packages/matter.js/src/device/Endpoint.ts
Apollon77: uniqueEndpointKey?
uniqueEndpointStorageKey?

packages/matter.js/src/device/Endpoint.ts
Apollon77: In fact I hope that noone will manage endpoint Ids himelf :-)) so the uniqueEndpointStorageKey-string woud be more the "first parameter" but Ialso do not want to limit devs.
But in general a "string | EndpointOptions" is a meehhh typing :-)

packages/matter.js/src/device/Endpoint.ts
lauckhart: Yeah those work, the "endpoint" part might be implied by fact it's in endpoint options but uniqueStorageKey or uniqueEndpointStorageKey seem fine

packages/matter.js/src/device/Endpoint.ts
lauckhart: Cool, I was wondering if the key would be dominant.  I prefer "meh typing" over "meh usage" but we can revisit once API usage becomes more clear as it would be backward compatible

packages/matter.js/src/device/Endpoint.ts
Apollon77: kk

lauckhart: I just realized that you were using the terminology from BasicInformationCluster.  So maybe your original uniqueId is the better choice if it maps directly into BasicInformation.UniqueID.  Maybe a /** comment would be helpful

lauckhart:
>     * The unique IDs that are compiled get a bit long potentially ... see e.g. https://github.com/project-chip/matter.js/pull/187/files#diff-7bc184520022fa8b621504a4aea223c78eab76837920a5a88da0554e517809cbR1566 ... On one side this should not be an issue ... we could also shorted the "id source" ... ?!

Forgot to respond to this part, not sure there's anything too bad about the long identifiers but if they'll appear in logs a lot maybe worth shortening.  Here's some ideas:

`basicInfoSerialNumber_node-matter-0000-customUniqueId_test-unique-id` -> `serial:node-matter-0000/custom:test-unique-id`
`basicInfoSerialNumber_node-matter-0000-index_0-customUniqueId_3333` -> `serial:node-matter-0000/[0]/custom:test-unique-id`

packages/matter.js/src/device/Endpoint.ts
Apollon77: All good the renaming mames sense because it is "Just a custom such unique id" ...so we can also rename it

packages/matter.js/src/protocol/interaction/InteractionServer.ts
Apollon77: But Attributeath is something different then AttributeServerWithPath :-) so another set of types ... would prefer to leave it for now

packages/matter.js/src/protocol/interaction/InteractionServer.ts
Apollon77: I was able to generalize this ... see next commit

packages/matter.js/src/protocol/interaction/InteractionServer.ts
Apollon77: But here it is more hard and bgger changes and typing issues ... owuld leav" slighly duplicated code" for now

Apollon77:
I decided against ":" because will be used as potentially filename :-) Thats why "_" renaming make sense

Apollon77:
@lauckhart please re-review. Updated

lauckhart (approved):
LGTM

lauckhart:
> I decided against ":" because will be used as potentially filename :-) Thats why "_" renaming make sense

Yeah I figured something like that.  Character mix isn't super important I was just offering up shorter versions.  Although I do assume files would be urlEncoded or somesuch regardless so we don't need to constrain user IDs.

PR #188, opened 2023-06-26
Split out examples to own package

Apollon77:
As discussed it makes, also in preparaton of enhancement packages like for ble, sense to split out the exmaples into an own package.


lauckhart (commented):
Took a quick scan, looks fine to me.  One question, could we maintain LICENSE, CODE_OF_CONDUCT.md and CONTRIBUTING.md in the root rather than in each submodule?

Apollon77:
We discussed that and decided to leave it formally in each package

lauckhart (approved):

PR #189, opened 2023-06-27
Fix BTP Package decoding logic

Apollon77:
While doing further testing I came across some decoding bugs where optional fields were not parsed correctly conditionally

### DIFF
--- a/packages/matter.js/test/codec/BtpTest.ts
+++ b/packages/matter.js/test/codec/BtpTest.ts
@@ -36,9 +36,9 @@ const DECODED_PAYLOAD = {
     },
     payload: {
         ackNumber: 0,
-        sequenceNumber: 68, // 4f
-        messageLength: 1024, // 0400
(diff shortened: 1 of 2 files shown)


packages/matter.js/test/codec/BtpTest.ts
```diff
@@ -37,7 +37,7 @@ const DECODED_PAYLOAD = {
     payload: {
         ackNumber: 0,
         sequenceNumber: 68, // 4f
-        msgLength: 1024, // 0400
+        messageLength: 1024, // 0400
         segmentPayload: ByteArray.fromHex("00000049b6a902a9a5773dbb8cafd90120a7c7000015300120cb0c120a3499327ddaec4ebe60889df0f1bf80d8a4dea1dd6ffef16ef58ecafe25028e17240300280418"),
```
turon, line 41: Are tests supposed to be cross-platform (can run on web?)
If so ByteArray should be UInt8Array.

turon (approved):

packages/matter.js/test/codec/BtpTest.ts
Apollon77: byteArray is an alias for UInt8Array, so all fine

PR #194, opened 2023-07-09
Enhance MessageCodec and check some more fields

Apollon77:
This PR adds parsing out some Security Flags for the matter message and checks them. We currently do not suport the handling for them but so we see

### DIFF
--- a/packages/matter.js/src/codec/MessageCodec.ts
+++ b/packages/matter.js/src/codec/MessageCodec.ts
@@ -13,6 +13,9 @@ import { DiagnosticDictionary } from "../log/Logger.js";
 export interface PacketHeader {
     sessionId: number,
     sessionType: SessionType,
+    hasPrivacyEnhancements: boolean,
+    isControlMessage: boolean,
+    hasMessageExtensions: boolean,
     messageId: number,
     sourceNodeId?: NodeId,
     destNodeId?: NodeId,
@@ -62,6 +65,12 @@ const enum PayloadHeaderFlag {
     RequiresAck = 0b00000100,
     HasSecureExtension = 0b00001000,
     HasVendorId = 0b00010000,
+}
+
+const enum SecurityFlag {
+    HasPrivacyEnhancements = 0b10000000,
+    IsControlMessage = 0b01000000,
+    HasMessageExtension = 0b00100000,
 }
 
 export class MessageCodec {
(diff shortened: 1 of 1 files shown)


packages/matter.js/src/codec/MessageCodec.ts
```diff
@@ -122,8 +131,14 @@ export class MessageCodec {
 
         const sessionType = securityFlags & 0b00000011;
         if (sessionType !== SessionType.Group && sessionType !== SessionType.Unicast) throw new Error(`Unsupported session type ${sessionType}`);
-
-        return { sessionId, sourceNodeId, messageId, destGroupId, destNodeId, sessionType };
+        const hasPrivacyEnhancements = (securityFlags & SecurityFlag.HasPrivacyEnhancements) !== 0;
+        if (hasPrivacyEnhancements) throw new Error(`Privacy enhancements not supported`);
```
lauckhart, line 135: Should this be MatterError rather than Error?

packages/matter.js/src/codec/MessageCodec.ts
```diff
@@ -13,6 +13,9 @@ import { DiagnosticDictionary } from "../log/Logger.js";
 export interface PacketHeader {
     sessionId: number,
     sessionType: SessionType,
+    hasPrivacyEnhancements: boolean,
+    isControlMessage: boolean,
+    hasMessageExtensions: boolean,
```
lauckhart, line 18: If I'm reading correctly these aren't actually used anywhere because we throw rather than set them.  Maybe you could set them in MessageCodec and throw the "unsupported" message at a higher level?

packages/matter.js/src/codec/MessageCodec.ts
Apollon77: The other errors Thrown here are also no MatterErrors. i will put on my list and do one big "all is MatterError" PR separately

packages/matter.js/src/codec/MessageCodec.ts
lauckhart: Yeah I think they're good to have in.  Since you already retrieve the values (but then immediately throw) I was thinking you could instead set them in PacketHeader and have whoever reads PacketHeader do the throw instead.  NBD though

lauckhart (approved):

PR #196, opened 2023-07-09
Make Broadcasting port dynamic and add UDC Announce

Apollon77:
This PR is preparing future adjustments and:
* make the ports announced by Broadcaster configurable
* adds a missing announcement type for a "commissioner" to Broadcaster (not yet used, will also come, but that one will be on a different port)

### DIFF
--- a/packages/matter.js/src/mdns/MdnsBroadcaster.ts
+++ b/packages/matter.js/src/mdns/MdnsBroadcaster.ts
@@ -31,17 +31,7 @@ export class MdnsBroadcaster implements Broadcaster {
         private readonly mdnsServer: MdnsServer,
     ) { }
 
-    /**
-     * Set the Broadcaster data to announce a device ready for commissioning in a special mode
-     *
-     * @param mode
-     * @param deviceName
-     * @param deviceType
-     * @param vendorId
-     * @param productId
-     * @param discriminator
-     * @param announcedNetPort
-     */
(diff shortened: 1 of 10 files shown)


packages/matter.js/src/mdns/MdnsBroadcaster.ts
```diff
@@ -126,6 +144,58 @@ export class MdnsBroadcaster implements Broadcaster {
...
     }
 
+    /**
+     * Set the Broadcaster data to announce a Commissioner (aka Commissioner discovery)
+     *
+     * @param deviceName
+     * @param vendorId
+     * @param productId
+     * @param announcedNetPort
+     * @param deviceType
```
lauckhart, line 154: Should probably delete these

packages/matter.js/src/mdns/MdnsBroadcaster.ts
```diff
@@ -18,18 +21,28 @@ import { Fabric } from "../fabric/Fabric.js";
...
+    /**
+     * Set the Broadcaster data to announce a device ready for commissioning in a special mode
+     *
+     * @param mode
+     * @param deviceName
+     * @param deviceType
+     * @param vendorId
+     * @param productId
+     * @param discriminator
+     * @param announcedNetPort
```
lauckhart, line 43: Should probably delete these if not going to fill them out

lauckhart (approved):

PR #199, opened 2023-07-10
Fix dynamic Structure changes for bridges

Apollon77:
We should also update partsLists entries for the changes to get pushed to the controllers. tested with Apple

Open is that we also need to update all subscriptions that might now miss newly added things. Todo for later


JimBuzbee:
With this fix, dynamically adding a new device to a bridge works well with Apple. With Google, seems to take a long time for a new device added to show up, but works as well.

Apollon77:
For google maybe reworking subscriptions will help ... I need to see in separate PR. Thank you @JimBuzbee for testing

lauckhart (approved):

PR #207, opened 2023-07-14
Optimize RegulatoryConfig commissioning

Apollon77:
This is another run to make commissioning of regulatory config really spec conform

Core Spec 5.5 state it that way.

PS: I will move the whole commissioning soon into an own Class that will manage that process better. for now lets keep it that way

### DIFF
--- a/packages/matter-node.js/test/IntegrationTest.ts
+++ b/packages/matter-node.js/test/IntegrationTest.ts
@@ -15,19 +15,20 @@ Crypto.get = () => new CryptoNode();
 import {
     OnOffCluster, BasicInformationCluster, OperationalCertStatus, OperationalCredentialsCluster, DescriptorCluster,
     IdentifyCluster, GroupsCluster, AccessControlCluster, ScenesCluster, GeneralCommissioningCluster,
-    RegulatoryLocationType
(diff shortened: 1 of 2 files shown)


packages/matter-node.js/test/IntegrationTest.ts
```diff
@@ -116,6 +117,23 @@ describe("Integration Test", () => {
...
 
+        // Override NetworkCommissioning Cluster for now unless configurable
+        commissioningServer.addRootClusterServer(
+            ClusterServer(
+                WifiAndEthernetAndThreadNetworkCommissioningCluster,
+                {
+                    maxNetworks: 1,
+                    interfaceEnabled: true,
+                    lastConnectErrorValue: 0,
+                    lastNetworkId: ByteArray.fromHex("0000000000000000000000000000000000000000000000000000000000000000"),
```
lauckhart, line 128: I think you can use new ByteArray(length) to get a zero-initialized ByteArray

lauckhart (approved):

packages/matter-node.js/test/IntegrationTest.ts
Apollon77: Yes but this is also that way in oher places. Need to clean that up soon anywaay, so decided to stay that way.

PR #210, opened 2023-07-16
Return AttributeStatus correctly for Read Requests

Apollon77:
The current logic was not correctly responsing to unsupported path's on read requests. This PR fixes this.

Fixes #128 

* for non-Wildcard path's (aka concrete paths) the correct error-status needs to be returned (depending on unsupported endpoint, cluster or attribute)
* for Wildcard path's such erors are simply ignored and just data not added

I verified it and a pairing with Apple - see #128 - is not including error 128 anymore.

Comparable PRs will also come for write/subscribe/invoke soon (or might be added here based on timing)

### DIFF
--- a/packages/matter.js/src/protocol/interaction/InteractionEndpointStructure.ts
+++ b/packages/matter.js/src/protocol/interaction/InteractionEndpointStructure.ts
@@ -6,7 +6,7 @@
 
 import { Endpoint } from "../../device/index.js";
 import {
-    AttributeServer, CommandServer, FabricScopedAttributeServer, FixedAttributeServer
+    AttributeServer, ClusterServerObj, CommandServer, FabricScopedAttributeServer, FixedAttributeServer
 } from "../../cluster/index.js";
 import { EventServer } from "../../cluster/server/EventServer.js";
 import { TypeFromSchema } from "../../tlv/index.js";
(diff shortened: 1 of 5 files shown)


packages/matter.js/src/protocol/interaction/InteractionEndpointStructure.ts
```diff
@@ -132,6 +132,30 @@ export class InteractionEndpointStructure {
...
+    hasEndpoint(endpointId: number): boolean {
+        return this.endpoints.has(endpointId);
+    }
+
+    getClusterServer(endpointId: number, clusterId: number): ClusterServerObj<any, any, any> | undefined {
+        const endpoint = this.endpoints.get(endpointId);
+        if (endpoint === undefined) {
+            return undefined;
+        }
+        return endpoint.getClusterServerById(clusterId);
```
lauckhart, line 148: NBD but might write as a one-liner: `this.getEndpoint(endpointId)?.getClusterServerById(clusterId)`

packages/matter.js/src/protocol/interaction/InteractionEndpointStructure.ts
```diff
@@ -132,6 +132,30 @@ export class InteractionEndpointStructure {
...
+    }
+
+    hasClusterServer(endpointId: number, clusterId: number): boolean {
+        const endpoint = this.endpoints.get(endpointId);
+        if (endpoint === undefined) {
+            return false;
+        }
+        return !!endpoint.getClusterServerById(clusterId);
```
lauckhart, line 156: Could be one-liner: `return !!this.getClusterServer(endpointId, clusterId)`

lauckhart (approved):

packages/matter.js/src/protocol/interaction/InteractionEndpointStructure.ts
Apollon77: good idea, change

packages/matter.js/src/protocol/interaction/InteractionEndpointStructure.ts
Apollon77: same, thx

PR #216, opened 2023-07-17
Default storagenames use "." at the beginning

Apollon77:
... so they can be hided in explorers and also for git.


turon (approved):
LGTM :+1:

PR #224, opened 2023-07-22
Conditional elements in "complete" clusters

lauckhart:
This modifies the "complete" clusters (clusters that have all elements regardless of feature selection) to use the "optionalIf" and "mandatoryIf" constructs for runtime validation.

Also includes a few other modest changes.  See individual commits.


Apollon77 (approved):
Awesome and great!

Approved (an branch update also kicked the two strange changes from other commit)

PR #229, opened 2023-07-25
Adjust Library to use the new generated Clusters and remove old

Apollon77:
The PR removes the former included Cluster files and converts anything to use the generated Clusters.


Apollon77:
@lauckhart There is one last thing broken ... it seems that the supportedFeatures is set incorrectly when using "with" to select features ...

So with this code
Bildschirmfoto 2023-07-26 um 12 42 53

(same result if I use the strings) when reading that on Server/Device side it logs

> 2022-10-25 03:56:40.000 DEBUG InteractionProtocol Read from udp://fdce:7c65:b2dd:7d46:923f:8a53:eb6c:cafe:19045 on session secure/29003: MA-rootdevice(0x0)/NetworkCommissioning(0x31)/featureMap(0xfffc)={"w":false,"t":false,"e":false} (version=0)

(so wrong because all false instead of true, also also strange is that "just initial lowercase letter).

The CLient/Controller then receives

> 2022-10-25 03:56:40.000 DEBUG InteractionClient Received read response: 0x0/NetworkCommissioning(0x31)/featureMap(0xfffc) = {"wiFiNetworkInterface":false,"threadNetworkInterface":false,"ethernetNetworkInterface":false}

Which in fact should be wrong.

You should be able to validate that by using the branch and running the test/IntegrationTest.ts in matter-node.js package.

packages/matter.js/src/cluster/ClusterHelper.ts
```diff
@@ -4,69 +4,85 @@
...
+    [Clusters.ThermostatUserInterfaceConfigurationCluster.id]: Clusters.ThermostatUserInterfaceConfigurationCluster,
+    [Clusters.ThreadNetworkDiagnosticsCluster.id]: Clusters.ThreadNetworkDiagnostics.Complete,
+    [Clusters.TimeFormatLocalizationCluster.id]: Clusters.TimeFormatLocalization.Complete,
+    [Clusters.TimeSyncCluster.id]: Clusters.TimeSync.Complete,
+    [Clusters.UnitLocalizationCluster.id]: Clusters.UnitLocalization.Complete,
+    [Clusters.UserLabelCluster.id]: Clusters.UserLabelCluster,
+    [Clusters.ValidProxiesCluster.id]: Clusters.ValidProxiesCluster,
+    [Clusters.WakeOnLanCluster.id]: Clusters.WakeOnLanCluster,
+    [Clusters.WiFiNetworkDiagnosticsCluster.id]: Clusters.WiFiNetworkDiagnostics.Complete,
+    [Clusters.WindowCoveringCluster.id]: Clusters.WindowCovering.Complete,
```
lauckhart, line 85: Hrm not as easy to populate programmatically as I was thinking.  I'll add something to make that easier but we can do after merge.

lauckhart (approved):

packages/matter.js/src/cluster/ClusterHelper.ts
Apollon77: Yes I also wanted to ask. But ok for now.

PR #233, opened 2023-07-26
Certification readiness overview

Apollon77:


lauckhart:
Good information to track, what do you think about using the wiki for stuff like this?  On one hand it'd be nice to have updates tied to source...  On the other hand we'd have to commit to keeping it updated or it'd get bit rot.  So more casual maintenance via wiki might make sense.

Could of course check in now and revisit once we see how we do.

lauckhart:
Another option could be to open a few topic issues with task lists.

Apollon77:
Sure, this was just to have one place as overview. Then I would say two options

1.) We create a new discussion type (like the feature requests now) like "certification relevant todos"
2.) We use the project in github we have and create all these as entries there and label them with "cert" relevant

I did not really found a good way to make todos and status transparent (we also have the TODO.md basically) ... in fact we right now have 2 devs actively working ... you (@lauckhart) and me :-) Me on the core side, and you on generation right now ... so it is more about how we organize work here... (and for future ith more devs)

Apollon77:
Converted to Project entries

PR #234, opened 2023-07-26
Testing for ClusterFactory and related fixes

lauckhart:
Fairly exhaustive testing for ClusterFactory.  Fixed one for-sure bug and two other things that might've been problematic.

Resolves issue in recent comment in #229


Apollon77 (approved):

Apollon77:
Thank you!

PR #241, opened 2023-07-30
FabricIndex support

### LINKED ISSUE #239: Clusters missing "fabricIndex" field whendata are fabricScoped
When data are defined as "Fabric scoped" they **automaticllay** include a fabricIndex field with the field id 0xFE. We miss them.

This field is mainly relevant For fabrics field in OperationalCredentialsCLuster FabricDescriptionStruct and leading to Amazon no continueing commissioning process.

lauckhart:
Generates fabricIndex field for all structures that should have it.  Modifies AttributeServer and InteractionServer as necessary to set the field to the appropriate value.

fixes #239

Scan the individual commits for details.

### DIFF
--- a/packages/matter.js/src/cluster/server/AttributeServer.ts
+++ b/packages/matter.js/src/cluster/server/AttributeServer.ts
@@ -12,6 +12,18 @@ import { Endpoint } from "../../device/Endpoint.js";
 import { Cluster } from "../Cluster.js";
 import { Fabric } from "../../fabric/Fabric.js";
 import { isDeepEqual } from "../../util/DeepEqual.js";
+import { FabricIndex } from "../../datatype/FabricIndex.js";
+import { ArraySchema } from "../../tlv/TlvArray.js";
+import { NullableSchema } from "../../tlv/TlvNullable.js";
+import { FieldType, ObjectSchema } from "../../tlv/TlvObject.js";
+import { MatterError } from "../../common/MatterError.js";
+import { Globals } from "../../model/index.js";
+
+/**
+ * Thrown when an operation cannot complete because fabric information is
+ * unavailable.
+ */
+export class FabricScopeError extends MatterError { };
 
 export abstract class BaseAttributeServer<T> {
     protected value: T;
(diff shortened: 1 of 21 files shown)


packages/matter.js/src/cluster/server/AttributeServer.ts
```diff
@@ -64,21 +75,73 @@ export class FixedAttributeServer<T> extends BaseAttributeServer<T> {
...
+                }
+            }
+        } else if (schema instanceof NullableSchema) {
+            this.setFabricIndex(value, fabricIndex, schema.schema);
+        } else if (schema instanceof ObjectSchema) {
+            if (typeof schema !== "object") {
+                return;
+            }
+
+            for (const k in schema.fieldDefinitions) {
```
Apollon77, line 117: maybe a find is better here? Or at least a break after id found because dplucate IDs should not happen :-)

packages/matter.js/src/cluster/server/AttributeServer.ts
```diff
@@ -64,21 +75,73 @@ export class FixedAttributeServer<T> extends BaseAttributeServer<T> {
...
+            }
+        }
+    }
+}
+
 /**
  * Attribute server for normal attributes that can be read and written.
  */
-export class AttributeServer<T> extends BaseAttributeServer<T> {
+export class AttributeServer<T> extends MutableAttributeServer<T> {
```
Apollon77, line 138: Hm ... hsould not all fabric scoped data structs be contained within FabricScoped attributes?  Then this should be only for the Fabric scioped class? Then maybe even no extra class needed

packages/matter.js/src/cluster/server/AttributeServer.ts
lauckhart: It continues the loop because if it's not 0xfe it recurses into subfields

packages/matter.js/src/cluster/server/AttributeServer.ts
lauckhart: I partially created the intermediate class so ACL checks could be in once place.  I didn't see in the spec where they specify fabric scoped fields can't appear under non-scoped attributes but you might be right there

packages/matter.js/src/cluster/server/AttributeServer.ts
Apollon77: Ahhh yes. Kk

packages/matter.js/src/cluster/server/AttributeServer.ts
Apollon77: Ok

Apollon77:
This PR will be integrated into #240 tomorrow

PR #250, opened 2023-08-01
WindowCovering tweaks & testing

lauckhart:
Tweaks a couple of names ("100ths" now camelizes as "100ths" and not "100Ths").

Fixes a conformance rule that was incorrectly matching pattern "FOO & BAR".  Makes a few attributes mandatory that were previously optional.

Adds a WindowCovering-specific test, fairly exhaustive but currently for a fairly limited feature set.


vves:
Thanks for the Ths to ths fix.

Apollon77 (approved):

vves (approved):

PR #261, opened 2023-08-08
Populate model w/ device types from specification

lauckhart:
Extends DeviceTypeModel tree a bit, loads & translates device types from specification into spec.ts and adds DeviceType models to standard elements.

Not used for anything yet.


Apollon77 (approved):

Apollon77:
Thank you ... this will be useful :-)

PR #264, opened 2023-08-09
Return correct Error codes on subscription initialization

Apollon77:
This PR adjusts (as we did for Read, Write/invoke earlier already) the error handing and detection also for subscriptions to return the correct status in all cases

* addresses https://github.com/orgs/project-chip/projects/11/views/1?pane=issue&itemId=34442421

### DIFF
--- a/packages/matter.js/src/protocol/interaction/SubscriptionHandler.ts
+++ b/packages/matter.js/src/protocol/interaction/SubscriptionHandler.ts
@@ -13,7 +13,10 @@ import {
 import { Logger } from "../../log/Logger.js";
 import { Time, Timer } from "../../time/Time.js";
 import { NodeId } from "../../datatype/NodeId.js";
-import { StatusCode, TlvAttributePath, TlvEventFilter, TlvEventPath, TlvEventReport } from "./InteractionProtocol.js";
+import {
+    StatusCode, TlvAttributePath, TlvAttributeReport, TlvAttributeStatus, TlvEventFilter, TlvEventPath, TlvEventReport,
+    TlvEventStatus
+} from "./InteractionProtocol.js";
 import { tryCatchAsync } from "../../common/TryCatchHandler.js";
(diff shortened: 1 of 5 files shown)


packages/matter.js/src/protocol/interaction/SubscriptionHandler.ts
```diff
@@ -112,38 +115,61 @@ export class SubscriptionHandler {
...
-                    if (existingListener !== undefined) {
-                        existingAttribute.removeValueChangeListener(existingListener);
+        if (this.attributeRequests !== undefined) {
+            this.attributeRequests.forEach(path => {
+                const attributes = this.endpointStructure.getAttributes([path]);
+
+                if (attributes.length === 0) {
+                    const { endpointId, clusterId, attributeId } = path;
+                    if (endpointId === undefined || clusterId === undefined || attributeId === undefined) { // Wildcard path: Just leave out values
+                        logger.debug(`Subscription attribute ${this.endpointStructure.resolveAttributeName(path)}: ignore non-existing attribute`);
```
lauckhart, line 129: Is this the correct error message?

lauckhart (approved):

packages/matter.js/src/protocol/interaction/SubscriptionHandler.ts
Apollon77: In fcat it is no error message it is just logging. When a wildcard path was given and that one did not matched to any attribute then this is no error according to specs,but ignored.

> if the path is NOT a concrete path ...  perform Request Path Expansion and process each expanded existent path as follows:
> * If the path indicates attribute data that is not readable, then the path SHALL be dis­ carded.
> * Else if reading from the attribute in the path requires a privilege that is not granted to access the cluster in the path, then the path SHALL be discarded.

... nothing more ...

> • If no error free existent paths remain, then AttributeRequests are considered empty.

PR #266, opened 2023-08-09
Move ClusterClient/Server into own files

Apollon77:
... and also create an AttribueServer factory method. Commits contain single steps

* addresses https://github.com/orgs/project-chip/projects/11/views/1?pane=issue&itemId=35394270
* addresses https://github.com/orgs/project-chip/projects/11/views/1?pane=issue&itemId=35394314


lauckhart (approved):
I didn't look closely at the code given it's just shifting around but 👍 to the structure.

PR #267, opened 2023-08-09
Change to Prettier as Formatter

Apollon77:
For a more standard based and convenient integrated automatic formatting (also supported by many IDEs "on save") I propose (again) switching to prettier.
I excluded all auto generated files.

This PR only contains the needed changes to config and package. PR https://github.com/Apollon77/matter.js/pull/7 based on this showcases the changes this would mean (and will update this PR with this directly before merge to avoid too many conflicts.

Note: The testing here fails right now by intent because the formatting is not applied and so the format check step blocks. So please still review. Will be green after we agreed on it and I apply the formatting

addressing https://github.com/orgs/project-chip/projects/11/views/1?filterQuery=prett&pane=issue&itemId=35396891


lauckhart (approved):
It's not perfect, I'm personally not a fan of the extra commas and it'd be nice to have a little more editorial control over how fields wrap in type parameters and objects...  But HUGE improvement I say go for it

Apollon77:
I get opinions from the others too. Thank you!

Apollon77:
@lauckhart 
>  I'm personally not a fan of the extra commas 

I was the same ... but in fact using them makes diffs in changes more simple because it only adds a new line and not also needs to adjust the line before. ergo: less conflicts :-)

PR #270, opened 2023-08-11
Upgrade Basic Info to revision 2

Apollon77:
I we use some fields from chip XMLs already because of reasons we should also upgrade the cluster revision accordingly.


Apollon77:
updated using override

lauckhart (approved):

PR #278, opened 2023-08-14
Allow to configure loglevel and logformat for examples

Apollon77:
The PR changes the default log format for shell usage to ANSI and adds parameters to all examples to allow configuration of loglevel and format.
Additionally I adjusted ANSI colors a bit to be darkgrey and so better readable in white shells.

* finishs https://github.com/orgs/project-chip/projects/11/views/1?pane=issue&itemId=34444331

### DIFF
--- a/packages/matter.js/src/log/Logger.ts
+++ b/packages/matter.js/src/log/Logger.ts
@@ -86,15 +86,15 @@ function plainLogFormatter(now: Date, level: Level, facility: string, values: an
 }
 
 const ANSI_CODES = {
-    PREFIX: "\x1b[90m\x1b[2m",
-    FACILITY: "\x1b[90m\x1b[1m",
-    LEVEL_DEBUG: "\x1b[90m",
-    LEVEL_INFO: "\x1b[32m",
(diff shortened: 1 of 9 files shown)


packages/matter.js/src/log/Logger.ts
```diff
@@ -86,9 +86,9 @@ function plainLogFormatter(now: Date, level: Level, facility: string, values: an
 }
 
 const ANSI_CODES = {
-    PREFIX: "\x1b[37m\x1b[2m",
-    FACILITY: "\x1b[37m\x1b[1m",
-    LEVEL_DEBUG: "\x1b[37m",
+    PREFIX: "\x1b[90m\x1b[2m",
+    FACILITY: "\x1b[90m\x1b[1m",
+    LEVEL_DEBUG: "\x1b[90m",
```
turon, line 91: May be good to add comment of what expected color is.

turon (approved):

packages/matter.js/src/log/Logger.ts
Apollon77: Good idea, added

PR #285, opened 2023-08-21
Crypto tweaks

### LINKED ISSUE #276: Crypto sign error
```
Crypto › sign & verify with raw keys › signs data with generated private key

    Cannot infer named curve from key length

      442 |
      443 |             default:
    > 444 |                 throw new KeyError("Cannot infer named curve from key length");
          |                       ^
      445 |         }
      446 |     }
      447 | }

      at inferCurve (../matter.js/src/crypto/Key.ts:444:23)
      at Key (../matter.js/src/crypto/Key.ts:540:13)
      at PrivateKey (../matter.js/src/crypto/Key.ts:564:12)
      at Object.<anonymous> (test/crypto/CryptoTest.ts:67:54)
```

seen in https://github.com/project-chip/matter.js/pull/275/checks#step:7:17602

lauckhart:
Removed test that generates EC key pair using node's crypto API directly. Due to test implementation it used a code path that attempted to infer the elliptic curve from the private key size but Node doesn't include significant bytes that are 0.  This would cause key size inference to fail.

We test with a dynamic key using our "createKeyPair" method so above test was redundant.

Also added code to pad the Node key when we generate our own key internally. This isn't absolutely necessary because we always provide the public key point rather than computing from the private key.  The public key is encoded in a format with fixed length coordinates so curve inference works correctly.

But the truncated private key is a bit weird and could lead to future confusion.  We now pad to full key length just to be safe.

Fixes #276

### DIFF
--- a/packages/matter-node.js/test/crypto/CryptoTest.ts
+++ b/packages/matter-node.js/test/crypto/CryptoTest.ts
@@ -61,13 +61,7 @@ describe("Crypto", () => {
             cryptoNode.verify(PublicKey(PUBLIC_KEY), ENCRYPTED_DATA, result);
         });
 
-        it("signs data with generated private key", () => {
-            const ecdh = crypto.createECDH("prime256v1");
-            ecdh.generateKeys();
-            const result = cryptoNode.sign(PrivateKey(ecdh.getPrivateKey()), ENCRYPTED_DATA);
-
-            cryptoNode.verify(PublicKey(ecdh.getPublicKey()), ENCRYPTED_DATA, result);
-        });
+        // See createKeyPair test for further tests with private key
     });
 
     describe("sign & verify with SEC1 private and SPKI public keys", () => {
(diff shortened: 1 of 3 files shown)


packages/matter-node.js/test/crypto/CryptoTest.ts
```diff
@@ -91,5 +85,18 @@ describe("Crypto", () => {
 
             cryptoNode.verify(key, ENCRYPTED_DATA, cryptoNode.sign(key, ENCRYPTED_DATA));
         });
+
+        // Too slow to leave enabled by default but useful for confirming key variance doesn't cause failure
+        // it("creates a large volume of private keys", () => {
+        //     for (let i = 0; i < 100000; i++) {
```
Apollon77, line 91: maybe just lower the iterations to 10000 or such?

Apollon77 (approved):

packages/matter-node.js/test/crypto/CryptoTest.ts
lauckhart: Still too slow for my taste (7 seconds at 10k iterations)...  1k would probably be OK but isn't much of a test

PR #292, opened 2023-08-22
Additional access control metadata

### LINKED ISSUE #290: Generate "Timed" Access-flag into clusters
Hey @lauckhart,

for the upcoming finalizations of the protocol I could need support on the generation side of things :-)

I would need to have the "Timed" access flkag to be generated in the relevant attributes and commands. I propose doing that like scene/persistent and such flags.

Thank you

lauckhart:
Adds ACLs for commands and events and "timed" flag to attributes and commands.

Closes #290


Apollon77 (approved):
Great, thank you!

PR #304, opened 2023-09-03
Small memory Footprint optimizations

Apollon77:
... both mainly relevant for Logging

### DIFF
--- a/packages/matter.js/src/log/Logger.ts
+++ b/packages/matter.js/src/log/Logger.ts
@@ -189,14 +189,7 @@ export class DiagnosticDictionary {
      *
      * @param entries the entries as [ "KEY", value ] tuples
      */
-    constructor(private readonly entries: { [KEY: string]: any } = {}) {
-        // Use getOwnPropertyNames because it follows insertion order as of ES6
-        for (const KEY of Object.getOwnPropertyNames(entries)) {
-            if (entries[KEY] === undefined) {
-                delete entries[KEY];
-            }
-        }
-    }
+    constructor(private readonly entries: { [KEY: string]: any } = {}) {}
 
     /**
      * Format the dictionary for human consumption.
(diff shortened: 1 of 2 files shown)


packages/matter.js/src/util/ByteArray.ts
```diff
@@ -31,7 +31,9 @@ declare global {
 }
 
 Uint8Array.prototype.toHex = function () {
-    return this.reduce((result, byte) => result + byte.toString(16).padStart(2, "0"), "");
+    const hexArray = new Array<string>();
+    this.forEach(byte => hexArray.push(byte.toString(16).padStart(2, "0")));
```
lauckhart, line 35: Curious as to perf implications, this looks like the opposite of changes to DiagnosticDictionary

packages/matter.js/src/log/Logger.ts
```diff
@@ -184,33 +184,20 @@ const defaultValueFormatter = (value: string) => value;
...
-    constructor(entries: { [KEY: string]: any } = {}) {
-        // Use getOwnPropertyNames because it follows insertion order as of
-        // ES6
+    constructor(private readonly entries: { [KEY: string]: any } = {}) {
+        // Use getOwnPropertyNames because it follows insertion order as of ES6
         for (const KEY of Object.getOwnPropertyNames(entries)) {
-            if (entries[KEY] !== undefined) {
-                this.set(KEY, entries[KEY]);
+            if (entries[KEY] === undefined) {
+                delete entries[KEY];
```
lauckhart, line 196: This is destructive to input object, maybe could just omit `undefined` in serializion?

lauckhart (approved):

packages/matter.js/src/util/ByteArray.ts
Apollon77: before it used "reduce" which builded a new string with each added hex-value, so for a 10 byte Buffer it created string with lengths 2, 4, 6, 8, 10, 12, 14, 16, 18, 20 -> 9 strngs that are just there a microsecond, summing up 72+20 (final string) chars (I think internally it is UTF16, so 2 bytes per char).

new approach is duplicating data once and created an array with 10x2 chars and then joins to 1 string ... so it is 20+20. Ok I have no idea about the overhead of an array, but should not be that much.

makes sense?

lauckhart: Sorry, that makes sense, my comment wasn't very clear.  I was specifically wondering about using Array.add() here vs Array.map() like you did in DiagnosticDictionary.serialize().  NBD either way I was just curious.

Apollon77: Map do not work because map on uint8array expects a number as result of map and do not allow string ;-)

packages/matter.js/src/log/Logger.ts
Apollon77: Good idea. Will do

packages/matter.js/src/util/ByteArray.ts
lauckhart: Cool, you might preinitialize `hexArray`'s length to `this.length * 2`, that would be another small optimization.

PR #310, opened 2023-09-08
Excessive housekeeping

lauckhart:
This knocks a bunch of housekeeping tasks off the TODO list.  A large number of changes due to interrelated nature but the commits are fairly organized and include detailed changelogs.  Expect problems due to the number of changes.

A few key items:

- New build & test that streamlines configuration and is way faster
- Tests dist code including ESM (node and browser) and CommonJS (node)
- Replaces Jest/Jasmine/Karma (see changelogs for reasoning)
- Upgrade to TS5 and typescript-eslint 6


package.json
```diff
@@ -28,16 +29,19 @@
...
-        "typedoc": "^0.24.8",
+        "typedoc": "^0.25.1",
         "typedoc-plugin-markdown": "^3.15.4",
-        "typedoc-plugin-missing-exports": "^2.0.1",
-        "typescript": "^4.9.5"
+        "typedoc-plugin-missing-exports": "^2.1.0",
+        "typescript": "^5.2.2"
+    },
+    "dependencies": {
+        "v8-profiler-next": "^1.9.0"
```
Apollon77, line 45: as base dependency?

packages/matter-node.js/package.json
```diff
@@ -58,6 +58,7 @@
     },
     "files": [
         "dist/**/*",
+        "src/**/*",
```
Apollon77, line 61: should we publish the ts files realkly in the npm package?

packages/matter-node.js/test/cluster/GeneralCommissioningServerTest.ts
```diff
@@ -28,8 +28,6 @@ import { SecureSession } from "@project-chip/matter.js/session";
 import * as assert from "assert";
 import { callCommandOnClusterServer, createTestSessionWithFabric } from "./ClusterServerTestingUtil";
 
-/* eslint-disable @typescript-eslint/no-non-null-assertion */
```
Apollon77, line 31: this can be removed because?

packages/matter.js/package-cjs.json
Apollon77, line 1: Why you removed this? Needed for proper "magic" when publishing package

packages/matter.js/package.json
```diff
@@ -24,45 +24,41 @@
...
-    "test": "npm run test-node && npm run test-web"
+    "test": "matter-test -w",
+    "coverage": "c8 matter-test esm"
   },
   "dependencies": {
     "bn.js": "^5.2.1",
     "elliptic": "^6.5.4"
   },
   "devDependencies": {
+    "@project-chip/matter.js-tools": "0.1.0",
```
Apollon77, line 39: but tool ´s package is privte and will never be on npm ... so that might get issues if someone uses npm i after instalation or such other cases where devDeps are also installed?

packages/matter.js/src/device/Endpoint.ts
```diff
@@ -141,7 +141,7 @@ export class Endpoint {
         if (cluster.id === DescriptorCluster.id) {
             this.descriptorCluster = cluster as unknown as ClusterServerObjForCluster<typeof DescriptorCluster>;
         }
-        this.clusterServers.set(cluster.id, cluster);
+        this.clusterServers.set(cluster.id, cluster as any);
```
Apollon77, line 144: why? do we know ahat changed in TS to need that now? Whats the exact error?

packages/matter.js/src/device/Endpoint.ts
```diff
@@ -165,7 +165,7 @@ export class Endpoint {
     >(cluster: Cluster<F, SF, A, C, E>): ClusterServerObj<A, E> | undefined {
         const clusterServer = this.clusterServers.get(cluster.id);
         if (clusterServer !== undefined) {
-            return clusterServer as ClusterServerObj<A, E>;
+            return clusterServer as unknown as ClusterServerObj<A, E>;
```
Apollon77, line 168: same here

packages/matter.js/package-cjs.json
Apollon77: ok found it... is replaced and written dynamically

package.json
lauckhart: Yeah, it's the same as the other dependencies (mocha etc.), it's used by the code published by the package...  `npm run test -- --profile` will dump a .cpuinfo file into build/profiles

packages/matter-node.js/package.json
lauckhart: I lean that way, yeah, WDYT?  My thinking:

Pluses:  I've found transpiled files in NPMs without the source to be obnoxious.  Stack traces just take you to transpiled code and you have to check out repo to see what's really going on.  Also, it gives users the option to do their own transpiling if they want it transpiled differently.

Negatives:  Makes the package larger.  But next to the size of 2x source maps, type declarations and transpiled files I think this effect is pretty minimal.

packages/matter-node.js/test/cluster/GeneralCommissioningServerTest.ts
lauckhart: New eslint warned that it was unnecessary.  I'm not sure why it was there but I assume it was a lint bug that got fixed.

packages/matter.js/package.json
lauckhart: Not 100% sure but I believe it's OK because dev dependencies are only installed at build time.  The build would fail on the NPM package anyway because of missing files.

We could publish the tools package to be safe, that shouldn't hurt anything.

packages/matter.js/src/device/Endpoint.ts
lauckhart: Gives the "types do not have sufficient overlap" error.  Seems like it should be OK without the cast...  May be a TS5 bug but it happened on both 5.0 and 5.2.

I'll add comments to this effect, that sufficient?

packages/matter.js/src/device/Endpoint.ts
lauckhart: Yeah same issue

package.json
Apollon77: but mocha is a dev-Dependency and not an official one, but most likely ok fpr the base package to have ... ok

packages/matter-node.js/package.json
Apollon77: In fact we do not deliver the tools nor tsconfig and nothing else to "transpile their own" ... so the src is just there for "easier refernce" ... Is it really wortth that? The sourcemaps should still produce helpful messages to match code against the repo.
But if you really want tzo allow custom transpiling or "fix locally" we need to add even more ... and then produce maybe other "issues" bevause it is multiple packages ...

Need to think nd we could discuss in the group

packages/matter-node.js/test/cluster/GeneralCommissioningServerTest.ts
Apollon77: interesting ... ok ... In fact yes we do such null assertations in the code and did not have disabled the check ... soo I just wonder if we loose anything by other issues right now?

package.json
lauckhart: Ah, sorry...  I didn't look at which tsconfig.json this was.  It should have only been in the tools/tsconfig.json, will remove here.

packages/matter.js/package.json
Apollon77: we can leave out for now and add if needed ...

packages/matter.js/src/device/Endpoint.ts
Apollon77: I wonder why he accepts any because especially this shozld not overkllap with

> private readonly clusterServers = new Map<ClusterId, ClusterServerObj<Attributes, Events>>();

Apollon77: because in the case you say I would have expected an "as unknown as Bla"

lauckhart: `any` cast implicitly to any type.  I usually only use `as unknown as Type` when the expected type isn't already known, like below where there's a return without a return type annotation.

packages/matter-node.js/package.json
Apollon77: ok, lets inlcude them. all fine

packages/matter.js/src/device/Endpoint.ts
Apollon77: We can check later ... might really be a TS5 thing :-(

Apollon77 (approved):

packages/matter-node.js/test/cluster/GeneralCommissioningServerTest.ts
lauckhart: I added comments detailing the issue.  Resolving for now per conversation in Discord

packages/matter.js/src/device/Endpoint.ts
lauckhart: I updated the comment with more detail.  Resolving for now per Discord conversation

packages/matter-node.js/package.json
lauckhart: Added back

PR #318, opened 2023-09-21
New method for defining clusters

lauckhart:
Please see individual commits for details.  The following is from the main one:

The current Cluster type is somewhat cryptic/cumbersome due to the long list
of generic parameters.  Additionally, we lose type information for any fields
that aren't covered by the generic parameters.

The most immediate need for ongoing device work is access to the cluster
name in the cluster type.

Rather than add yet another generic parameter, this commit rewrites cluster
factory with alternate cluster types that preserve all information from the
cluster down to the element level.

The new types do not require a long list of generic parameters.  This works
using the new "const generic parameter" that's available now that we're on TS5.

We could conceivably migrate all existing references to Cluster<> to this new
approach but to keep things wieldly this commit only uses the new types for
defining clusters.  That unblocks the device API without inflicting changes
that are too massive.

Includes new generated clusters.  Also fixes a couple of places where the old
definitions seem to have been too permissive.

### DIFF
--- a/packages/matter.js/src/util/Type.ts
+++ b/packages/matter.js/src/util/Type.ts
@@ -56,5 +56,13 @@ export type MakeMandatory<T> = Exclude<T, undefined>;
 
 /** Create a branded type */
 declare const __brand: unique symbol;
-type Brand<B> = { [__brand]: B };
+// Don't think it should be necessary to export Brand<B> but it will cause
+// the following error under some circumstances:
+//
+//   Exported variable 'XXX' has or is using name '__brand' from external
+//   module "../src/util/Type" but cannot be named.ts(4023)
+//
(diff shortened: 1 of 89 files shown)


packages/matter.js/src/util/Type.ts
```diff
@@ -56,5 +56,13 @@ export type MakeMandatory<T> = Exclude<T, undefined>;
 
 /** Create a branded type */
 declare const __brand: unique symbol;
-type Brand<B> = { [__brand]: B };
+// DOn't think it should be necessary to export Brand<B> but it will cause
```
Apollon77, line 59:
```suggestion
// Don't think it should be necessary to export Brand<B> but it will cause
```

Apollon77 (approved):

PR #324, opened 2023-09-22
GHA: Install and Build also on macOS and test in LTS Node.js versions

Apollon77:
see subject :-)


Apollon77:
I removed the windows tests for now tikl checked ... somehow build tool has issues on windows

lauckhart (approved):

PR #329, opened 2023-09-23
Make sure Disable IPv4 disables it also for MDNS (and Windows GHA build optimization)

Apollon77:
This PR does:
* Breaking: Move "disableIpv4" config options from the MatterNode classes to the MatterServer and also consider it for MDNS init
* I decided to leave "disableIPv4" notation for the main API because we can leave it enabled by default because it is optionally supported.
* Breaking: Adjust MatterServer second parameter to an options object
* Respect disabled IPv4 in MDNS Broadcaster and do not send own IPv4 addresses
* Respect disabled IPv4 in MDNS Scanner and do not return IPv4 addresses back to the code when received
* Adjust MDNS class factories to option object style rather than too many optional parameters
* make ipv4 interface names and addresses optional as parameter in some places
* It includes changelog entries from former PRs

### DIFF
--- a/packages/matter.js/src/net/UdpMulticastServer.ts
+++ b/packages/matter.js/src/net/UdpMulticastServer.ts
@@ -78,7 +78,7 @@ export class UdpMulticastServer {
                         const broadcastTarget = iPv4 ? this.broadcastAddressIpv4 : this.broadcastAddressIpv6;
                         if (broadcastTarget == undefined) {
                             // IPv4 but disabled, so just resolve
-                            return Promise.resolve();
+                            return;
                         }
                         try {
                             await (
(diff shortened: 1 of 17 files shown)


packages/matter.js/src/net/UdpMulticastServer.ts
```diff
@@ -73,14 +75,15 @@ export class UdpMulticastServer {
                 await Promise.all(
                     ips.map(async ip => {
                         const iPv4 = isIPv4(ip);
+                        const broadcastTarget = iPv4 ? this.broadcastAddressIpv4 : this.broadcastAddressIpv6;
+                        if (broadcastTarget == undefined) {
+                            // IPv4 but disabled, so just resolve
+                            return Promise.resolve();
```
lauckhart, line 81: I think you can just return here, Promise.all() takes any thenable object so undefined should be OK

lauckhart (approved):

packages/matter.js/src/net/UdpMulticastServer.ts
Apollon77: It is in fact even easier ... it is an "asny function" ... so it will always return a promise :-)

PR #330, opened 2023-09-23
BLE optimizations

Apollon77:
* Inizialize Bleno like Noble and move the HCI handling
* Try to reset device on noble start (ignore failures)

### DIFF
--- a/packages/matter-node-ble.js/src/ble/BlenoBleServer.ts
+++ b/packages/matter-node-ble.js/src/ble/BlenoBleServer.ts
@@ -28,12 +28,7 @@ function initializeBleno(server: BlenoBleServer, hciId?: number) {
     if (hciId !== undefined) {
         process.env.BLENO_HCI_DEVICE_ID = hciId.toString();
     }
-    try {
-        Bleno = require("@abandonware/bleno");
-    } catch (error: any) {
-        logger.error(`Error loading Bleno: ${error.message}`);
-        throw error;
-    }
+    Bleno = require("@abandonware/bleno");
 
     class BtpWriteCharacteristicC1 extends Bleno.Characteristic {
         constructor() {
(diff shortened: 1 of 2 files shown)


packages/matter-node-ble.js/src/ble/BlenoBleServer.ts
```diff
@@ -23,8 +23,17 @@ import { BleOptions } from "./BleNode";
...
-    Bleno = require("@abandonware/bleno");
+function initializeBleno(server: BlenoBleServer, hciId?: number) {
+    // load Bleno driver with the correct device selected
+    if (hciId !== undefined) {
+        process.env.BLENO_HCI_DEVICE_ID = hciId.toString();
+    }
+    try {
+        Bleno = require("@abandonware/bleno");
+    } catch (error: any) {
+        logger.error(`Error loading Bleno: ${error.message}`);
```
lauckhart, line 34: This will end up reporting the error twice...  Maybe you could do:

```js
    error.message = `Error loading BLE peripheral module: ${error.message}`;
    error.stack = `Error loading BLE peripheral module: ${error.message}`;
    throw error;
```

Unfortunately you have to handle it twice because of the stupid way errors work.  Could make a utility function for that.  Or another option:

```js
    logger.error("Error loading Bleno:", error);
    throw new Error("BLE peripheral module did not load");
```

Maybe at some point we could augment MatterError with a "cause" field so we can properly chain exceptions.

lauckhart (approved):

packages/matter-node-ble.js/src/ble/BlenoBleServer.ts
Apollon77: Right ... and in upper places we have no try/catch so we crah ... so I would remove the logging here completely ...  because makes no sense

PR #331, opened 2023-09-23
Add Commissioning Failsafe Request Timeout Handling (Client)

Apollon77:
This PR adds the correct handling of extended response timeouts during an armed failsafe timer. In this case a request can take up to 30s to respond which means that the normal "4 retransmissions" timeframe (maximum 4.2s in total) needs to be extended. I realized this by adding another parameter to define a minimum response timeout. because of the fact that the specification do not fully describe how to realize this extended waiting time I decided to simply continue to send retransmissions with default increasing backoff time and check the timout once such a retransmission triggers. The Commissioner class knows when failsafe is enabled and so manually uses this flag whenever it makes sense.

I did not add an integration test for this, but manually tested it via a Meross EU device which needs 6+ seconds to connect to a Wifi network and so the normal 4 times retransmissions was always failing. Should I add a test or do we accept that for now? I could also add a TODO for adding test later

### DIFF
--- a/packages/matter.js/src/cluster/client/ClusterClient.ts
+++ b/packages/matter.js/src/cluster/client/ClusterClient.ts
@@ -284,9 +284,13 @@ export function ClusterClient<F extends BitSchema, A extends Attributes, C exten
         commandToId[requestId] = commandName;
         commands[commandName] = async <RequestT, ResponseT>(
             request: RequestT,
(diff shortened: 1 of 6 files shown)


lauckhart (approved):
TODO for a test is OK by me

Apollon77:
Added https://github.com/orgs/project-chip/projects/11/views/1?pane=issue&itemId=39585537 for general MessageExchange tests

PR #338, opened 2023-09-23
Do not decline higher Interaction Model revisions, just log

Apollon77:
Declining the execution with higher Interaction model revisions was a bit hard ... so just log that we see it. In fact not "Bad" because the  controller needs to take care in this case

### DIFF
--- a/packages/matter.js/src/protocol/interaction/InteractionServer.ts
+++ b/packages/matter.js/src/protocol/interaction/InteractionServer.ts
@@ -193,9 +193,8 @@ export class InteractionServer implements ProtocolHandler<MatterDevice> {
         );
 
         if (interactionModelRevision > INTERACTION_MODEL_REVISION) {
-            throw new StatusResponseError(
-                `Interaction model revision ${interactionModelRevision} not supported`,
-                StatusCode.InvalidAction,
+            logger.debug(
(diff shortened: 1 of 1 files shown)


packages/matter.js/src/protocol/interaction/InteractionServer.ts
```diff
@@ -193,9 +193,8 @@ export class InteractionServer implements ProtocolHandler<MatterDevice> {
         );
 
         if (interactionModelRevision > INTERACTION_MODEL_REVISION) {
-            throw new StatusResponseError(
-                `Interaction model revision ${interactionModelRevision} not supported`,
-                StatusCode.InvalidAction,
+            logger.debug(
+                `Interaction model revision of sender ${interactionModelRevision} is higher than supported ${INTERACTION_MODEL_REVISION}.`,
```
lauckhart, line 197: Lots of replication here, maybe a helper function?

lauckhart (approved):

packages/matter.js/src/protocol/interaction/InteractionServer.ts
Apollon77: I also thought, but decided for now to leave that way until we have first specs infos and aybe the first "revision specific logics" and see how we need to hande it really

PR #339, opened 2023-09-23
Commissioner optimizations

Apollon77:
Keep track of the network commissioning success to prevent unneeded tries. Additionally, try to reset the failsafe timer in case of errors so that the user do not need to wait for the timer to expire.

### DIFF
--- a/packages/matter.js/src/protocol/ControllerCommissioner.ts
+++ b/packages/matter.js/src/protocol/ControllerCommissioner.ts
@@ -108,6 +108,7 @@ type CollectedCommissioningData = {
     vendorId?: VendorId;
     productId?: number;
     supportsConcurrentConnection?: boolean;
+    successfullyConnectedToNetwork?: boolean;
 };
 
 /** Error that throws when Commissioning fails and process can not be continued. */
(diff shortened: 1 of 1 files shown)


packages/matter.js/src/protocol/ControllerCommissioner.ts
```diff
@@ -116,7 +117,7 @@ export class CommissioningError extends MatterError {}
 /** Error that throws when Commissioning fails but process can be continued. */
 class RecoverableCommissioningError extends CommissioningError {}
 
-const DEFAULT_FAILSAFE_TIME_MS = 60000;
+const DEFAULT_FAILSAFE_TIME_MS = 60_000; // 60 seconds
```
lauckhart, line 120: Maybe we could rename 60_000 to ONE_MINUTE_MS

lauckhart (approved):

packages/matter.js/src/protocol/ControllerCommissioner.ts
Apollon77: :-)) downside is that you then require other files again ... or you have it duplicated ... Would leave for now :-)

PR #346, opened 2023-09-24
Refactor DataReport logic and add dynamic array chunking

Apollon77:
... and also implement tag compression, but keep disabled because not yet supported.

This PR does:
* Enhance existing DataDecoding Logic
* Make dataVersion in types optional and handle tag-compressed dataVersions (restore them)
* Add DataEncoder to allow dynamic chunking and tag-compression. This also introduces new intermediate types for data formats used internally to allow dynamic compression and chunking
* Enhance AnySchema to allow the calculation of the bytesize of a TlvStream (used to encode data)
* Use the new encoder for reads and subscriptions
* Refactor the DataReport sending logic

Beside the added tests and many manual tests with Apple, Google and Tuya (and amazon) the Integratiintest was unachanged, so proofed via this that we ourself at löeast can decode what we send.

### DIFF
--- a/packages/matter.js/src/protocol/interaction/AttributeDataEncoder.ts
+++ b/packages/matter.js/src/protocol/interaction/AttributeDataEncoder.ts
@@ -3,10 +3,12 @@
  * Copyright 2022-2023 Project CHIP Authors
  * SPDX-License-Identifier: Apache-2.0
  */
+import { MatterFlowError } from "../../common/MatterError.js";
 import { AttributeId } from "../../datatype/AttributeId.js";
 import { ClusterId } from "../../datatype/ClusterId.js";
 import { EndpointNumber } from "../../datatype/EndpointNumber.js";
 import { NodeId } from "../../datatype/NodeId.js";
+import { Logger } from "../../log/Logger.js";
 import { ArraySchema } from "../../tlv/TlvArray.js";
 import { TlvSchema, TlvStream, TypeFromSchema } from "../../tlv/TlvSchema.js";
 import {
@@ -115,13 +117,17 @@ export function canAttributePayloadBeChunked(attributePayload: AttributeReportPa
 export function chunkAttributePayload(attributePayload: AttributeReportPayload): AttributeReportPayload[] {
     const { attributeData } = attributePayload;
     if (attributeData === undefined) {
-        // Should never happen, but we can make sure it do not break
-        return [attributePayload];
+        throw new MatterFlowError(
+            `Can not chunk an AttributePayload with just a attributeStatus: ${Logger.toJSON(attributePayload)}`,
+        );
     }
     const { schema, path, dataVersion, payload } = attributeData;
     if (!(schema instanceof ArraySchema) || !Array.isArray(payload)) {
-        // Should never happen, but we can make sure it do not break
-        return [attributePayload];
+        throw new MatterFlowError(
+            `Can not chunk an AttributePayload with attributeData that is not an array: ${Logger.toJSON(
+                attributePayload,
+            )}`,
+        );
     }
     const chunks = new Array<AttributeReportPayload>();
     chunks.push({ attributeData: { schema, path: { ...path, listIndex: undefined }, payload: [], dataVersion } });
--- a/packages/matter.js/src/protocol/interaction/AttributeDataEncoder.ts
+++ b/packages/matter.js/src/protocol/interaction/AttributeDataEncoder.ts
@@ -1,0 +1,249 @@
+/**
+ * @license
+ * Copyright 2022-2023 Project CHIP Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+import { AttributeId } from "../../datatype/AttributeId.js";
+import { ClusterId } from "../../datatype/ClusterId.js";
+import { EndpointNumber } from "../../datatype/EndpointNumber.js";
+import { NodeId } from "../../datatype/NodeId.js";
+import { ArraySchema } from "../../tlv/TlvArray.js";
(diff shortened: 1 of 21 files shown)


packages/matter.js/src/protocol/interaction/AttributeDataEncoder.ts
```diff
@@ -0,0 +1,249 @@
...
+    } = attributeData;
+    return schema instanceof ArraySchema && Array.isArray(payload) && payload.length > 0 && listIndex === undefined;
+}
+
+/** Chunk an AttributeReportPayload into multiple AttributeReportPayloads. */
+export function chunkAttributePayload(attributePayload: AttributeReportPayload): AttributeReportPayload[] {
+    const { attributeData } = attributePayload;
+    if (attributeData === undefined) {
+        // Should never happen, but we can make sure it do not break
+        return [attributePayload];
```
lauckhart, line 119: I'm wondering if it wouldn't be better to break with an error message.  Probably more philosophical than anything else

packages/matter.js/src/protocol/interaction/AttributeDataEncoder.ts
```diff
@@ -0,0 +1,249 @@
...
+    }
+    const { schema, path, dataVersion, payload } = attributeData;
+    if (!(schema instanceof ArraySchema) || !Array.isArray(payload)) {
+        // Should never happen, but we can make sure it do not break
+        return [attributePayload];
+    }
+    const chunks = new Array<AttributeReportPayload>();
+    chunks.push({ attributeData: { schema, path: { ...path, listIndex: undefined }, payload: [], dataVersion } });
+    payload.forEach(element => {
+        chunks.push({
```
lauckhart, line 129: If I understand the spec correctly every chunk will require a network round trip before the next one goes.  I think you said CHIP does it that way but oof seems like a lot of latency.  Is it too difficult to integrate w/ serialization so we can pack messages optimally into the assumed 1280 MTU?  Maybe at least a comment

packages/matter.js/src/protocol/interaction/AttributeDataEncoder.ts
```diff
@@ -0,0 +1,249 @@
...
+                payload: element,
+                dataVersion,
+            },
+        });
+    });
+
+    // If the path of the initial AttributePayload was not compressed, we should compress the chunked data
+    // TODO Enable once Tag Compression is supported https://github.com/project-chip/connectedhomeip/issues/29359
+    /*
+    if (path.enableTagCompression !== true) {
```
lauckhart, line 142: Given that you put the time in to implement maybe turning it on could be an option?  Then there'd be a reference implementation ;)

packages/matter.js/src/protocol/interaction/InteractionMessenger.ts
```diff
@@ -246,26 +275,18 @@ export class InteractionServerMessenger extends InteractionMessenger<MatterDevic
                         dataReport.moreChunkedMessages = undefined;
                         break;
                     }
-                    const eventReportBytes = TlvAttributeReport.encode(eventReport).length;
+                    if (!firstEventAddedToReportMessage) {
+                        firstEventAddedToReportMessage = true;
+                        messageSize += 3; // Array element with is added now with length needs 3 bytes
```
lauckhart, line 280: Array element "width" I think?

packages/matter.js/src/tlv/TlvCodec.ts
```diff
@@ -339,6 +339,34 @@ export class TlvCodec {
         }
     }
 
+    public static getTagByteLength(tag?: TlvTag) {
```
lauckhart, line 342: I can see the byte length stuff is implemented and tested but I can't find where it's used, is that a future commit or did I miss it?

lauckhart (approved):

packages/matter.js/src/protocol/interaction/AttributeDataEncoder.ts
Apollon77: You are right. I will change. Then we see such cases.

packages/matter.js/src/protocol/interaction/AttributeDataEncoder.ts
Apollon77: Yes it requires to be answered with "ok". The new logic is at least packaging it better then before. Packaging more would require a lot of logic to decide what's best. Especially when the sizes differ because of array chunking and tag compression and then additional requirements apply. So not sure if "one less roundtrip" is worth that computing effort. But yes could be an further optimization for the future.

packages/matter.js/src/protocol/interaction/AttributeDataEncoder.ts
Apollon77: We can not turn it on because else no one will accept the payload. In fact it will also be a "by interaction model revision" decision later once it is officially supported. So later we need kind of a "add compression when interaction model revision >=X"

packages/matter.js/src/protocol/interaction/InteractionMessenger.ts
Apollon77: LOL. No. Word too much ;-)

packages/matter.js/src/tlv/TlvCodec.ts
Apollon77: It is used in the method that puts  together the datareport and where the size is calculated.

Apollon77: PS: aah ok this one is used in AnySchema ... and the one from AnySchema ist used by encoding logic

PR #353, opened 2023-09-27
Add more chip tests

Apollon77:
This PR adds all tests possible from https://github.com/orgs/project-chip/projects/11/views/1?pane=issue&itemId=34439142

One Commissioning issue was also fixed here


lauckhart (approved):
Looks like that timeout's too tight but I'm headed out for a wedding tomorrow so approving in case I don't get to it later.

lauckhart (approved):
Looks good. Now have to place device in airplane mode 🙂

PR #358, opened 2023-09-29
Add missing delayedAnnouncement assignment

### LINKED ISSUE #357: InternalError: onData listener already set
Run `npm run matter-device -- -ble` and it throws.

It seems that both `this.matterServer.start()` and `commissioningServer.advertise({ble: true})` tried to set the listener.

vilic:
Addresses #357


Apollon77:
Thank you, will merge. SO you for the lucky path your PR is right, but not long term, because cannling start()/advertise() twice shpuld be allowed or with a defined error not allowed :-) So I would keep the issue open

Apollon77 (approved):

PR #360, opened 2023-10-01
Use scope id instead of name for multicast interface on Windows

### LINKED ISSUE #359: Failed to add example device on Windows
I am trying to pair a `matter-node.js` example device on Windows 11 with Apple Home, but overall, it fails predicably.

Output pairing via Bluetooth (5.3 adapter with WinUSB driver): https://gist.github.com/vilic/5d90880764d53d5eea6d4bbb989d540f

Pairing without Bluetooth is even harder, and I haven't got one single complete success. Most of the time it fails to connect at the beginning.

It turns out that interface name in IPv6 address gets ignored and falls back to seemingly unpredictable interface if there's multiple ones. In my case, I have `Wi-Fi` and `vEthernet (Default Switch)`, and `::%<name>` always falls back to the second one, and making the pairing unsuccessful.

vilic:
On Windows, it seems that `::%<name>` always falls back to some interface ignoring the name given. To make it work with multiple network interfaces, scope id might be required.

Fixes #359

### DIFF
--- a/packages/matter-node.js/src/net/NetworkNode.ts
+++ b/packages/matter-node.js/src/net/NetworkNode.ts
@@ -12,9 +12,9 @@ import { UdpChannelNode } from "./UdpChannelNode.js";
 
 export class NetworkNode extends Network {
     static getMulticastInterface(netInterface: string, ipv4: boolean) {
+        const netInterfaceInfo = networkInterfaces()[netInterface];
+        if (netInterfaceInfo === undefined) throw new NetworkError(`Unknown interface: ${netInterface}`);
         if (ipv4) {
-            const netInterfaceInfo = networkInterfaces()[netInterface];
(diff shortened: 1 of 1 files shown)


Apollon77:
Thank you, did you found any sources for this behaviour on Google or such? Can you link them? (WOukld like t understand this windows behaviour)

vilic:
@Apollon77 
- https://superuser.com/questions/99746/why-is-there-a-percent-sign-in-the-ipv6-address (find "Windows" to highlight relevant content)
- https://learn.microsoft.com/en-us/windows/win32/wininet/ip-version-6-support#scope-id

Apollon77:
Thank you, will check later today and merge then too I assume :-)

Apollon77:
Found some time to read and yes windows is doing stuff differently. ;-))

Apollon77 (approved):

PR #363, opened 2023-10-02
MDNS Optimizations

Apollon77:
This PR does:
* Add missing values from the messages and their parsing. Additionally optimize some things and improve typing. Also introduce a "Partially Pre Encoded" type for DnsMessage to allow pre encoding.
* Also correctly set the real MDNS message size by removing all UDP and IP headers.
* Enhance DataReader to set the offset 
* Optimize MdnsScanner query sending by pre-encoding records
* Makes sure to also respect maximum MDNS message size when the Scanner sends Responses which was not the case before. Additionally it fulfills some more requirements from MDNS specs, e.g.:
* send via Unicast is requested and allowed
* adds "Known answers suppression" (e.g. remove answers already provided in query as known from response)
* adds "duplicate record suppression" by preventing records being sent from last second
* optimize Message assembly and splitting, also implement pre-encoding


Apollon77:
PS: I tried to also run chip tests by that one commit, but did not worked :-( I run them locally without an issue.

packages/matter.js/src/mdns/MdnsServer.ts
```diff
@@ -77,28 +91,74 @@ export class MdnsServer {
...
+                    (_, index) => answersTimeSinceLastMultiCast[index].timeSinceLastMultiCast > 1000,
+                );
+                if (answers.length === 0) continue; // Nothing to send
+
+                answers.forEach(answer =>
+                    this.recordLastSentAsMulticastAnswer.set(this.buildDnsRecordKey(answer, netInterface), now),
+                );
+            }
+
+            this.sendRecords(
```
lauckhart, line 150: In general even if we want stuff to execute asynchronously we should probably be keeping track of the promises and doing an await during teardown, before repeating the operation, on cancel, etc.

Otherwise I worry hangs and/or other unexpected behavior could lead to resource leaks and generally difficult-to-diagnose issues.

packages/matter.js/src/net/UdpMulticastServer.ts
```diff
@@ -66,31 +66,43 @@ export class UdpMulticastServer {
...
-                        } catch (error) {
-                            logger.info(`${netInterface}: ${(error as Error).message}`);
-                        }
-                    }),
-                );
-            }),
-        );
+
+        // When we know the network interface and the unicast target, we can send unicast
+        if (uniCastTarget !== undefined && netInterface !== undefined) {
```
lauckhart, line 73: Why do we need the net interface? If we have an IP could we let the kernel's ARP/NDP tables handle routing?

Also curious about the case where there's a unistrut with no net interface... If that's illegal maybe we should throw an exception?  As written it'd end up doing a broadcast which is probably wrong

lauckhart (changes requested):
Good stuff, I did add q's about some of the async & error handling but core logic looks good from high level

packages/matter.js/src/MatterDevice.ts
```diff
@@ -281,11 +281,17 @@ export class MatterDevice {
             // TODO Change when refactoring MatterDevice away
             this.initialCommissioningCallback();
         }
+        this.sendFabricAnnouncements([fabric], true).catch(error =>
```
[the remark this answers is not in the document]
Apollon77: I also thought but in fact cancelling a Promise is not possible. so wouod be only the await case, but on shutdown this is also not too relevant - especially if the promise "hangs". In fact on shutdown e close all underlaying network connections, so the "catch" case should be called because exceptions would happen. Thats why I decided to just go that way. 

If you have a better idea, please tell.

packages/matter.js/src/mdns/MdnsServer.ts
```diff
@@ -77,28 +91,74 @@ export class MdnsServer {
...
+                {
+                    messageType: DnsMessageType.Response,
+                    transactionId,
+                    answers,
+                    additionalRecords,
+                },
+                netInterface,
+                uniCastResponse ? remoteIp : undefined,
+            ).catch(() => {
+                // ignore because already caught in UdpMulticastServer
```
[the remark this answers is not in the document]
Apollon77: The exceptioin is already handled on level "deeper", so this is a "will never happen" case, but in order to execute it in parallel without away TS required to have a catch  here :-) So it is a "conventions require it, but noop anyway"

packages/matter.js/src/mdns/MdnsServer.ts
Apollon77: see above. on shutdown we close all underlaying network sockets and such which should end all these gracefully anyway - additional to "I know no wway to cancel a Promise", beside what said ... closing/ending underlaying socket

packages/matter.js/src/net/UdpMulticastServer.ts
Apollon77: Depending on the user settings we open one socket per network interface the user wants us to use (ok he can limit to one, or we use "all"). With this we keep track forn an incoming message via which interface it came in and then alsoi send the response to this interface. This is especially relevant (at least from my understanding so far) for "link local ipv6" traffic where the address is only reachable via "the same" interface. nd so we use the interface to select the right outgoing socket.
it is correct that we alredy should augument the "ip" with this but I'm not fure if we do in all cases/places ...

In fact I decided to stay in this PR with the "we did that way in the past" way and do not also started this topic. We already discussed doing a refactoring here and I still agree that we should be able to get rid of the network interface info, but thats it's own and bigger topic.

packages/matter.js/src/mdns/MdnsServer.ts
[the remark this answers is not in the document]
lauckhart: What's "the exception" though?  Say DnsCode was undefined due to circular dependency, then line 191 would throw but wouldn't this catch() call swallow that error?  Just a specific example; seems like it'd swallow any unexpected error in this code path.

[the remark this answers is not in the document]
Apollon77: Ok you are right ... I added some more logic in there ... so just log the error?

packages/matter.js/src/mdns/MdnsServer.ts
lauckhart: I meant canceling the operation, not via the Promise interface.  Closing the underlying network sockets seems like the right thing to do.  Then the async networking code notes termination and returns, the promise is fulfilled and the termination code exits cleanly.

Ideally we'd then have a timeout promise in a Promise.any() that prints a loud "network logic failed to terminate" if the cleanup code didn't exit cleanly in the allotted time.

Point though is that we've got a handle on ongoing async operations so they can't hang without our noticing.

I could take on as a TODO but need to note the places in the code where we're letting an async operation "dangle".

packages/matter.js/src/net/UdpMulticastServer.ts
lauckhart: Ah right sorry didn't consider that this was part of previous conversation.

R.e. link-local, yeah, we'll need to have the "%zone-id" in the address.

packages/matter.js/src/mdns/MdnsServer.ts
Apollon77: As told in discord I would more think about a general MDNS sending  timeout thing on socket level. Roadmap todo added

lauckhart: Resolving per conversation in Discord

packages/matter.js/src/MatterDevice.ts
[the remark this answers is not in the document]
lauckhart: This is along the lines of below, resolving as well

lauckhart (approved):

PR #375, opened 2023-10-07
Call addMembership() for mdns server socket

### LINKED ISSUE #374: UdpMulticastServer not receiving messages
Trying out the controller example (`ControllerNode`), and found `MmdnsServer` (`UdpMulticastServer`) not receiving any message.

Did some test and found that `addMembership()` might be necessary (for multiple local-link interfaces, some router configuration, or Windows, IDK, as it seems to be required on Ubuntu in WSL2 on my computer as well)?

I did some dirty modification (by adding memberships on different interfaces to `"ff02::fb"` if udp6 listening on `5353`) and now the controller is able to pair successfully.

See also https://nodejs.org/api/dgram.html#socketaddmembershipmulticastaddress-multicastinterface

Some test script for convenience.

```js
import { createSocket } from "dgram";

const server = createSocket({ type: "udp6", reuseAddr: true });

server.bind(5353, () => {
    // In my case, the OS chooses `::%66` by default.
    // server.addMembership("ff02::fb", "::%66");
    server.addMembership("ff02::fb", "::%9");
});

server.on("message", (data, { address }) => {
    console.log(address, data);
});

const client = createSocket({ type: "udp6", reuseAddr: true });

client.bind(5353, () => {
    client.setMulticastInterface("::%9");

    setInterval(() => client.send("Hello", 5353, "ff02::fb"), 1000);
});
```

---

Okay, this issue is not IPv6 specific, `addMembership()` is also needed for UDP4.

vilic:
Fixes #374


Apollon77:
Thank you. Interesting. I really wonder why I never had any issue on raspi and macOS. I also had the membership thing in before but it never really made any difference.

But maybe this also could describe issues under windows that were reported. 

I will release the current main branch as 0.6 tomorrow. After this I will check out your changes on my systems.

vilic:
@Apollon77 I just come out with a theory and did some and I think that might be the cause.

On Ubuntu (WSL 2), and probably macOS and raspi, membership might be "shared" cross processes. While on Windows it's not.

I tried similar script, and adding membership in another process works on Ubuntu but not Windows. And as there are probably already other services setting up mDNS servers on macOS and raspi, so it's not a problem without adding membership again.

Apollon77:
Thank you. And yes these might be the reasons. I will test out code mainly to make sure that it "still works" for the systems where it worked before ;-)

Apollon77 (approved):

Apollon77:
@vilic There was a reportin discord about these erros in log

```
Error: addMembership ENODEV
    at Socket.addMembership (node:dgram:852:11)
    at UdpChannelNode.create (/etc/nixos/matter.js/packages/matter-node.js/dist/cjs/net/UdpChannelNode.js:100:18)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async UdpMulticastServer.create (/etc/nixos/matter.js/packages/matter.js/dist/cjs/net/UdpMulticastServer.js:61:50)
    at async MdnsServer.create (/etc/nixos/matter.js/packages/matter.js/dist/cjs/mdns/MdnsServer.js:73:7)
    at async MdnsBroadcaster.create (/etc/nixos/matter.js/packages/matter.js/dist/cjs/mdns/MdnsBroadcaster.js:58:7)
    at async MatterServer.start (/etc/nixos/matter.js/packages/matter.js/dist/cjs/MatterServer.js:91:30)
    at async main (/etc/nixos/index.js:105:5) {
  errno: -19,
  code: 'ENODEV',
  syscall: 'addMembership'
}
```

vilic:
@Apollon77 Okay I'll send a new PR soon. Possibly due to no network according to some: https://github.com/mafintosh/multicast-dns/pull/34

PR #380, opened 2023-10-08
Fix build-docs script source link generation

Apollon77:


packages/matter.js/tsconfig.json
```diff
@@ -25,7 +25,7 @@
       "src/log/export.ts",
       "src/math/export.ts",
       "src/mdns/export.ts",
-      "src/model/export.ts",
+      "src/model/import.ts",
```
lauckhart, line 28: Is this supposed to be index.ts?

Apollon77: yeeees :-) fixed

lauckhart (approved):

PR #385, opened 2023-10-11
Test commands throwing StatusResponseError

lauckhart:
This just adds a test case confirming we do the right thing when commands throw StatusResponseError.


Apollon77:
@lauckhart please format :-)
https://github.com/project-chip/matter.js/actions/runs/6488615306/job/17621381844?pr=385#step:4:10

Apollon77 (approved):

PR #389, opened 2023-10-13
Setup TypeScript project references

### LINKED ISSUE #376: Why moving away from TypeScript project references?
Having been playing with Matter.js for some days and the building process during debugging is really painful. And I noticed it was once configured with project references but was removed in [this commit](https://github.com/project-chip/matter.js/commit/6c7676fe6c60796595d5b566d809299524b4e84e) (at least for `matter-node.js` package).

What is the reason not using project references, especially considering better incremental building & editor experience?

vilic:
Closes #376

### DIFF
--- a/chip-testing/test/tsconfig.json
+++ b/chip-testing/test/tsconfig.json
@@ -1,0 +1,11 @@
+{
+    "extends": "../../tools/tsconfig.base.json",
+    "compilerOptions": {
+        "outDir": "../build/test",
+        "target": "es2018",
+        "module": "nodenext",
+        "moduleResolution": "nodenext",
+        "types": ["node", "mocha"]
+    },
+    "references": [{ "path": "../src" }]
+}
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -4,11 +4,11 @@
         "ecmaVersion": 2018, // Allows for the parsing of modern ECMAScript features
         "sourceType": "module", // Allows for the use of imports
         "project": [
-            "packages/*/tsconfig.json",
-            "tools/tsconfig.json",
-            "codegen/tsconfig.json",
-            "models/tsconfig.json",
-            "chip-testing/tsconfig.json"
+            "packages/*/{src,test}/tsconfig.json",
+            "tools/{src,test}/tsconfig.json",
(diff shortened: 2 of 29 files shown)


Apollon77:
@vilic seems you had an outdated base .. can you please resolve the conflicts?

Apollon77 (dismissed):
Please rebase and also cmake sure to take over the tscontidf changes (mainly typedoc) I did sdome days ago

vilic:
@Apollon77 Done. :D

chip-testing/test/tsconfig.json
```diff
@@ -0,0 +1,11 @@
+{
+    "extends": "../../tools/tsconfig.base.json",
+    "compilerOptions": {
+        "outDir": "../build/test",
+        "target": "es2018",
+        "module": "nodenext",
```
lauckhart, line 6: Am I correct in understanding module & moduleResolution are just here because you moved them from a level up?  Probably we should just remove and use the default node16 but doesn't have to be this PR.

chip-testing/test/tsconfig.json
```diff
@@ -0,0 +1,11 @@
+{
+    "extends": "../../tools/tsconfig.base.json",
+    "compilerOptions": {
+        "outDir": "../build/test",
+        "target": "es2018",
```
lauckhart, line 5: @Apollon77 the `target` tsconfig option proliferates with this PR.  I think we should just change to ES2018 in tsconfig.base.json.

My reasoning:

1. We use a number of ES2017 and ES2018 features that I'm pretty sure can't be backfilled (e.g. regexp "s" flag is ES2018)
2. More generally I'm 99% positive any hypothetical ES5 runtime has 0% chance of running matter.js
3. It would make transpiled code and a few of the new tsconfigs more compact

WDYT?

Doesn't need to be this PR (but could be 😉) just mentioning here because of the redundancy.

lauckhart (approved):
Thanks @vilic, this looks great

chip-testing/test/tsconfig.json
vilic: Yes, I just kept them from the original upper level `tsconfig.json`.

PR #394, opened 2023-10-13
Only run non-linux tests when packages change

Apollon77:


lauckhart (approved):
We could limit even more to only packages that affect tests but nbd

PR #396, opened 2023-10-14
Catch and log error adding membership

vilic:
Related to #375


Apollon77:
Thank you very much!

I asked someone in discord to verify.

Apollon77:
Ps: maybe just log a warn? Because as we know it still might work anything as expected ...

vilic:
@Apollon77 Done.

Apollon77 (approved):

PR #410, opened 2023-10-18
MatterServer enhancements

Apollon77:
This PR is changing the following:
* Ports for Commissioningservers are now handled automaticaly if wanted and generating port numbers starting 5540 and increasing. Own ports can still be provided and are used if not conflicting
* Add property "uniqueStorageKey" to the addCommissioningServer and addCommissioningController methods to allow to specify a unique storage key
* Allow to remove Server and Controllers from MatterServer instance, including with the option to kill the storage

### DIFF
--- a/packages/matter-node.js/src/storage/StorageBackendDisk.ts
+++ b/packages/matter-node.js/src/storage/StorageBackendDisk.ts
@@ -75,11 +75,10 @@ export class StorageBackendDisk implements Storage {
     }
 
     clearAll(contexts: string[]): void {
-        if (!contexts.length) throw new StorageError("Context must not be empty!");
         const contextKey = contexts.join(".");
         if (contextKey.includes("..") || contextKey.startsWith(".") || contextKey.endsWith("."))
             throw new StorageError("Context must not contain dots!");
-        const startContextKey = `${contextKey}.`;
+        const startContextKey = contextKey.length ? `${contextKey}.` : "";
         for (const key of Object.keys(this.localStorage)) {
             if (key.startsWith(startContextKey)) {
                 this.localStorage.removeItem(key);
--- a/packages/matter-node.js/src/storage/StorageBackendDisk.ts
+++ b/packages/matter-node.js/src/storage/StorageBackendDisk.ts
@@ -57,4 +57,20 @@ export class StorageBackendDisk implements Storage {
         if (!contexts.length || !key.length) throw new StorageError("Context and key must not be empty strings!");
         this.localStorage.removeItem(this.buildStorageKey(contexts, key));
     }
+
+    /** Returns all keys of a storage context without keys of sub-contexts */
+    keys(contexts: string[]): string[] {
+        if (!contexts.length) throw new StorageError("Context must not be an empty string!");
+        const contextKey = contexts.join(".");
+        if (contextKey.includes("..") || contextKey.startsWith(".") || contextKey.endsWith("."))
+            throw new StorageError("Context must not contain dots!");
+        const keys = [];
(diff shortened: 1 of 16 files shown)


packages/matter-node.js/src/storage/StorageBackendDisk.ts
```diff
@@ -57,4 +57,32 @@ export class StorageBackendDisk implements Storage {
         if (!contexts.length || !key.length) throw new StorageError("Context and key must not be empty strings!");
         this.localStorage.removeItem(this.buildStorageKey(contexts, key));
     }
+
+    /** Returns all keys of a storage context without keys of sub-contexts */
+    keys(contexts: string[]): string[] {
+        if (!contexts.length) throw new StorageError("Context must not be an empty string!");
+        const contextKey = contexts.join(".");
+        if (contextKey.includes("..") || contextKey.startsWith(".") || contextKey.endsWith("."))
+            throw new StorageError("Context must not contain dots!");
```
lauckhart, line 66: Could we get a helper method for validating the key so we don't need to copy these 4 lines?

packages/matter.js/src/CommissioningController.ts
```diff
@@ -122,7 +125,7 @@ export class CommissioningController extends MatterNode {
     /** Internal method to initialize a MatterController instance. */
     private async initializeController() {
         if (this.mdnsScanner === undefined || this.storage === undefined) {
-            throw new ImplementationError("Add the node to the Matter instance before!");
+            throw new ImplementationError("Add the node to the Matter instance before.");
```
lauckhart, line 128: OMG this diff would be a lot shorter if we were validating state in helper methods.  Could we do this now?

E.g. something like assertParented(this), assertConnected(this), assertStarted(this), etc.

You can assert "mdnsScanner" and "storage" are both defined with a single method too, [I made an example](https://www.typescriptlang.org/play?#code/MYGwhgzhAEBiD29oG8BQ0PQHZgLYFMAuaCAFwCcBLLAc2gF5oAiWc-AEyYG51MB3fJRoALUgH5iWAK64ARvnI9M0MDXwTsM+YtQBfVKgBmUrMFKV4WFVAWkAClQBuYYAE8AahfCkOACkOIxAjwAJTEkBC2MAFIlDDB0ABkKNACQqKSWgoANCpqmXIK0LoovBiUhtD+iAB0aSKkDPSMJuz4htQc0AA+3dAxNar4TS1YbR1YHCGlysqkwuTwfNj4ywCi5IvkviFKmPr6qMCWZP2IDCvLwTs8x1insmDsAOqCDQXaFwP1orcnjY92ABBfKaQrkL61IY8VARWwOSjONyeeDePwxXZHf7QGiIF5vDJgz6Mb4E0h-e6NXHwYGg6TgyHwQZqHhAA)

lauckhart (changes requested):

packages/matter-node.js/src/storage/StorageBackendDisk.ts
Apollon77: done

packages/matter.js/src/CommissioningController.ts
Apollon77: This works if the properties are public, but not when they are private (as they are currently)

Apollon77: So if you have a idea working without making that public I'm happy to adjust ... I found nothing ... I also do not like it but here we lack the Typescript support for that :-(

lauckhart (approved):

PR #413, opened 2023-10-18
Makes sure to always announce all fabrics

Apollon77:
... else they would be expired which is wrong behaviour

### DIFF
--- a/packages/matter.js/src/MatterDevice.ts
+++ b/packages/matter.js/src/MatterDevice.ts
@@ -148,29 +148,23 @@ export class MatterDevice {
         }
         const fabrics = this.fabricManager.getFabrics();
         if (fabrics.length) {
-            const fabricsToAnnounce: Fabric[] = [];
+            let fabricsToAnnounce = 0;
             for (const fabric of fabrics) {
                 const session = this.sessionManager.getSessionForNode(fabric, fabric.rootNodeId);
                 if (
-                    session !== undefined &&
-                    session.isSecure() &&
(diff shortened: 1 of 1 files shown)


packages/matter-node.js/test/IntegrationTest.ts
```diff
@@ -1505,29 +1505,15 @@ describe("Integration Test", () => {
...
-        await MockTime.yield();
-        await MockTime.yield();
-        await MockTime.yield();
-        await MockTime.yield();
-        await MockTime.advance(150);
-        await MockTime.advance(150);
-        await MockTime.yield();
-
-        await promise;
+        // For closing all down we need to use the real Time implementation
```
lauckhart, line 1508: Could we maybe do yield/advance in a loop until the expireAllAnnouncements promise resolves?  Could even do it as a utility on mock time.

lauckhart (approved):

packages/matter-node.js/test/IntegrationTest.ts
lauckhart: Oops didn't mean to approve but you get this one for free ;)

Apollon77: innthe meantime it is not just expiring it is also more stuff happening under the hood, mainly because we add more automatisms ... no idea how we can handle that in general  I add a todo in next commit to clean that up, ok?

lauckhart: Sure sounds good

PR #419, opened 2023-10-19
Only run the nightly tests on the main repo not on local forks

sammachin:
There are a couple of github actions that run on a nightly cron but if you have a fork of the repo these will likely fail.
This checks that the repo is project-chip/matter.js when then run and if not they will be skipped

### DIFF
--- a/.github/workflows/auto-approve-releases.yml
+++ b/.github/workflows/auto-approve-releases.yml
@@ -10,8 +10,8 @@ jobs:
   auto-approve:
     if: |
       github.actor == 'Automator77' &&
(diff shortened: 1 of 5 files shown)


Apollon77:
Does it work with multiple if statements on one job? Or should we combine them into one?

Apollon77 (changes requested):
Please see https://github.com/sammachin/matter.js/actions/runs/6577241130/workflow and combine the if lines. Two not supported.

Apollon77:
Ok, I adjusted it

lauckhart (approved):

PR #421, opened 2023-10-20
Fix incorrect net interface zone for UdpChannelNode on Windows

vilicvane:
@Apollon77  I did some refactoring but found out that it's not that much "from the very beginning". The refactored code (might fix some other unnoticed bugs) actually has little to do with this fix, and the direct fix might be just removing `UdpChannelNode#netInterface` (not necessary to remove now after this fix, but you know).

What do you guys think? Should I keep it after all? And does the refactoring still relevant? (The previous `getMembershipMulticastInterfaces()` might be buggy as it didn't filter out `netInterface` if provided.)


Apollon77:
Need to check this on a big screen with a fit brain tomorrow ;-) too late for today.

Apollon77 (approved):

PR #438, opened 2023-10-27
Better throw on unexpected subscription errors

Apollon77:
To expose unexpected errors we should throw them als in the case of subscription "sending of updates". So "retransmissions ended" is an expected error, all others would throw.

@vilic also as discussed


vilic:
Hmm, I won't do it this way. Though it helps crash the process if the lib user doesn't want to handle it and just let the program restarts.

Throwing the error this way, the lib user cannot handle it at all (other than handle global `unhandledException`). So, I think the best timing is after the events thing gets merged and this error should be able to be handled where the lib user initiates all these procedures.

In practice, I would catch it (or `on('error', ...)`) in the scope (exposed to lib user) that uses it and throw it again (or `emit('error', ...)`).

Apollon77:
I know the downside of this way and we have this in many places right now that way. The "error handling concept" ( if you want to call it that way) right now is "fail really to make issues transparent. In fact on this place there should only be that retransmission error. All others are unexpected and logic errors. 

Right now we do not need events at all so there idd as not error event to use and introducing this should be a global thing and not just one local place. I will add a todo for introducing events (was already on my list) but especially for errors this will be a bigger thing at all. 

Said that: you were right that the former way was hiding the errors completely and thus pr changes this. Not more not less.

Apollon77:
FYI: Todo added in project https://github.com/orgs/project-chip/projects/11?pane=issue&itemId=42962103

lauckhart (approved):
OK, I think I'm following you guys.  I think this PR is doing the right thing, here are my thoughts:

I'm not sure that unexpected errors should crash the process.  You can make the argument it could leave the process in a bad state but it's far more likely you see e.g. an undefined property reference that aborts a single buggy function.  Then everything else continues as normal.

What I expect in most servers is to see an error in the log files that allows the user to diagnose this type of issue.  But the server remains functional unless it's a sensitive code point that explicitly panics.

Likewise I'm not sure it makes sense to pass the error to the library caller.  If you throw at the point where caller awaits the server you're effectively crashing the process.  If you emit some type of event, the caller won't have context and won't be able to perform any type of recovery.  So again maybe better to just have a top-level try-catch that logs the error and moves on?

So I guess my inclination is we just make sure we've got a failsafe try/catch around all of our async operations that log the error, clean up and fail as gracefully as possible.

Of course that then leads to the issue of "untracked promises" whereby we initiate an async operation but do not keep track of it.  Then there's nobody to catch and log the error.  In this case that's in `activateSendingUpdates` and totally needs to be fixed.  We've already discussed that though and are tracking separately.

PR #458, opened 2023-10-30
Really remove text QRCode being returned

Apollon77:
Ok in one of the last PRs we removed return QR code always instead of gnerating them as needed. Seems we missed a place where it still was returned. All examples are already ot using this anymore.


lauckhart (approved):

woody-apple:
Filed: https://github.com/dropseed/pullapprove/issues/77

PR #467, opened 2023-10-31
Session close refactoring

Apollon77:
This PR is refactoring session handling especially in the area of "closing sessions" to assure that:
* Sessions send close messsages when they should and not when not
* Sessions are closed direvctly or delayed
* When closing a session delayed then it is closed after the exchange really ended (and not after the pure interaction till now - because exchanges might have a close timer to make sure all  retries happened)
* Adding Session closiure handling also for Controller 8that one was ignoring that till now)
* Adjust chip-testing to execute log tests correctly and allow to trigger log tests by commit message
*


packages/matter.js/src/protocol/ExchangeManager.ts
```diff
@@ -53,6 +56,7 @@ export class ExchangeManager<ContextT> {
     private readonly exchanges = new Map<number, MessageExchange<ContextT>>();
     private readonly protocols = new Map<number, ProtocolHandler<ContextT>>();
     private readonly transportListeners = new Array<Listener>();
+    private readonly sessionsInClosing = new Set<number>();
```
lauckhart, line 59: Kind of an odd name, maybe "closingSessions" or "sessionsInCloseSession", or do I misunderstand the meaning?

packages/matter.js/src/session/SecureSession.ts
```diff
@@ -217,16 +230,25 @@ export class SecureSession<T> implements Session<T> {
     }
 
     /** Ends a session. Outstanding subscription data will be flushed before the session is destroyed. */
-    async end(sendClose: boolean) {
+    async end(sendClose: boolean, delayClose = false) {
```
lauckhart, line 233: "delayClose" is confusing, it makes it sound like an arbitrary timeout or something.  Maybe "deferClose" or "closeAfterExchange"?

lauckhart (approved):

packages/matter.js/src/protocol/ExchangeManager.ts
Apollon77: I renamed

packages/matter.js/src/session/SecureSession.ts
Apollon77: also renamed "closingAfterExchangeFinished"

PR #468, opened 2023-10-31
Introduce Device callbacks for session and commissioning changes

Apollon77:
To allow developers to react on actions happening with the matetr device/server this PR adds:
* A new callback triggered on commissioning changes as trigger to update shown data in Device UIs or such. These are changes on commissioned fabrics (add/remove/update)
* A new callback ztriggered on activesession changes as trigger to update shown data in device UIs or such. These are new acrive sessions, disconnected sessions, but also new/removed subscriptions on sessions

### DIFF
--- a/packages/matter.js/src/MatterDevice.ts
+++ b/packages/matter.js/src/MatterDevice.ts
@@ -240,10 +240,10 @@ export class MatterDevice {
         salt: ByteArray,
         isInitiator: boolean,
         isResumption: boolean,
-        idleRetransTimeoutMs?: number,
-        activeRetransTimeoutMs?: number,
+        idleRetransmissionTimeoutMs?: number,
+        activeRetransmissionTimeoutMs?: number,
     ) {
-        const session = await this.sessionManager.createSecureSession(
+        const session = await this.sessionManager.createSecureSession({
             sessionId,
             fabric,
             peerNodeId,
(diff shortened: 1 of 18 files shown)


packages/matter.js/src/MatterDevice.ts
```diff
@@ -76,9 +76,14 @@ export class MatterDevice {
         private readonly discriminator: number,
         private readonly initialPasscode: number,
         private readonly storage: StorageContext,
-        private readonly initialCommissioningCallback: () => void,
+        private readonly commissioningChangedCallback: (fabricIndex: FabricIndex) => void,
+        private readonly sessionChangedCallback: (fabricIndex: FabricIndex) => void,
     ) {
-        this.fabricManager = new FabricManager(this.storage);
+        this.fabricManager = new FabricManager(this.storage, (fabricIndex: FabricIndex, peerNodeId: NodeId) => {
```
lauckhart, line 82: Three of the four places we use the new second parameter just stub it.  I know they're just tests but given that the callback isn't used in the constructor WDYT about just leaving as a property that FabricManager can set?

lauckhart (approved):

packages/matter.js/src/session/SecureSession.ts
```diff
@@ -71,6 +72,7 @@ export class SecureSession<T> implements Session<T> {
             encryptKey,
             attestationKey,
             closeCallback,
+            subscriptionChangedCallback,
```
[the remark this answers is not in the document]
Apollon77: done. I left the highest left as is, all others changed to object style

packages/matter.js/src/MatterDevice.ts
Apollon77: I just made it optional

PR #478, opened 2023-11-04
Update README.md

jvmahon:
Updated the supported devices table in README.md to include Matter 1.2 device types and more clearly identify the device class being referenced, also added HomeAssistant platform


Apollon77:
Thank you very much for your contribution. What you mean with "class" ? I thin "device-type-id" (maybe "device-type"? or just "id") is the official namem right? If we reference more to the specs then we should also have consistent naming, right?

jvmahon:
> Thank you very much for your contribution. What you mean with "class" ? I thin "device-type-id" (maybe "device-type"? or just "id") is the official namem right? If we reference more to the specs then we should also have consistent naming, right?

I was referring to the information in the first column shown here:  
image

I think you're right that "ID" may be better.    So as to avoid crowding the table columns, I could change Class to "ID" and a sentence before the table explaining "ID refers to the Device Type ID as specified in the MDL"

Make sense?

Apollon77:
Yes I think this would be ideal. Thank you very much

jvmahon:
> Yes I think this would be ideal. Thank you very much

Done. Thanks.

Apollon77 (approved):

jvmahon:
Not sure if people on the project CHIP software teams have good contacts at Apple / Google / Amazon / HomeAssistant / etc., but what is really needed is a "definitive" guide of device types and plans to support (if not already implemented).

In addition, one really needs to know what FeatureMap bitmap values are supported by a particular Hub. For example, each Hub supports extended color lights, but does each support Bit 0 (Hue/Saturation), Bit 1 (Enhanced Hue), or Bit 2 (Color loop). If a developer really wants to develop something across a broad range, it would be useful to know which of the FeatureMap attributes are supported (and, therefore, worth implementing, or which are just extra work with no support in the "real world."  Note sure how one gets this important information unless they can go to the definitive source (i.e., Apple, Google, Amazon, SmartThings, etc.).

Apollon77:
I agree and yes this would be awesome. The reality is that these information are not - and most likely will not - be publicly available. best that copuld be done from community is to share experiences and try to get an unofficial overview that way.

PR #491, opened 2023-11-06
Upgrade Interaction protocol revision to 11 (Matter 1.2) 

Apollon77:
This PR is implementing the following:
* Based on discovered information the event error handling in DataReports (Read and Subscription) was adjusted (https://github.com/orgs/project-chip/projects/11/views/1?pane=issue&itemId=43796949)
* Refactor and centralized ConcretePath checks.
* Server: The server was adjusted to get the interaction model revision of the senders call and also responds with this version. So depending on the caller we now answer with the same revision.
* Controller: The Controller is adjusted to specify which interaction model revision to use when sending the requests

@bzbarsky-apple and @woody-apple with your Spec knowledge it would be cool if you could check and advise if my interpretation was correct:
* Is the event error handling I found anything which was changed for model revision 10 -> 11 ? If not please provide info what else.
* Is my assumption that the device should respond ideally with the same revision as the controller correct? (A part in me is really unsure because this is certification wise interesting for devices, but on the other side also challenging for a controller because with this itcould happen that device and controller will simply not understand each other - ok not with this one change here in particular, but in general) If now it would be cool to get some insights on how this should work interoperability wise.

Thank you

Ingo

PS: deled because was bullshit ;-)


bzbarsky-apple:
> Is the event error handling I found anything which was changed for model revision 10 -> 11

As far as I know, yes.

> Is my assumption that the device should respond ideally with the same revision as the controller correct? 

No.

Apollon77:
@bzbarsky-apple  Thank you for the confirmation that my initial thought was wroing ... assumed that already because of certification thinkings.

The follow up question tho is how this should work in the future and which obligations things might bring for controllers ... There is nothing in the specification for that. As long as any interaction protocol change is "non-breaking" (e.g. by just introducing "new things that can simply be ignored") all is fine. As soon as this changes the controller need to know how to speak to the device and how to interpret the response ... Are there any thought or "controller best practices" already or will they be defined once we have such a "breaking" change case? :-)

Apollon77:
replaced by #595

PR #509, opened 2023-11-10
MatterServer node start optimizations

Apollon77:
This PR optimized the start handling in MatterServer. it mainly makes siure that nodes get also started when added to the MatterServer after this one itself was initially started (and so started all previously added nodes). With this it behaves the same for a developer irrelevant when he adds a node


Apollon77:
@GermanBluefox FYI ... as discussed today

lauckhart (approved):

PR #555, opened 2023-12-02
Fix TemperatureMeasurement definition

### LINKED ISSUE #553: Wrong Temperaturemeasurement attribute definition
Hi,
in TemperatureMeasurementCluster.ts I get:
`
            minMeasuredValue: Attribute(0x1, TlvNullable(TlvInt16.bound({ min: -27315 })), { default: 32768 }),
`
but for an INT16 the maximum number is 32767 (as correct definition Number.ts)
This break my work because I initialize all attributes with default value and I get schema error violation.
`
      throw new import_MatterError.ValidationError(`Invalid value: ${value} is above the maximum, ${this.max}.`);
`

Apollon77:
This PR adds a manual override for invalid default values from chip repo (see https://github.com/project-chip/connectedhomeip/issues/30775)

fixes #553 

@lauckhart I needed to also include attribute 0x00 and also the contrains for the other two because else the minimum values were missing after generation with the override

### DIFF
--- a/models/src/local/TemperatureMeasurementOverrides.ts
+++ b/models/src/local/TemperatureMeasurementOverrides.ts
@@ -1,0 +1,35 @@
+/**
+ * @license
+ * Copyright 2022-2023 Project CHIP Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { LocalMatter } from "../local.js";
+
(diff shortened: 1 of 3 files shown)


lauckhart (approved):
Cool, thanks. I'd assumed I'd screwed something up here

PR #580, opened 2023-12-13
Add validation handling to Invoke processing

Apollon77:
This PR is adding two things:
* Validation of request data for invoke request
* thrown ValidationError's are converted into a ConstraintError StatusCode like required in specs


packages/matter.js/src/cluster/server/CommandServer.ts
```diff
@@ -49,14 +54,39 @@ export class CommandServer<RequestT, ResponseT> {
...
+                logger.debug(`Invoke ${this.name} response : ${Logger.toJSON(response)}`);
+                return {
+                    code: StatusCode.Success,
+                    responseId: this.responseId,
+                    response: this.responseSchema.encodeTlv(response),
+                };
+            },
+            ValidationError,
+            async (error: ValidationError) => {
+                const statusError = new StatusResponseError(error.message, StatusCode.ConstraintError);
```
lauckhart, line 86: Could we make ValidationError a StatusResponseError? I've been doing that with pretty much any error that could be triggered by remote interaction.

If it's thrown locally doesn't seem like it'd be a problem. But if it doesn't seem appropriate to throw "status response" locally we could maybe move the status code into the base error class or rename StatusResponseError.

lauckhart, line 86: I'm also wondering, does ConstraintError cover everything? There's also DataTypeError and InvalidAction.

lauckhart (approved):

packages/matter.js/src/cluster/server/CommandServer.ts
Apollon77: In most places in specs I found so far in fact ConstraintError was used as soon as incoming data are "not as expected" ... many (all) of these things we handle at Validation already so I decided to make this convenient adjustment.

We have some places where we check if an error is StatusResponseError because we know itthat only can be an internal error ... we would need to adjust these places then to differentiate "generated" StatusResponseErrors from "received" ones ...

I could do that .... WDYT?

Apollon77: We have some places where we check if an error is StatusResponseError because we know itthat only can be an internal error ... we would need to adjust these places then to differentiate "generated" StatusResponseErrors from "received" ones ...

I could do that .... WDYT?

Apollon77: BTW "INVALID_DATA_TYPE" I find nowhere in specs as when to be used :-)) it is only in the StatusError list

lauckhart: Right so if we have a ValidationError on input it should go to the wire but on output we should handle more as an internal error.  Is that what you're saying?

lauckhart: Just taking the definition of INVALID_DATA_TYPE from the status table I'm assuming e.g. the ValidationError regarding type in TlvNumericSchema.validate should be INVALID_DATA_TYPE.  Took a quick perusal through CHIP, I *think* that's what they're doing.

I like the idea of the throw site selecting the status code but an alternative would be to throw different exceptions and using the class to map to different status codes at a higher level.  Kind of the same semantically.

But if decide we really only need CONSTRAINT_ERROR then this is all moot anyway. :)

Apollon77: Ok I checked all places. Adjusted ... PR gets bigger now because I needed to resolve circular imports. Please have a look

Apollon77: I would more see it that way: All other Statuscodes are explicitely stated in the specs how to use and when to return for what. if INVALID_DATA_TYPE is used in chip but is not mentioned in the specs beside the tabe then thats a big in chip or in specs . I will check and open a chip issue :-)

But good is that we an also map that easiely because I think we use UnexpectedDataError right now .. so lets check it and change afterwards when we know the result. Ok?

Apollon77: https://github.com/project-chip/connectedhomeip/issues/31020

lauckhart: Sounds good.  I think the spec addresses it OK based on the description in the table even if they don't call it out explicitly everywhere a value type is expected.

We do use UnexpectedDataError in decode but we throw ValidationError if the value isn't a number or bigint in validate.  Seems like it should probably be UnexpectedValueError there too (if the data can come from the wire) or ImplementationError (if it can only come from code).

Anyway I'll close this thread as I don't think it makes sense to block merge.

PR #582, opened 2023-12-15
High level API fixes December 23

### LINKED ISSUE #579: [BUG] commissioningController autoconnect:true fails on startup
1. When starting up a commissioningController with empty storage and the property {autoConnect: true}, the program begins searching for previously connected nodes. If it cannot find one (due to starting with empty storage), the program throws an ImplementationError, whereas it should simply exit the async connect() function.
Screenshot from 2023-12-13 16-01-03

2. If the controller has information about previously connected nodes, with the same autoConnect: true field, it starts reconnecting to all the nodes upon restart. If it fails to connect to one (e.g., the devices went offline in the meantime), it will not proceed to the next node for reconnection.

Apollon77:
This PR addresses a few issueson the high level API reported by developers, see commits and changelog (Changelog also contains missed stuff from matter.js core)

fixes #579

### DIFF
--- a/packages/matter.js/src/MatterServer.ts
+++ b/packages/matter.js/src/MatterServer.ts
@@ -237,7 +237,12 @@ export class MatterServer {
         node.setMdnsBroadcaster(this.mdnsBroadcaster);
         node.setMdnsScanner(this.mdnsScanner);
         if (this.started) {
-            await node.start();
+            try {
+                await node.start();
+            } catch (error) {
+                // TODO: Find a better way how to report back such issues and which nodes errored
+                logger.error(`Failed to start node: ${error}`);
+            }
         }
     }
 
--- a/packages/matter.js/src/MatterServer.ts
+++ b/packages/matter.js/src/MatterServer.ts
@@ -224,7 +224,12 @@ export class MatterServer {
         }
         this.started = true;
         for (const node of this.nodes) {
-            await this.prepareNode(node);
+            try {
+                await this.prepareNode(node);
+            } catch (error) {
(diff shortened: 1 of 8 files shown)


packages/matter.js/src/MatterServer.ts
```diff
@@ -224,7 +224,12 @@ export class MatterServer {
         }
         this.started = true;
         for (const node of this.nodes) {
-            await this.prepareNode(node);
+            try {
+                await this.prepareNode(node);
+            } catch (error) {
+                // TODO: Find a better way how to report back such issues and which nodes errored
```
lauckhart, line 230: Is there any identifier we could include in error message from the node object that is in scope?

Could also maybe add a todo for retry unless that happens at a higher level.

lauckhart (approved):

packages/matter.js/src/MatterServer.ts
Apollon77: We might have the unique storage key or at least the index. Will add. Good idea

For connect -now that you state it - it should normally not error out but go into "wait" state. Need to check that

Apollon77: @lauckhart See additional commit. I now also added uniqueness check for storagekey (which indeed was missing) ... and so we also have that to log. Additoonally: With current code exceptions should only occure on invalid config, so I would not do retry. If a node is commussioned then it should end in "waiting for device to come online for connection" and not throw an error

lauckhart: Looks good

PR #590, opened 2023-12-18
Session Refactoring to bring logic more near to specs

Apollon77:
Description for now short: see single commits ... I will update here b tomorrow and will also check if tests are missing and possible

### DIFF
--- a/packages/matter.js/src/MatterController.ts
+++ b/packages/matter.js/src/MatterController.ts
@@ -662,7 +662,7 @@ export class MatterController {
         );
     }
 
-    getNextAvailableSessionId() {
+    async getNextAvailableSessionId() {
         return this.sessionManager.getNextAvailableSessionId();
     }
 
--- a/packages/matter.js/src/MatterController.ts
+++ b/packages/matter.js/src/MatterController.ts
@@ -369,7 +369,8 @@ export class MatterController {
         }
 
         // Do PASE paring
-        const paseUnsecureMessageChannel = new MessageChannel(paseChannel, this.sessionManager.getUnsecureSession());
+        const unsecureSession = this.sessionManager.createUnsecureSession();
+        const paseUnsecureMessageChannel = new MessageChannel(paseChannel, unsecureSession);
         const paseExchange = this.exchangeManager.initiateExchangeWithChannel(
             paseUnsecureMessageChannel,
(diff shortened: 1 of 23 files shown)


packages/matter.js/src/MatterController.ts
```diff
@@ -662,22 +665,34 @@ export class MatterController {
...
+            fabric,
+            peerNodeId,
+            peerSessionId,
+            sharedSecret,
+            salt,
+            isInitiator,
+            isResumption,
+            idleRetransmissionTimeoutMs,
+            activeRetransmissionTimeoutMs,
+        } = args;
```
lauckhart, line 695: If I'm reading correctly, looks like you could just stick ...args into createSecureSession and avoid the de-/re-structuring?

packages/matter.js/src/protocol/ExchangeManager.ts
```diff
@@ -146,35 +146,95 @@ export class ExchangeManager<ContextT> {
...
+                );
+                this.exchanges.set(exchangeIndex, exchange);
+                await exchange.send(MessageType.StandaloneAck, new ByteArray(0));
+                await exchange.close();
+                logger.debug(
+                    `Ignoring unsolicited message ${messageId} for protocol ${message.payloadHeader.protocolId}.`,
+                );
+            } else {
+                if (protocolHandler === undefined)
+                    throw new MatterFlowError(`Unsupported protocol ${message.payloadHeader.protocolId}`);
```
lauckhart, line 232: This is coming from the wire?  Maybe log a different message from below but as warning?  Don't like malformed wire messages to create what's effectively an internal error

lauckhart (approved):
I didn't compare super closely to the spec but it's got to be an improvement

packages/matter.js/src/MatterController.ts
Apollon77: right, adjusted in two places

packages/matter.js/src/protocol/ExchangeManager.ts
Apollon77: Yes and nmo ... ideally we should jhave all relevant ProtocolHandlers registered ... so when a mesage comes in with unexpected one could also be on us ...

Apollon77: but I adjust to warning logging

PR #593, opened 2023-12-21
Declare current examples Legacy by renaming files

Apollon77:
To prepare the ground for converted exmaples 8and keep git kleen before I create files with the same name again as the moved ones) lets declare them legacy for now


lauckhart (approved):
Hoo boy you're optimistic ;)

PR #595, opened 2023-12-23
Upgrade Interaction protocol revision to 11 (Matter 1.2) - Redone

Apollon77:
This PR is implementing the following:
* Based on discovered information the event error handling in DataReports (Read and Subscription) was adjusted (https://github.com/orgs/project-chip/projects/11/views/1?pane=issue&itemId=43796949)
* Refactor and centralized ConcretePath checks.
* Interaction protocol version is increased to 11
* Some tests are enhanced and asjusted accordingly

This is a recreated simplified version of #491 and replaces this.

### DIFF
--- a/packages/matter.js/src/protocol/interaction/InteractionServer.ts
+++ b/packages/matter.js/src/protocol/interaction/InteractionServer.ts
@@ -58,7 +58,7 @@ import { StatusCode, StatusResponseError } from "./StatusCode.js";
 import { SubscriptionHandler } from "./SubscriptionHandler.js";
 
 export const INTERACTION_PROTOCOL_ID = 0x0001;
-export const INTERACTION_MODEL_REVISION = 10;
+export const INTERACTION_MODEL_REVISION = 11;
 
 const logger = Logger.get("InteractionServer");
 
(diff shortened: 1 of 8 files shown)


packages/matter.js/src/protocol/interaction/InteractionServer.ts
```diff
@@ -381,39 +415,48 @@ export class InteractionServer implements ProtocolHandler<MatterDevice> {
                 const { path, dataVersion } = values[0];
                 const attributes = this.endpointStructure.getAttributes([path], true);
```
lauckhart, line 416: Hmm ok after reading through this bit maybe should be separate PR and/or on device-gen branch.

If the path is insufficiently concrete I think response should be INVALID_ACTION and processing should abort.  But if I'm reading correctly we'd just ignore here. 

I'm also wondering if logic could be more straightforward if we determine if the attribute exists then check to see if it's writtable. With the current flow I think we ask for the attribute then have to determine whether it wasn't returned because because it didn't exist or because it was read-only. 

Doing this on mobile so I may be reading incorrectly. Parents are sleeping in the office with my workstation. 😉

Apollon77: Lol, all good, no need to hurry.

In fact we are again at:

if this getAttributes for one path returns nothing (like in this case) then this attribute bath does not exist! So it might be a valid "concrete path" but we do not have this cluster or attribute or endpoint ... This according to specs should return unsupportedWrite ...

But in fact you are right. the "validateConcrete*Path" methods should/could also check for the "invalidAction " cases ... they do not do that right now.

lauckhart: OK, I already added the concrete path validation in the device-gen branch so maybe we merge this and we can look at simplifying during/after merge into device-gen

lauckhart (approved):

packages/matter.js/src/protocol/interaction/InteractionServer.ts
Apollon77: I added a TODO

PR #636, opened 2024-01-15
Discovery and Session Refactoring to match 1.2 specs better

Apollon77:
This PR will update several session timings and intervals for resubbkissions as defiened by 1.2 specs and especially increases the conformance for pairing with battery based devices with usually much higher session intervals. It makes sure to use the discovery details throughout the hole process and store the details for usage for reconnections and such.

I did very the behaviour in manual tests with 4 different battery based device types and also other wifi and thread device types and commissionings of them.

* Matter-Core functionality:
  * Enhancement: Implemented handling for session interval parameters as defined by Matter 1.2 specification
  * Enhancement: Improved discovery data handling and use MDNS data for Pase/Case connections session interval parameters
  * Enhancement: Storing session parameter with session resumption details to reuse on session restores
  * Fix: Only set session active timestamp if we create a session based on an incoming message and not when we are the creator of the session to prevent too fast resubmissions
* matter.js API:
  * Enhancement: Stored Discovery and Basic information data for commissioned nodes and allow API access for easy determination of devices without need to connect to them
* matter.js shell:
  * Enhancement: Adjusted logic to output detailed node information on nodes command

more details, see commits


packages/matter.js/src/common/Scanner.ts
```diff
@@ -51,6 +31,43 @@ export type CommissionableDevice = {
 
     /** Pairing instructions */
     PI?: string;
+
+    /** Sleep Idle Interval */
+    SII?: number;
```
lauckhart, line 36: Is this sufficiently camel-cased?

packages/matter.js/src/session/SecureSession.ts
```diff
@@ -107,8 +90,10 @@ export class SecureSession<T> implements Session<T> {
             attestationKey,
             closeCallback,
             subscriptionChangedCallback,
-            idleRetransmissionTimeoutMs,
-            activeRetransmissionTimeoutMs,
+            idleIntervalMs,
```
lauckhart, line 93: Maybe we could clump the intervals into a single object since they travel together.  Keys and/or other authenticated peer information might be candidates too

lauckhart (approved):

packages/matter.js/src/common/Scanner.ts
Apollon77: I choose to be 1:1 to the MDNS txt key. That's why all uppercase. Wdyt?

lauckhart: Yeah all good, was just kidding :)

packages/matter.js/src/CommissioningController.ts
```diff
@@ -279,9 +284,60 @@ export class CommissioningController extends MatterNode {
...
         return pairedNode;
     }
 
+    private async enhanceDeviceDetails(nodeId: NodeId, pairedNode: PairedNode, dataAvailableLocally: boolean) {
+        const controller = this.assertControllerIsStarted();
+
+        const globalAttributeKeys = Object.keys(GlobalAttributes({}));
+        if (dataAvailableLocally) {
+            const basicInformationClient = pairedNode.getRootClusterClient(BasicInformation.Cluster);
+            if (basicInformationClient !== undefined) {
```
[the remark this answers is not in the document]
Apollon77: Done

packages/matter.js/src/CommissioningController.ts
```diff
@@ -279,9 +284,60 @@ export class CommissioningController extends MatterNode {
...
+                const basicInformationData = {} as Record<string, SupportedStorageTypes>;
+                for (const attributeName of Object.keys(basicInformationClient.attributes)) {
+                    if (!globalAttributeKeys.includes(attributeName)) {
```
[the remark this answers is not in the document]
Apollon77: makes totally sense ...

packages/matter.js/src/session/SecureSession.ts
Apollon77: Very good idea ... I adjusted it ...

PR #645, opened 2024-01-16
Standalone Ack Improvements

Apollon77:
This PR enhances the standalone ack handling as defined in Matter 1.2 specification. It cleans up some invalid implementation from the past and also introduces the Standalone ack timeout (200ms) to send standalone ack even when no new message is sent out within this timeframe.


packages/matter.js/src/protocol/MessageExchange.ts
```diff
@@ -110,6 +117,15 @@ export class MessageExchange<ContextT> {
     private readonly retransmissionRetries: number;
     private readonly messagesQueue = new Queue<Message>();
     private receivedMessageToAck: Message | undefined;
+    private receivedMessageAckTimer = Time.getTimer(MRP_STANDALONE_ACK_TIMEOUT_MS, () => {
+        if (this.receivedMessageToAck !== undefined) {
+            const messageToAck = this.receivedMessageToAck;
+            this.receivedMessageToAck = undefined;
+            this.sendStandaloneAckForMessage(messageToAck).catch(error =>
```
lauckhart, line 124: We don't have a standard way of tracking "fire-and-forget" operations right now but maybe could add a TODO?

Reasoning is if we shutdown while the ack is outstanding it could crash when it processes, and if this type of operation were o hang they would pile up indefinitely and could cause leaks.

packages/matter.js/src/protocol/MessageExchange.ts
```diff
@@ -202,18 +228,42 @@ export class MessageExchange<ContextT> {
...
-        const { expectAckOnly = false, minimumResponseTimeoutMs, requiresAck } = options ?? {};
-        if (this.sentMessageToAck !== undefined)
+        const {
+            expectAckOnly = false,
+            minimumResponseTimeoutMs,
+            requiresAck,
+            includeAcknowledgeMessageId,
+        } = options ?? {};
+        if (messageType === MessageType.StandaloneAck && requiresAck) {
+            throw new MatterFlowError("A standalone ack must not send reliable.");
```
lauckhart, line 251: This message is a little hard to parse, maybe "A standalone ack may not require acknowledgement"?

lauckhart (approved):

packages/matter.js/src/protocol/MessageExchange.ts
Apollon77: TODO added

packages/matter.js/src/protocol/MessageExchange.ts
Apollon77: done

PR #655, opened 2024-01-25
Update Aggregator.ts

Luligu:
Must match the
BRIDGED_DEVICE_WITH_POWERSOURCE_INFO: DeviceTypeDefinition
in packages\matter.js\src\device\DeviceTypes.ts.


Apollon77 (approved):
You are right!

PR #656, opened 2024-01-25
Commissioning fixes

Apollon77:
This PR includes some commissioning fixes that were found by running the most current MatterSDK tests against matter.js

For details see commits

### DIFF
--- a/packages/matter.js/src/cluster/server/AdministratorCommissioningServer.ts
+++ b/packages/matter.js/src/cluster/server/AdministratorCommissioningServer.ts
@@ -6,6 +6,7 @@
 
 import { MatterDevice } from "../../MatterDevice.js";
 import { InternalError } from "../../common/MatterError.js";
+import { CRYPTO_GROUP_SIZE_BYTES, CRYPTO_PUBLIC_KEY_SIZE_BYTES } from "../../crypto/CryptoConstants.js";
 import { FabricIndex } from "../../datatype/FabricIndex.js";
 import { VendorId } from "../../datatype/VendorId.js";
 import { Logger } from "../../log/Logger.js";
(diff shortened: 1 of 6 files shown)


packages/matter.js/src/cluster/server/AdministratorCommissioningServer.ts
```diff
@@ -23,6 +28,33 @@ const logger = Logger.get("AdministratorCommissioningServer");
 export const MAXIMUM_COMMISSIONING_TIMEOUT_S = 15 * 60; // 900 seconds/15 minutes
 export const MINIMUM_COMMISSIONING_TIMEOUT_S = 3 * 60; // 180 seconds/3 minutes
 
+const PAKE_PASSCODE_VERIFIER_LENGTH = CRYPTO_GROUP_SIZE_BYTES + CRYPTO_PUBLIC_KEY_SIZE_BYTES;
+
+/**
+ * Monkey patching Tlv Structure of openCommissioningWindow command to prevent data validation of the fields to be
+ * handled as ConstraintError because we need to return a special error.
+ * We do this to leave the model in fact for other validations and only apply the change for our Schema-aware Tlv parsing.
+ */
```
lauckhart, line 37: Hoo boy OK, I feel like you're chasing random bits of c++ code 🙂

Feels like this might benefit from a lower-level hook where we can just remap the error code

lauckhart (approved):

packages/matter.js/src/cluster/server/AdministratorCommissioningServer.ts
Apollon77: Yes ... an error handler hook or "do not do validate on schema decode" flag or such ... but as discussed ... lets see what we can do in device-branch because there this monkey patch do not work

PR #659, opened 2024-01-26
Migrate to python driven chip tests

Apollon77:
This PR is adjusting the former "integrated in chip tool testing" to the new yaml based python runner using chip tool and also a first pure python test case. For this the PR removes the old JS based test runner and just offers comparable test apps and add instructions (and CI) to run the tests using the Matter SDK. For this the full matter sdk checkout is needed.

This PR additionally includes:
* two more commissiong flow adjustments found while finalizing the tests
* some minimal BLE tweaks
* allow the "command line "logic to be called with --name additionally to -name as before.

The latter two changes got committed  unintended ... sorry


packages/matter-node-ble.js/src/ble/NobleBleChannel.ts
```diff
@@ -79,6 +80,10 @@ export class NobleBleCentralInterface implements NetInterface {
                 `Peripheral ${address.peripheralAddress} is already connected. Only one connection supported right now.`,
             );
         }
+        if (peripheral.state !== "disconnected") {
+            // Try to cleanup strange in between states
```
lauckhart, line 84: Strange = storage?

packages/matter-node.js/test/IntegrationTest.ts
```diff
@@ -963,6 +962,7 @@ describe("Integration Test", () => {
             assert.deepEqual(pushedUpdates, [{ value: true, time: startTime + 2 * 1000 + 100 }]);
         });
 
+        /*
```
lauckhart, line 965: Could include a note on why commented and/or when to uncomment

lauckhart (approved):

packages/matter-node.js/test/IntegrationTest.ts
Apollon77: agreed

packages/matter-node-ble.js/src/ble/NobleBleChannel.ts
Apollon77: no really 

strange "in between states"

Apollon77: adjustes text

PR #667, opened 2024-01-29
Correctly set Group key limits for now

Apollon77:
Because we do not support Groups for now we should just set the limits correctly to 1 (IPK).


vves (approved):
perfect

PR #671, opened 2024-02-01
Shell enhancements for Attributes, Events and Comments

Apollon77:
This PR adds the following:
* Enhance ValidationError to transport the "structuure-aware" fieldname of the affected field to make finding the error more easy
* Enhance the Shell by programmatically generating yargs structure for azttribues, commands and events
* ENhanced the Shell readme a lot with many information and examples


packages/matter-node-shell.js/src/util/String.ts
```diff
@@ -0,0 +1,9 @@
+/**
+ * @license
+ * Copyright 2022-2023 Project CHIP Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+export function firstLetterToLowerCase(str: string) {
```
lauckhart, line 7: I use "camelize" for this, kind of overkill but shorter to type :)

lauckhart (approved):

packages/matter-node-shell.js/src/util/String.ts
Apollon77: Changed

PR #672, opened 2024-02-01
Allow multiple loggers and add File Logger

Apollon77:
This PR does:
* Allowed multiple Loggers and log targets to be registered. Logging to console is still default
* Add a File logger in Node.js package
* Added support in shell for Debug logging into a Logfile additionally to e.g. Info logging in console


packages/matter-node.js/src/log/FileLogger.ts
```diff
@@ -0,0 +1,28 @@
+/**
+ * @license
+ * Copyright 2022-2023 Project CHIP Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { Level } from "@project-chip/matter.js/log";
+import { createWriteStream } from "fs";
```
lauckhart, line 8: It would be probably be better to use async I/O for this although if node buffers internally it might not matter.  NodeJsActionTracer.ts in device-gen is an example of using async I/O with a public API that's synchronous

lauckhart (approved):

packages/matter-node.js/src/log/FileLogger.ts
Apollon77: adjusted

PR #694, opened 2024-02-16
Devices-Gen test Fixes and Updates

Apollon77:
This PR contains all my collected fixes and updates that makes test basically working ... Details see commits ... Tell me where I did wrong :-)


packages/matter.js/src/behavior/definitions/identify/IdentifyServer.ts
```diff
@@ -18,4 +18,52 @@ const Base = IdentifyBehavior.for(Identify.Complete);
...
+
+    override initialize() {
+        this.internal.identifyTimer = Time.getPeriodicTimer(
+            "Identify time update",
+            1000,
+            this.callback(this.#identifyTick),
+        );
+
+        // So whenever the attribute OR the identify command was invoked we react to it.
+        this.events.identifyTime$Change.on(() => {
```
lauckhart, line 31: This should probably use reactTo

packages/matter.js/src/cluster/server/AdministratorCommissioningServer.ts
```diff
@@ -198,11 +198,13 @@ class AdministratorCommissioningManager {
      * This method is used internally when the commissioning window timer expires or the commissioning was completed.
      */
     private endCommissioning() {
-        logger.debug("End commissioning window.");
         if (this.commissioningWindowTimeout !== undefined) {
             this.commissioningWindowTimeout.stop();
             this.commissioningWindowTimeout = undefined;
+        } else {
```
lauckhart, line 204: Trivial - should invert the tests and just return if this.commissioningWindowTimeout === undefined

packages/matter.js/src/environment/MdnsService.ts
```diff
@@ -32,15 +32,16 @@ export class MdnsService {
             const vars = environment.get(VariableService);
             const network = environment.get(Network);
 
-            const enableIpv4 = options?.ipv4 ?? environment.get(VariableService).boolean("mdns.ipv4") ?? true;
+            const netInterface = vars.get("mdns.networkInterface", options?.networkInterface);
+            const enableIpv4 = vars.boolean("mdns.ipv4") ?? options?.ipv4 ?? true;
             this.#broadcaster = await MdnsBroadcaster.create(network, {
                 enableIpv4,
-                multicastInterface: vars.get("mdns.discoverInterface", options?.discoverInterface),
```
lauckhart, line 38: I thought it was nice allowing for announcement and discovery on different interfaces but nbd

packages/matter.js/src/node/NodeLifecycle.ts
```diff
@@ -52,6 +56,13 @@ export class NodeLifecycle extends PartLifecycle {
         return this.#online;
     }
 
+    /**
+     * Emits when the node gets initialized
```
lauckhart, line 60: Should maybe detail how this is different from "ready"

lauckhart (approved):
LGTM, I made some minor comments but we can address after merge

packages/matter.js/src/behavior/definitions/identify/IdentifyServer.ts
Apollon77: Ok ... for info and understanding: why? ;-)

packages/matter.js/src/environment/MdnsService.ts
Apollon77: We decided that several time back on master that it makes no sense to have it different. ;-)

packages/matter.js/src/behavior/definitions/identify/IdentifyServer.ts
Apollon77: Ok I think. got the reason ... we just read state so should be ok, but yes reactTo is better.

PR #706, opened 2024-02-26
A wrong operator, in the right place?

grkek:
I think there is an issue with this, how can one message type be both at the same time?

### DIFF
--- a/packages/matter.js/src/mdns/MdnsServer.ts
+++ b/packages/matter.js/src/mdns/MdnsServer.ts
@@ -92,7 +92,7 @@ export class MdnsServer {
         const message = DnsCodec.decode(messageBytes);
         if (message === undefined) return; // The message cannot be parsed
         const { transactionId, messageType, queries, answers: knownAnswers } = message;
-        if (messageType !== DnsMessageType.Query && messageType !== DnsMessageType.TruncatedQuery) return;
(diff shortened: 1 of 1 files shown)


grkek:
I am guessing this is a TypeScript thing

Apollon77:
In fact not Typescript related ... this code place only handles Queries and Truncated Queries ... so if the type is not matching both of them we ignore it.

grkek:
> In fact not Typescript related ... this code place only handles Queries and Truncated Queries ... so if the type is not matching both of them we ignore it.

How can one variable equal both at the same time, wouldn't || operator be more suitable?

Apollon77:
Read the code correctly ... it check for unequalness : !== ... but maybe the two equal-signs confuse you :-)

JavaScript know to different ways to compare:

* "==" and "!=" do a value wise comparism by ignoring the type, so "false == 0" or "0 == ''" would be true as example
* "===" and "!==" do a value AND typewise comparism ...

grkek:
> Read the code correctly ... it check for unequalness : !== ... but maybe the two equal-signs confuse you :-)
> 
> JavaScript know to different ways to compare:
> 
> * "==" and "!=" do a value wise comparism by ignoring the type, so "false == 0" or "0 == ''" would be true as example
> * "===" and "!==" do a value AND typewise comparism ... means "false === 0" is false becasue the type does not match

Had to zoom in to see, thank you!

PR #708, opened 2024-03-01
Update fixes after Master-Merge 2nd march 24

Apollon77:
see commits. With thiese fixes the chip tests run again

### DIFF
--- a/packages/matter.js-tools/package.json
+++ b/packages/matter.js-tools/package.json
@@ -55,7 +55,8 @@
         "yargs": "^17.7.2"
     },
     "optionalDependencies": {
-        "v8-profiler-next": "^1.10.0"
+        "v8-profiler-next": "^1.10.0",
+        "@esbuild/darwin-x64": "^0.20.1"
     },
     "devDependencies": {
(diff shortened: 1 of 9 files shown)


packages/matter.js-tools/package.json
```diff
@@ -55,7 +55,8 @@
         "yargs": "^17.7.2"
     },
     "optionalDependencies": {
-        "v8-profiler-next": "^1.10.0"
+        "v8-profiler-next": "^1.10.0",
+        "@esbuild/darwin-x64": "^0.20.1"
```
lauckhart, line 59: If this causes the optional dependencies to be locked when node_modules is present but package_lock.json is not then we should include all of esbuild's platform dependencies here although I'm not sure how important AIX support is. :)  Otherwise we should probably take this out and perhaps have npm install abort if package-lock.json is missing but node_modules is present.

lauckhart (approved):

packages/matter.js-tools/package.json
Apollon77: will be fixed by new PR

PR #710, opened 2024-03-03
Commissioning events and Sensor example

Apollon77:
This PR adds Commissioning changed events and re-wires/refactors the Fabric Manager to be event based.

It also adds a Sensor example


packages/matter.js/src/fabric/FabricManager.ts
```diff
@@ -6,74 +6,96 @@
...
-    private nextFabricIndex = 1;
-    private readonly fabrics = new Map<FabricIndex, Fabric>();
+    #nextFabricIndex = 1;
+    readonly #fabrics = new Map<FabricIndex, Fabric>();
+    #initializationDone = false;
+    #fabricStorage: StorageContext;
+    #events = {
+        added: Observable<[fabric: Fabric]>(),
+        updated: Observable<[fabric: Fabric]>(),
+        removed: Observable<[fabric: Fabric]>(),
```
lauckhart, line 32: Nbd but maybe use "deleted" to match conventions

lauckhart (approved):

packages/matter.js/src/fabric/FabricManager.ts
Apollon77: done too

PR #713, opened 2024-03-06
Fabrics adjustment and Environment aware Controller

Apollon77:
This PR is containing my last Todos (beside peer deps) - can be reviewed by commit
It is mainly containing:
* Further adjustments and Failsafe cleanups regarding the automatically updating fabrics structures for new API
* Environment Aware Controller with an already updatedStorage structure

Old Controller STorage was:
* start with Matter-Server unique ID (e.g. 0)
```
0.MatterController.commissionedNodes
0.MatterController.fabric
0.RootCertificateManager.nextCertificateId
0.RootCertificateManager.rootCertBytes
0.RootCertificateManager.rootCertId
0.RootCertificateManager.rootKeyIdentifier
0.RootCertificateManager.rootKeyPair
0.SessionManager.resumptionRecords
```

new structure would be
* the node id is own storage, so "0" is absent here
```
credentials.fabric
credentials.nextCertificateId
credentials.rootCertBytes
credentials.rootCertId
credentials.rootKeyIdentifier
credentials.rootKeyPair
sessions.resumptionRecords
nodes.commissionedNodes
```

The "nodes.commissionedNodes" is for now an array with all the nodes and their meta data. I would porptopse to accept for now and "Migrate" later if needed into an array and substructure. it is easiely convertable later.


packages/matter.js/src/behavior/definitions/general-commissioning/ServerNodeFailsafeContext.ts
```diff
@@ -10,19 +10,14 @@ import { Endpoint } from "../../../endpoint/Endpoint.js";
...
 
 /**
  * {@link FailsafeContext} for {@link Node} API.
  */
 export class ServerNodeFailsafeContext extends FailsafeContext {
     #node: Node;
-    #operationalCredentialsRestored = false;
     #storedState?: {
         networks: Map<Endpoint, NetworkCommissioningBehavior.State["networks"]>;
-        nocs: OperationalCredentialsBehavior.State["nocs"];
```
lauckhart, line 23: I feel like we're losing some code here that works with "native" matter semantics.  I know it's kind of redundant right now but if we move to using the native attributes as backing for Fabric API then we'll have to recreate this.  Just commenting it out could lead to bitrot but it might be nice to have as reference in the future.

lauckhart (approved):

PR #720, opened 2024-03-08
Accept zero-length nullable fields (behavior only)

lauckhart:
Only tackles behavior state, still need to do TLV

### DIFF
--- a/packages/matter.js/src/behavior/state/validation/constraint.ts
+++ b/packages/matter.js/src/behavior/state/validation/constraint.ts
@@ -42,8 +42,15 @@ export function createConstraintValidator(
 
         case Metatype.string:
         case Metatype.bytes:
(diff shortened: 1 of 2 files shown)


Apollon77 (commented):
Hm ...  do we need this here really?

Should we not force the developers to behave correctly. -even if the matter data standard allows such strange things?

Else we need to handle always the "there but empty" case in all command implementations. My idea would be to make sure on Tlv level to acceopt and correctly convert it when parsing ... but honestly I would loveto be sure that we do nt also need it internally. SO honestly I would püropose to scratch this PR in favor or "accurate code" in matter.js ... and just make sure we handle such incoming data correctly on Tlv level.

WDYT?

lauckhart:
Yeah I think that's fine

PR #721, opened 2024-03-08
Change NetworkServer address configuration

lauckhart:
Now takes addresses as a list of listening address instead of one each of IPv4, IPv6 and BLE.  Adds support for dual-mode (IPv4 + IPv6) sockets.

Adds one dual-mode UDP listener on all addresses if none is configured.  Adds one BLE listener if BLE is supported and "ble" option is not set to false.

Throws an error if there are multiple BLE addresses or the HCI address is specified in the listener list (these require lower-level changes I did not make yet).

I want to make similar changes for Mdns at some point but figured I'd see how these go over first.

Includes some additional convenience methods for variable access.

### DIFF
--- a/packages/matter.js/src/behavior/system/network/NetworkServer.ts
+++ b/packages/matter.js/src/behavior/system/network/NetworkServer.ts
@@ -5,7 +5,7 @@
  */
 
 import { Ble } from "../../../ble/Ble.js";
-import { ImplementationError } from "../../../common/MatterError.js";
+import { ImplementationError, NotImplementedError } from "../../../common/MatterError.js";
 import { Logger } from "../../../log/Logger.js";
 import { SubscriptionOptions } from "../../../protocol/interaction/SubscriptionOptions.js";
 import { TypeFromPartialBitSchema } from "../../../schema/BitmapSchema.js";
@@ -27,13 +27,11 @@ export class NetworkServer extends NetworkBehavior {
     declare internal: NetworkServer.Internal;
 
     override initialize() {
-        if (this.state.ble === undefined) {
-            // TODO make working again when State init gets fixed!
-            this.state.ble = Ble.enabled;
-        } else if (this.state.ble && !Ble.enabled) {
-            logger.warn("Disabling Bluetooth commissioning because BLE support is not installed");
-            this.state.ble = false;
-        }
+        const vars = this.endpoint.env.vars;
+
+        this.state.port = vars.number("network.port", this.state.port);
+
+        this.state.listen = this.#configureListeners(vars.list("network.listen", this.state.listen));
 
         const discoveryCaps = this.state.discoveryCapabilities;
         switch (discoveryCaps.ble) {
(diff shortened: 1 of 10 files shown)


lauckhart:
Oh yeah I forgot, this also includes changes to UDP API so that we don't turn REUSE_ADDR sockopt on by default but only when necessary.  This prevents users from accidentally starting multiple servers on the same port and is actually why I was in this code in the first place.

packages/matter.js/src/behavior/system/network/NetworkServer.ts
```diff
@@ -101,21 +99,156 @@ export class NetworkServer extends NetworkBehavior {
...
     export class State extends NetworkBehavior.State {
-        listeningAddressIpv4?: string = undefined;
-        listeningAddressIpv6?: string = undefined;
-        ipv4 = true;
+        /**
+         * An array of {@link Address} objects configuring the interfaces the server listens on.
+         *
+         * Configurable also with variable "network.listen".  You may configure a single listener using:
+         *
+         *   * `network.listen.protocol` either "ble", "udp4", "udp6" or "udp" (default is "udp" for dual IPv4/6)
```
Apollon77, line 220: we need to see where we document all these ... I started as main comment also on Environment  (in this case the NodeEnvironment). So we need to update there.

packages/matter.js/src/behavior/system/network/ServerNetworkRuntime.ts
```diff
@@ -81,22 +81,56 @@ export class ServerNetworkRuntime extends NetworkRuntime {
...
+                    const udp = await UdpInterface.create(
+                        this.owner.env.get(Network),
+                        address.protocol,
+                        address.port,
+                        address.address,
+                    );
+
+                    transports.add(udp);
+
+                    // We advertise the first (most likely only) UDP port we open as our official address
```
Apollon77, line 109: I would notz allow different ports at all, if not announced noone will use it ... so honestly I see no sense in allowing different ports

packages/matter.js/src/behavior/system/network/ServerNetworkRuntime.ts
```diff
@@ -81,22 +81,56 @@ export class ServerNetworkRuntime extends NetworkRuntime {
...
+                        const state = agent.network.state;
+                        if (state.operationalPort === -1) {
+                            state.operationalPort = udp.port;
+                        }
+                    });
+
+                    break;
+
+                case "ble":
+                    // TODO - HCI ID and/or multiple BLE transports?  (Former seems valuable, latter not really but
```
Apollon77, line 120: just a single BLE is making sense. forbid multiple

packages/matter.js/src/behavior/system/network/ServerNetworkRuntime.ts
```diff
@@ -81,22 +81,56 @@ export class ServerNetworkRuntime extends NetworkRuntime {
...
+                    // would include for completeness)
+                    const ble = Ble.get().getBlePeripheralInterface();
+                    transports.add(ble);
+                    bleTransports.add(ble);
+                    break;
+            }
+        }
+
+        // AFAICT no way to query transport interface for protocol type so just store BLE transports separately so we
+        // can identify them for removal after commissioning completes
```
Apollon77, line 130: BLE is by definition commisioning only

packages/matter.js/src/behavior/system/network/ServerNetworkRuntime.ts
```diff
@@ -187,20 +187,23 @@ export class ServerNetworkRuntime extends NetworkRuntime {
             this.#bleBroadcaster = undefined;
         }
 
-        if (this.#bleTransport) {
-            this.owner.env.runtime.add(this.#removeBleTransport(this.#bleTransport));
-            this.#bleTransport = undefined;
+        if (this.#bleTransports) {
+            for (const ble of this.#bleTransports) {
```
Apollon77, line 191: only one?!

Apollon77 (changes requested):
See comments. I think we should include hciId support directly because thats needed if you want to use ble in many cases.

We can also change ble.enable and ble.hciId that I used in example full to the new style network.* stuff and such, no problem

packages/matter.js/src/behavior/system/network/NetworkServer.ts
```diff
@@ -101,21 +99,156 @@ export class NetworkServer extends NetworkBehavior {
...
+        const listen = Array<NetworkServer.Address>();
+        let hasUdp = false;
+        let hasBle = false;
+        let disabledBle = false;
+        for (const addr of config) {
+            if (typeof addr !== "object") {
+                throw new ImplementationError("Listen address is not an object");
+            }
+
+            let { protocol, port } = addr as Record<string, any>;
```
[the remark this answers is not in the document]
lauckhart: OK, I changed to "transport"

packages/matter.js/src/behavior/system/network/NetworkServer.ts
```diff
@@ -101,21 +99,156 @@ export class NetworkServer extends NetworkBehavior {
...
+            switch (protocol) {
+                case "ble":
+                    if (Ble.enabled) {
+                        if (hasBle) {
+                            throw new NotImplementedError("Currently only a single BLE transport is allowed");
+                        } else {
+                            hasBle = true;
+                        }
+                        if (address !== undefined) {
+                            throw new NotImplementedError("Currently you may not specify HCI ID for BLE transport");
```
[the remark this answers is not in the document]
lauckhart: That's the goal but will require refactoring Ble.ts to convert Ble from a singleton and add HCI option.  Then we need to make it so multiple servers that share the same BLE address can share the broadcaster.  We need something similar for the MDNS side of things so figured it might make sense to do that first.

So anyway, thought this was a good intermediate place to stop.

packages/matter.js/src/behavior/system/network/NetworkServer.ts
lauckhart: I think in a separate .md file would be ideal, I've been planning to document in the code near where the options are used.  Environment seems to generic to document the vars, many of which will be irrelevant if you're a controller vs. device.

This was one of the things I was going to address in the documentation push.

packages/matter.js/src/behavior/system/network/ServerNetworkRuntime.ts
lauckhart: Yes, this is so we can identify which ones to remove

packages/matter.js/src/behavior/system/network/NetworkServer.ts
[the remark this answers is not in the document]
Apollon77: Maybe lets have a chat about that ... current realizy on nodejs and also on mobile devices is that there is in fact one BLE radio to use ... also bleno/noble is kind of limited to one right now and weould need a rewrite to make that configurable. Multiple servers using the same BLE radio in fact make not sense at all ... that would only be a topic when bleno would be rewritten for such a support (there was such a PR somewhen but would be need to revamped completely). So in my eyes ble can only work for one of the Server/Controllers on the Host right now.

Ok for MDNS I thought we have whats needed already, but ok maybe I miss something ... so maybe lets have a chat what you plan/imagine...

packages/matter.js/src/behavior/system/network/ServerNetworkRuntime.ts
Apollon77: ... and see above.formally only one BLE per host ... soooo multiple servers the server should define if ble should be (not) added too?

packages/matter.js/src/behavior/system/network/ServerNetworkRuntime.ts
Apollon77: was mainly regarding the "AFAICT" :-)

PR #722, opened 2024-03-08
Remove all code related to persistent versioning

lauckhart:
Turns out that Matter isn't really designed for this and connectedhomeip doesn't do it.


Apollon77 (approved):

Apollon77:
As reference https://github.com/project-chip/connectedhomeip/issues/32411#event-12019115593

PR #725, opened 2024-03-09
Also remove Data version persistence in legacy code path

Apollon77:
As already done for the new API in #722 , also remove data version persistence for the legacy API


Apollon77:
As reference: https://github.com/project-chip/connectedhomeip/issues/32411#event-12019115593

lauckhart (approved):

PR #729, opened 2024-03-10
Add convenience infos for node implementors

Apollon77:
This PR adds:
* Allows to announce a Node als later as convenience method on SeverNode
* Add a "exposed Fabric Infos" object on CommissioningBehaviorfor access like we had in old API

### DIFF
--- a/packages/matter.js/src/behavior/system/commissioning/CommissioningBehavior.ts
+++ b/packages/matter.js/src/behavior/system/commissioning/CommissioningBehavior.ts
@@ -100,11 +100,11 @@ export class CommissioningBehavior extends Behavior {
         const fabrics = this.endpoint.env.get(FabricManager).getFabrics();
         const commissioned = !!fabrics.length;
         if (fabricAction === FabricAction.Removed) {
-            delete this.state.commissionedFabrics[fabricIndex];
+            delete this.state.fabrics[fabricIndex];
         } else {
             const fabric = fabrics.find(fabric => fabric.fabricIndex === fabricIndex);
             if (fabric !== undefined) {
-                this.state.commissionedFabrics[fabricIndex] = {
+                this.state.fabrics[fabricIndex] = {
                     fabricIndex: fabric.fabricIndex,
                     fabricId: fabric.fabricId,
                     nodeId: fabric.nodeId,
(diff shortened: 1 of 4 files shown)


packages/matter.js/src/node/ServerNode.ts
```diff
@@ -189,6 +189,10 @@ export class ServerNode<T extends ServerNode.RootEndpoint = ServerNode.RootEndpo
         await this.#mutex;
     }
 
+    async advertiseNow() {
+        await this.act(agent => agent.get(NetworkServer).advertiseNow());
```
lauckhart, line 193: This is why I was thinking a generic convenience method for invoking behavior methods might be nice... E.g. `await node.invoke.network.advertiseNow()`

packages/matter.js/src/behavior/system/commissioning/CommissioningBehavior.ts
```diff
@@ -235,6 +260,7 @@ export namespace CommissioningBehavior {
 
     export class State implements CommissioningOptions {
         commissioned = false;
+        commissionedFabrics: Record<FabricIndex, ExposedFabricInformation> = {};
```
lauckhart, line 263: Is there an "uncommissionedFabric"?  Wondering if we should just call this "fabrics" since it seems to map 1:1 with known fabrics

lauckhart (approved):

packages/matter.js/src/node/ServerNode.ts
Apollon77: But still would need deep
Inside structure knowledge. Unthinkable exposing this on the node where you as dev also call "bringOnline" might be more natural?! Let's see which other Casios we find. I do not see too many more right now tbh.

packages/matter.js/src/behavior/system/commissioning/CommissioningBehavior.ts
Apollon77: Valid. I also thought about it. Should I then also rename the commissionedFabricsChanged event to just be fabricsChanged? (And before you ask I would leave the event because it transports info about the change itself (fabricindex and action). The normale state object change event can not transport this.

PR #748, opened 2024-03-16
Adjust endpoint state type to read-only

lauckhart:
Should help make it more clear that endpoint state must be modified using set().  Also added runtime error for non-TS usage.


Apollon77 (approved):
Thank you

PR #755, opened 2024-03-20
Allow also Async Storage implementations

Apollon77:
This PR adds the possibility to also have an Async storage implementation. The Sync Storages we had now are also in and both work and react a bit different depending what is used.

Typing wise it is a bit Meh because of some stones TS throw in our way, so it is not nice. We can remove some of the typing stuff when we remove the legacy API or we leave it as is for performance reasons.

The good effect of this change is that the legacy API still excusively (!) uses the SyncStorages (and only these!), but the new API can use both and will work.

The PR also includes some more small things I discovered while developing and testing:
* Enhance asyncNew to return the correct type and not any
* Enhance object checks and make them more robust
* Introduce new Storage method "contexts" to query not the keys of the current storage context level but the sub contexts. This needs to be implemented for each custom storage!
* Fix set and subscribe for Cluster client methods to return the promise of the communication and so result. This was missing so logic was running uncontrolled aside


Apollon77:
@lauckhart Here we go. Please have a look. as said the BehaviorTest is failing in any case - the matter.js testing also shows other failures but the real places is hidden. I would assume it is because invokeInitializer got async now ... but I can not really spit the issue. maybe you see it. Thank you!

Thats why I left the PR as draft for now until we fixed that. After fix I will also run tests with the new pure Async storage

packages/matter.js/src/cluster/server/ClusterServer.ts
```diff
@@ -108,7 +110,7 @@ export function ClusterServer<
             return datasource;
         },
 
-        set datasource(newDatasource: ClusterDatasource | undefined) {
+        _setDatasource<S extends Storage>(newDatasource: ClusterDatasource<S> | undefined) {
```
lauckhart, line 113: Yeah so the ClusterDatasource for the new code is specifically designed be accessed synchronously.  In ClusterBehaviorBacking.invokeInitializer I think we should just assert that it did not return a promise.  Then invokeInitializer wouldn't need to be async.

packages/matter.js/src/cluster/server/ClusterServer.ts
```diff
@@ -131,10 +133,20 @@ export function ClusterServer<
...
+                const promises = new Array<PromiseLike<void>>();
                 for (const eventName in events) {
-                    (events as any)[eventName].bindToEventHandler(datasource.eventHandler);
+                    const bindResult = (events as any)[eventName].bindToEventHandler(datasource.eventHandler);
+                    if (bindResult !== undefined && MaybePromise.is(bindResult)) {
+                        promises.push(bindResult);
+                    }
+                }
+                if (promises.length > 0) {
+                    return Promise.all(promises).then(() => Promise.resolve()) as S extends SyncStorage
```
lauckhart, line 144: Whoo boy I know why you have to cast a Promise to void but that's nasty looking.  What if you create a utility type such as:

```js
type StorageOperationResult<S extends SyncStorage | AsyncStorage, T = void> =
    S extends SyncStorage ? T : Promise<T>
```

packages/matter.js/src/environment/VariableService.ts
```diff
@@ -243,8 +244,8 @@ export function merge(a: Record<string, any>, b: Record<string, any>) {
     for (const key in b) {
         const aval = a[key];
         const bval = b[key];
-        if (typeof bval === "object") {
```
lauckhart, line 246: This is an example of where using "any" instead of "unknown" causes trouble, I'd probably fix that too.

packages/matter.js/src/fabric/FabricManager.ts
```diff
@@ -69,9 +76,16 @@ export class FabricManager {
             throw new MatterFlowError(`Fabric with index ${fabricIndex} already exists.`);
         }
         this.#fabrics.set(fabricIndex, fabric);
-        fabric.addRemoveCallback(() => this.removeFabric(fabricIndex));
+        fabric.addRemoveCallback(async () => this.removeFabric(fabricIndex));
         fabric.setPersistCallback((isUpdate = true) => {
-            this.persistFabrics();
+            const persistResult = this.persistFabrics();
+            if (MaybePromise.is(persistResult)) {
```
lauckhart, line 82: You can use MaybePromise.then() for this code pattern.  I removed it in a lot of places because it made our stacks too deep but in places like this I think it's fine to use.

packages/matter.js/src/storage/Storage.ts
```diff
@@ -12,11 +13,34 @@ export class StorageError extends MatterError {}
...
+    close(): MaybePromise<void>;
+    get(contexts: string[], key: string): MaybePromise<SupportedStorageTypes | undefined>;
+    set(contexts: string[], key: string, value: SupportedStorageTypes): MaybePromise<void>;
+    delete(contexts: string[], key: string): MaybePromise<void>;
+    keys(contexts: string[]): MaybePromise<string[]>;
+    contexts(contexts: string[]): MaybePromise<string[]>;
+    clearAll(contexts: string[]): MaybePromise<void>;
+}
+
+export abstract class MaybeAsyncStorage implements Storage {
```
lauckhart, line 26: Huh, could get rid of this type given that the base storage class is inherently "maybe async"?  So just have Storage and SyncStorage?  That feels like it'd be cleaner.

packages/matter.js/src/storage/StorageContext.ts
```diff
@@ -4,57 +4,96 @@
...
+                    if (v !== undefined) return v;
+                    if (defaultValue === undefined) {
+                        throw new StorageError(
+                            `No value found for key ${key} in context ${this.thisContexts} and no default value specified!`,
+                        );
+                    }
+                    return defaultValue;
+                }) as S extends SyncStorage ? T : Promise<T>;
+            }
+            return value as S extends SyncStorage ? T : Promise<T>;
```
lauckhart, line 31: See my comment above r.e. StorageOperationResult utility type, I think it'd be an improvement, although maybe could shorten to "StorageResult" since it'll be used more than I was thinking..

lauckhart (changes requested):

packages/matter.js/src/storage/Storage.ts
Apollon77: Thats because of https://github.com/microsoft/TypeScript/issues/57905. The "Storage" needs to be without the generic, else anything screwes up TS wise. We can kill this once it is fixed.. I will add this as a reference

packages/matter.js/src/behavior/internal/BehaviorBacking.ts
```diff
@@ -107,7 +107,7 @@ export abstract class BehaviorBacking {
      * This is an optional extension point for derivatives.  Errors thrown here are recorded and place the behavior into
      * crashed state.
      */
-    protected invokeInitializer(behavior: Behavior, options?: Behavior.Options) {
+    protected async invokeInitializer(behavior: Behavior, options?: Behavior.Options) {
```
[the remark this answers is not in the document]
Apollon77: The reason is that in ClusterServer the eventserver is initialized and potentially collected events are submitted ... each submitted event increases eventNumber ... this is async ... thats kind of the "reason".

packages/matter.js/src/behavior/state/managed/values/ValuePatcher.ts
```diff
@@ -195,10 +196,7 @@ function ListPatcher(schema: ValueModel, owner: RootSupervisor): ValueSupervisor
                 const oldValue = index < target.length ? target[index] : undefined;
                 if (newValue === undefined || newValue === null || oldValue === undefined || oldValue === null) {
                     // If creating a new object, apply as a patch to the object's defaults before insertion
-                    if (entryDefaults && typeof newValue === "object" && newValue !== null) {
-                        if (entryDefaults === undefined) {
-                            entryDefaults = getDefaults(entry);
-                        }
+                    if (entryDefaults && isObject(newValue === "object")) {
```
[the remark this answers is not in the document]
Apollon77: Ohh yes typo

packages/matter.js/src/behavior/internal/ClusterServerBacking.ts
```diff
@@ -330,7 +325,7 @@ function createEventHandler(backing: ClusterServerBehaviorBacking, name: string)
         return;
     }
 
-    observable.on((payload, _context) => {
-        eventServer.triggerEvent(payload);
+    observable.on(async (payload, _context) => {
+        await eventServer.triggerEvent(payload);
```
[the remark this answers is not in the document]
lauckhart: OK I pushed a commit that does this.  @Apollon77 I'll let you resolve this conversation once you've reviewed my commits so it doesn't automerge

lauckhart (approved):

Apollon77:
Thank you, so lets bring that in :-)

PR #757, opened 2024-03-22
Added moveToLevel command handler to LevelControlServer.ts.

Controlhaus:


Apollon77:
Thank you, very minimal but better then nothing for now :-)

Apollon77 (approved):

PR #769, opened 2024-03-27
Fixes 312

Apollon77:
The PR fixes the PASE establishment issue, replaces elliptic and BN deps by @noble/curves


packages/matter.js/src/crypto/Key.ts
```diff
@@ -518,21 +523,21 @@ export function Key(properties: Partial<Key>) {
...
 
         // Compute
-        const elliptic = new ec(crv.toLowerCase().replace(/-/g, ""));
-        const ecKey = elliptic.keyFromPrivate(that.privateKey).getPublic();
+        const ecKey = ProjectivePoint.fromPrivateKey(that.privateKey);
 
         // Install
-        that.xBits = new ByteArray(ecKey.getX().toArray());
-        that.yBits = new ByteArray(ecKey.getY().toArray());
+        that.xBits = numberToBytesBE(ecKey.x, 32); // ???
```
lauckhart, line 539: I think this is correct for p256, I'd probably create a variable and set it in the `crv` switch statement above.  Might be a field on `ecKey` you could use for this, e.g. using "k" from the hasher then `(2 * k / 128)`.  But not clear to me k is accessible just from reading the source.

lauckhart (commented):
Cool, reviewed what you have so far, LGTM

packages/matter.js/src/crypto/Key.ts
Apollon77: No, not accessible. I added a normal variable for now

lauckhart (approved):

PR #780, opened 2024-03-30
Migrate specifications to global namespace

lauckhart:
Previously @links in specification documents required a corresponding import for esdoc tooling to work.  This changes to a global "MatterSpecification" namespace.  Since it only contains types a global seems OK.  Also changes to a slightly more concise hierarchical naming convention.


packages/matter.js/src/spec/Specifications.d.ts
Apollon77, line 1: in matter-node we put such "just d.ts defs" in dir "dts"... should we do also here?

Apollon77 (approved):

packages/matter.js/src/spec/Specifications.d.ts
lauckhart: Fine by me, do you want me to move it?

Apollon77: Would be great, then we are consitent fir such ... if "dts" is not clear name then better adjust that on both places

PR #781, opened 2024-03-30
Makes Identify default custer server more convenient

Apollon77:
The PR enhances the Identify Server default implementation a bit and adds convenient states and events.

@lauckhart Should we name such additional stuff on state freely (as here) or should we previx them somehow to prevent overlapping risk with future new cluster attributes? WDYT?


lauckhart (commented):
I prefer naming freely. Risk of collision seems low and even if there is it seems likely the semantics will be the same for the new attribute or event. IIRC there are a couple of places I risked a small future breaking change to make the API cleaner

lauckhart (approved):

PR #795, opened 2024-04-03
LevelControl and other simple Cluster implementations and tests

Apollon77:
See commits, fucntionality is split up. beside some smaller fixes and tweaks the PR implements LevelControl cluster handlers and also adds chip test for all "commandless" clusters listed in the readme. For this it also adds some logic.


Apollon77:
@lauckhart if you have a better idea for the "custom validation" cases please tell me. I was not successful with using custom get/set (also because a custom setter also requires to have a custom getter and I was not able to access other state attributes in such a case

packages/matter.js/src/behavior/definitions/level-control/LevelControlServer.ts
```diff
@@ -4,14 +4,517 @@
...
+            this.internal.transitionIntervalTimer.stop();
+        }
+        await super[Symbol.asyncDispose]?.();
+    }
+}
+
+export namespace LevelControlServerLogic {
+    export class Internal {
+        /** Minimum Level value which always is defined. */
+        minLevel = 0;
```
lauckhart, line 490: Is the duplicate field just because state.minLevel could be nullish?  Could create this.minLevel getter that returns state.minLevel ?? 0.  Just to avoid the confusion of having it defined twice

packages/matter.js/src/behavior/definitions/on-off/OnOffServer.ts
```diff
@@ -42,7 +45,7 @@ export class OnOffServer extends Base {
         await super[Symbol.asyncDispose]?.();
     }
 
-    override on() {
+    override async on() {
```
lauckhart, line 48: What's the benefit of making these async?

packages/matter.js/src/behavior/internal/Reactors.ts
```diff
@@ -211,6 +211,10 @@ class ReactorBacking<T extends any[], R> {
         if (!(error instanceof Error)) {
             error = new Error(`${error}`);
         }
+        if (error instanceof MatterError) {
+            logger.error("Unhandled MatterError", this.#augmentError(error));
```
lauckhart, line 215: If we rethrow I think we shouldn't also log or the stack trace will end up in logs twice.

packages/matter.js/src/behavior/internal/Reactors.ts
```diff
@@ -211,6 +211,10 @@ class ReactorBacking<T extends any[], R> {
         if (!(error instanceof Error)) {
             error = new Error(`${error}`);
         }
+        if (error instanceof MatterError) {
```
lauckhart, line 214: I'm wondering if we should be differentiating between MatterError and other errors here.  Seems like if we catch InternalError or ImplementationError for example it should be handled the same way as a bare Error.  Should we just allow all errors through like it used to work?

packages/matter.js/src/behavior/internal/Reactors.ts
```diff
@@ -211,6 +211,10 @@ class ReactorBacking<T extends any[], R> {
...
+            logger.error("Unhandled MatterError", this.#augmentError(error));
+            throw error;
```
lauckhart, line 216: If the context is "offline" there may be nobody listening to catch the error -- so this would be unhandled and crash the process.  So maybe rethrow could have additional conditions added?

packages/matter.js/src/behavior/state/transaction/Tx.ts
```diff
@@ -506,6 +506,10 @@ class Tx implements Transaction {
                 () => participant.postCommit?.(),
                 () => postCommitNextParticipant(),
                 error => {
+                    if (error instanceof MatterError) {
```
lauckhart, line 509: See my comments in Reactors.ts, the same applies here.

lauckhart (changes requested):
Looks good, might want to tweak error handling but look at comments and LMK

packages/matter.js/src/behavior/definitions/level-control/LevelControlServer.ts
Apollon77: Yes in fact it is because of the nullish. So I need this value internally because the attribute itself can be null. I thought. Getter on "class" level will not work. I will try it. Good idea.

packages/matter.js/src/behavior/definitions/on-off/OnOffServer.ts
Apollon77: Ohh yes wanted to discuss that with you. As soon as someone implements these commands with real hardware behind it Theresia a high chance that he needs it Async. But then the inter cluster logic in default implementations would break. Also overriding the method in a "class extends DefaultServer" gives issues. But bigger issue is with inter cluster usage. Because lint tells me that await is not needed and such. 

So i started to implement all "interacts with a real device" command handlers as async... in factorization brings question if they should not all be async for above reasons .... ?! Wdyt?

packages/matter.js/src/behavior/internal/Reactors.ts
Apollon77: This I tried but then listeners registered eg in initialize of a cluster that always are offline are seen as offline rather then "executed online as change handler but the reactor specified offline so the online info is lost here - or I used the wrong flag. This#offline was not working as expected in this case.

packages/matter.js/src/behavior/definitions/on-off/OnOffServer.ts
lauckhart: Hmm taking a runtime performance hit to make tooling happy is not ideal though perhaps not significant enough to worry about.  Could we manually specify MaybePromise as the return type?  Then the fact these aren't async won't affect the types.

packages/matter.js/src/behavior/definitions/level-control/LevelControlServer.ts
Apollon77: Great, seems to work

packages/matter.js/src/behavior/definitions/on-off/OnOffServer.ts
Apollon77: you mean that way?

packages/matter.js/src/behavior/internal/Reactors.ts
Apollon77: So you mean rethrowing all errors? Ok we can do that and see if it gives issues ... I just comment out here for "reference" if we need to re-add

packages/matter.js/src/behavior/internal/Reactors.ts
Apollon77: Ok with  new approach reverted

packages/matter.js/src/behavior/definitions/on-off/OnOffServer.ts
Apollon77: Adjusted to Maxbepromise ..

packages/matter.js/src/behavior/internal/Reactors.ts
Apollon77: Reverted

packages/matter.js/src/behavior/state/transaction/Tx.ts
Apollon77: Reverted

Apollon77:
@lauckhart Ok, I was able to implement the two value checks without needing to throw in an observer, so I reverted this. did you mean it that way how it should work?

packages/matter.js/src/cluster/definitions/LocalizationConfigurationCluster.ts
```diff
@@ -34,11 +34,7 @@ export namespace LocalizationConfiguration {
              *
              * @see {@link MatterSpecification.v11.Core} § 11.3.4.1
              */
-            activeLocale: WritableAttribute(
-                0x0,
-                TlvString.bound({ maxLength: 35 }),
```
Apollon77, line 39: Thisis wrong and requires fix!!

PR #798, opened 2024-04-05
Update Events in subscriptions

### LINKED ISSUE #784: SubscriptionHandler not sending received events from bridged device added after matter server start and is connected
Hi,
this is the issue:
The controller (Apple in my case) always subscribe correctly 
```
2024-03-31 21:54:12.065 DEBUG  InteractionServer    Received subscribe request from udp://fe80::14b6:bb16:5d3a:1293%9:51251 on session secure/48170 (keepSubscriptions=true, isFabricFiltered=false)
2024-03-31 21:54:12.066 DEBUG  InteractionServer    Subscribe to attributes:*/*/*, events:!*/*/*
```
First case without issue: I create an aggregator and add a device to it then add the aggregator to the commissioning server etc etc I then start matter server and trigger the event StateChange, the event is correctly sent
```
2024-03-31 21:52:22.669 DEBUG  SubscriptionHandler  Sending subscription update message for ID 2529153003 with 1 attributes and 1 events
2024-03-31 21:52:22.669 DEBUG  MessageExchange      New exchange protocol: 1 id: 30278 session: secure/24489 peerSessionId: 2131 active threshold ms: 4000 active interval ms: 300 idle interval ms: 300 retries: 5
2024-03-31 21:52:22.669 DEBUG  SubscriptionHandler  Sending subscription changes for ID 2529153003: MA-contactsensor(0x3b)/BooleanState(0x45)/stateValue(0x0)=true (1805472651)
2024-03-31 21:52:22.670 DEBUG  InteractionMessenger Sending DataReport chunk with 1 attributes and 1 events: 85 bytes
```
ISSUE: when I create an aggregator then add the aggregator to the commissioning server etc etc and then I start matter server and only after I add a device to it and trigger the event StateChange then the event is not sent to the controller
```
2024-03-31 22:06:57.427 DEBUG  EventHandler         Received event: {"eventNumber":1824,"eventId":0,"clusterId":69,"endpointId":59,"epochTimestamp":1711915617427,"priority":1,"data":{"stateValue":false}}
2024-03-31 22:06:57.489 DEBUG  SubscriptionHandler  Sending subscription update message for ID 2231693661 with 1 attributes and 0 events
```

Basically the problem is that when a device is added to the aggregator after that matter server is started and is connected, that device is fully working but the events are not sent. If the controller takes more time to connect and the device is added after matter server start but before is connected then the events are sent correctly.

Thanks.

Apollon77:
When Endpoint structures were changed then events were not updated in the subscriptions which caused events from newly added endpoints after the initial subscription were never sent out in the subscription

fixes #784 

The change was tested locally and will be validated after merge by @Luligu

### DIFF
--- a/packages/matter.js/src/protocol/interaction/SubscriptionHandler.ts
+++ b/packages/matter.js/src/protocol/interaction/SubscriptionHandler.ts
@@ -349,9 +349,12 @@ export class SubscriptionHandler {
     updateSubscription() {
         const { newAttributes } = this.registerNewAttributes();
 
-        for (const attributeWithPath of newAttributes) {
-            const { path, attribute } = attributeWithPath;
+        for (const { path, attribute } of newAttributes) {
             const { version, value } = attribute.getWithVersion(this.session, true);
+
+            // We do not do any version filtering for attributes that are newly added to make sure controller gets
+            // most current state
+
             this.outstandingAttributeUpdates.set(attributePathToId(path), {
                 path,
                 schema: attribute.schema,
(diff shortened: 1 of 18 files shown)


packages/matter.js/src/protocol/interaction/SubscriptionHandler.ts
```diff
@@ -360,6 +363,30 @@ export class SubscriptionHandler {
...
+            .flatMap(({ path, event: { schema } }): EventPathWithEventData<any>[] => {
+                // But we use eventFilters because we do not want to send all events to the controller
+                const matchingEvents = this.eventHandler.getEvents(path, this.eventFilters);
+                return matchingEvents.map(event => ({
+                    schema,
+                    path,
+                    event,
+                }));
+            })
+            .sort((a, b) => {
```
lauckhart, line 377: I think you could write as `.sort((a, b) => (a.event?.eventNumber ?? 0) - (b.event?.eventNumber ?? 0)))`

lauckhart (approved):

packages/matter.js/src/protocol/interaction/SubscriptionHandler.ts
Apollon77: Nope ... nort expects a number as result but eventNumber normally is a Bigint ... I will fix some stuff around that mixtures here

Issue #772, opened 2024-03-28
DoorLockDevice probably not fully implemented


GreydonDesu:
I tried putting a DoorLockDevice into the Matter Device example. `npm run build-clean` and `npm run matter-device` works perfectly fine and Home Assistant can register it as usual.

If I try to change the lockState (Default at device generation is Unlock) of it with Home Assistant, this error comes from Home Assistant:
```
Failed to call service lock/lock. InteractionModelError: UnsupportedCommand (0x81)
```
Same error happens vice versa (trying to unlock a DoorLock with Default set to Lock)

The version I'm on is v0.8.0-alpha.0-20240328-f7f45db5.

Those are the logs from the device:
```logs
2024-03-28 15:44:26.697 DEBUG  MessageExchange      New exchange protocol: 1 id: 24988 session: secure/33974 peerSessionId: 33974 active threshold ms: 4000 active interval ms: 500 idle interval ms: 300 retries: 5
2024-03-28 15:44:26.698 DEBUG  MessageExchange      Message « id: 33974/24988/159970605 type: 1/10 acked: undefined reqAck: true duplicate: false payload: 152500e80324ff0b18
2024-03-28 15:44:26.698 DEBUG  InteractionServer    Received timed request (1000ms) from udp://fe80::6ff6:cce4:9f1a:ff3d%11:49715 on session secure/33974
2024-03-28 15:44:26.698 DEBUG  MessageExchange      Starting timed interaction with Transaction ID 24988 for 1000ms from udp://fe80::6ff6:cce4:9f1a:ff3d%11:49715 on session secure/33974
2024-03-28 15:44:26.699 DEBUG  ExchangeManager      Message » id: 33974/24988/215677691 type: 1/1 acked: 159970605 reqAck: true duplicate: false payload: 1524000024ff0b18
2024-03-28 15:44:26.702 DEBUG  MessageExchange      Message « id: 33974/24988/159970606 type: 1/8 acked: 215677691 reqAck: true duplicate: false payload: 152800290136021537002400012501010124020118350118181824ff0b18
2024-03-28 15:44:26.702 DEBUG  InteractionServer    Received invoke request from udp://fe80::6ff6:cce4:9f1a:ff3d%11:49715 on session secure/33974: DoorLock(0x1)/DoorLock(0x101)/unknown(0x1), suppressResponse=false
[...]

PR #804, opened 2024-04-08
Add minimal DoorLock capabilities

GreydonDesu:
As known in #772 this is the minimal cluster for a DoorLockDevice.

Tested the device with Home Assistant and it does not throw an error.

### DIFF
--- a/packages/matter.js/src/behavior/definitions/door-lock/DoorLockServer.ts
+++ b/packages/matter.js/src/behavior/definitions/door-lock/DoorLockServer.ts
@@ -4,11 +4,19 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-/*** THIS FILE WILL BE REGENERATED IF YOU DO NOT REMOVE THIS MESSAGE ***/
-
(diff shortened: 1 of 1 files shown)


packages/matter.js/src/behavior/definitions/door-lock/DoorLockServer.ts
```diff
@@ -4,11 +4,19 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-/*** THIS FILE WILL BE REGENERATED IF YOU DO NOT REMOVE THIS MESSAGE ***/
-
+import { DoorLock } from "../../../cluster/definitions/index.js";
```
Apollon77, line 7: Please import the specific file and not "all clusters" via index.

```suggestion
import { DoorLock } from "../../../cluster/definitions/DoorLockCluster.js";
```

Apollon77 (changes requested):
Other then that cool, thank you

Apollon77 (approved):

Apollon77:
Thank you!

PR #823, opened 2024-04-18
Enhanced migration docs for composed bridged devices

Apollon77:


docs/MIGRATION_GUIDE_08.md
```diff
@@ -188,6 +188,47 @@ const endpoint = await serverNode.add(
...
+await composed.add(composedWC);
+await composed.add(composedLight);
+```
+
+For hints on options to add endpoints into a structure see next paragraph.
+
+### How to add endpoints into a structure?
+The new API provides basically two ways to add endpoints into a structure:
+
+* "Lazy" way: When using `endpoint.parts.add()` or when you provide the parts in the configuration of the Endpoint when creating it. While this is convenient it has, as of current 0.8.x versions, the disadvantage that the Endpoint is not directly initialized, and so you can not directly add event handlers or such to the endpoint. This is only done when the node is online or an endpoint was added with full initialization (see next point). is planned to be changed in future versions.
```
lauckhart, line 229: Events will change but the endpoint still won't be fully initialized until added to an initialized patent so state is available

lauckhart (approved):

docs/MIGRATION_GUIDE_08.md
Apollon77: Ok  I add state, and event we do remove when we do it

PR #827, opened 2024-04-18
Generate Endpoints by default too

Apollon77:
@lauckhart that costest me some time to notice it was not executed :-)


lauckhart (approved):
I do my best to keep life exciting

PR #838, opened 2024-04-25
Add another workaround for TS 27965

lauckhart:
Due to TS bug with mapped types (the same one that forced "ClusterInterface" on us) we need to manually designate methods that should be carried forward for altered behaviors.  Otherwise TS considers the methods function properties and won't allow for override with a method (though it's still easy to override with slightly uglier property syntax).

### DIFF
--- a/packages/matter.js/src/behavior/cluster/ClusterBehavior.ts
+++ b/packages/matter.js/src/behavior/cluster/ClusterBehavior.ts
@@ -15,7 +15,7 @@ import type { BehaviorBacking } from "../internal/BehaviorBacking.js";
 import type { RootSupervisor } from "../supervision/RootSupervisor.js";
 import { Schema } from "../supervision/Schema.js";
 import { NetworkBehavior } from "../system/network/NetworkBehavior.js";
-import { createType, type ClusterOf } from "./ClusterBehaviorUtil.js";
+import { ExtensionInterfaceOf, createType, type ClusterOf } from "./ClusterBehaviorUtil.js";
 import type { ClusterEvents } from "./ClusterEvents.js";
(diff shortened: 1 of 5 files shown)


packages/matter.js/src/behavior/cluster/ClusterBehavior.ts
```diff
@@ -170,6 +170,8 @@ export class ClusterBehavior extends Behavior {
 }
 
 export namespace ClusterBehavior {
+    export const foo = 4;
```
Apollon77, line 173: whats that? Leftover from tests or by reason?

packages/matter.js/test/behavior/cluster/ClusterBehaviorTest.ts
```diff
@@ -29,12 +29,29 @@ import { My, MyBehavior, MyCluster } from "./cluster-behavior-test-util.js";
...
+        return true;
+    }
+
+    bar() {
+        return 4;
+    }
+}
+
+namespace BehaviorWithCustomMethods {
+    export declare const ExtensionInterface: {
```
Apollon77, line 50: Can this also being extended?

So BehaviorWithCustomMethods adds bar() ..

Now I extend all this and have BehaviorWithEvenMoreCustomMethods which adds fun() ... ?

lauckhart: Yes you'd have to use an intersection type.  Hopefully wouldn't happen very much:

```typescript
namespace BehaviorWithEvenMoreCustomMethods {
    export declare const ExtensionInterface: BehaviorWithCustomMethods & {
        fun(): boolean;
    }
}
```

Apollon77 (approved):

PR #842, opened 2024-04-26
Implement new "fieldName$Changing" event

lauckhart:
- Transaction commit now supports a "pre-commit" phase that executes before phase-1 commit
- Adds attribute$Changing events to Behaviors that emit in transaction pre-commit
- $Changing events allow for state mutation and will cycle for a limited number of times until state is stable
- Datasource.reference now always returns the same managed value for a supervisor session
- Observable.isObserved allows for emitter optimization if an Observable has no observers
- Adds isObserved to ObservableProxy and moves into Observable.ts
- Fixes the floating promise in FailsafeTimer; it tended to kill a test run without an easy way to identify the cause
- Adds "deep copy" functionality for cloning an object.  Supports circular references and data views
- Includes a bunch of new tests for $Changing events


Apollon77 (approved):
Wow, awesome work! Thank you very much!

PR #856, opened 2024-04-29
Datasource tests and bitmap value management

lauckhart:
Adds various permutations of tests for Datasource eventing.  Includes a few WindowCovering-specific tests.


packages/matter.js/test/behavior/definitions/window-covering/WindowCoveringServerTest.ts
```diff
@@ -0,0 +1,98 @@
...
+                oldValue: { lift: 1, tilt: 1, global: 2 },
+                newValue: { global: 1, lift: 1, tilt: 1 },
+            },
+            {
+                name: "operationalStatus$Changed",
+                oldValue: { lift: 2, tilt: 2, global: 2 },
+                newValue: { lift: 1, tilt: 1, global: 1 },
+            },
+            {
+                name: "operationalStatus$Changed",
```
Apollon77, line 82: Any idea why the orderhere is that way around? Changed first and afterwards the Changing?

Apollon77 (approved):

packages/matter.js/test/behavior/definitions/window-covering/WindowCoveringServerTest.ts
lauckhart: Off the top of my head the best idea I have is "WTF?"

lauckhart: Was a bug in ObservableProxy, good catch!

PR #860, opened 2024-05-01
Switch cluster

Apollon77:
This PR implements the switch cluster as by Matter specification ... Chip tool was not that helpful this time because only Latching switches are implemented there, so also here are some smaller questions open (e.g. when a device can max report 3 multipresses ... what happens when the fourth one happens? ;-))), but they should be ok for now.

I added extensive Unit tests for that cluster in matter,js because the chip tests do/can not really test things that well


packages/matter.js/src/behavior/definitions/switch/SwitchServer.ts
```diff
@@ -4,14 +4,280 @@
...
+        );
+    }
+
+    override initialize() {
+        this.state.rawPosition = this.state.currentPosition;
+        this.internal.currentUnstablePosition = this.state.currentPosition;
+        this.internal.previouslyReportedPosition = this.state.currentPosition;
+
+        // Validate positions set to respect constraints, TODO: Remove with 1.3 update
+        this.reactTo(this.events.rawPosition$Changing, this.#assertPositionInRange, { offline: true });
```
lauckhart, line 66: I didn't see any tests that cover these, did you test manually?  The "offline" flag I think would make it run out-of-band so any errors these throw would be logged but wouldn't prevent the write.

See `ReactorBacking.#react` in Reactors.ts...  If "offline" it ignores the parent context.  Then when there's no parent context it captures any errors and passes them to `#unhandledError` which logs them.

lauckhart (approved):

packages/matter.js/src/behavior/definitions/switch/SwitchServer.ts
Apollon77: I add tests. And yes I can remove offline Here I think.

PR #876, opened 2024-05-10
Log params with objects correctly for invokes

Apollon77:
When invoke data contained sub objects these were not correctly logged, e.g.

> Invoke binford-6100.onoff1.colorControl.moveToColorTemperature online#9065e96@1b669 colorTemperatureMireds: 32639 transitionTime: 0 optionsMask: [object Object] optionsOverride: [object Object]

With this change it loggs it as

> Invoke binford-6100.onoff1.colorControl.moveToColor online#5d3525a@1b669 colorX: 33000 colorY: 20000 transitionTime: 0 optionsMask: {"executeIfOff":false} optionsOverride: {"executeIfOff":false}

... still not perfect but better. I decided for now to not enhance the whole logging to support such sub structs (or did I miss anything?)

@lauckhart  if you have a better idea just shoot

### DIFF
--- a/packages/matter.js/src/log/Format.ts
+++ b/packages/matter.js/src/log/Format.ts
@@ -9,6 +9,7 @@ import { ImplementationError, InternalError } from "../common/MatterError.js";
 import { ByteArray } from "../util/ByteArray.js";
 import { Diagnostic } from "./Diagnostic.js";
 import { Level } from "./Level.js";
+import { Logger } from "./Logger.js";
 
 /**
  * Get a formatter for the specified format.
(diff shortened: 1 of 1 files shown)


packages/matter.js/src/log/Format.ts
```diff
@@ -419,6 +420,9 @@ function renderValue(value: unknown, formatter: Formatter, squash: boolean): str
             })
             .join(squash ? "" : " ");
     }
+    if (typeof value === "object") {
+        return formatter.text(Logger.toJSON(value));
```
lauckhart, line 424: Maybe use serialize() here instead of toJSON?  Serialize is a little more readable and won't crash if there are circular references in the object...  Also maybe special case Date just before this so dates don't get quoted.  Other than that seems fine

lauckhart (changes requested):

lauckhart (approved):

PR #877, opened 2024-05-10
Enhance General Diagnostics Cluster

### LINKED ISSUE #873: GeneralDiagnosticsCluster NetworkInterfaces attribute in CommissioningServer
Hi,
testing Matterbridge with Home Assistant I noticed in the HA logs that they report like errors the lack of NetworkInterfaces attribute in the GeneralDiagnosticsCluster of the CommissioningServer.
I added it in Matterbridge and now the HA logs are clear and the commissioning now takes less then a second.
I think that, even if it is not mandatory, would be useful to add the NetworkInterfaces attribute directly in matter.js.
Probably also other controllers may use it for diagnostic or pairing process (also Apple Home now commission Matterbridge immediatly).
For the same reason (reported errors in HA logs) I also added productId: OptionalFixedAttribute, to BridgedDeviceBasicInformation and all errors disappeared from logs.
It would be useful to add it as well like optional fixed attribute.
Thanks.

Apollon77:
This PR enhances the General Diagnostics cluster
* Add upTime counter (new and legagy API)
* Automatically fill network Interfaces attribute (new and legagy API), as it seems that HA uses it internally, fixes #873 
* Add operationalHour counter and persistence (new API only)
* Add convenience methods to register and clear faults and automatically update the relevant attributes and send events if enabled
* some needed enhancements


packages/matter.js/src/behavior/definitions/operational-credentials/OperationalCredentialsServer.ts
Apollon77, line 1: Please leave this comment open even after review. I need to add Changelog :-)

packages/matter.js/src/behavior/definitions/general-diagnostics/GeneralDiagnosticsServer.ts
```diff
@@ -39,10 +73,306 @@ export class GeneralDiagnosticsServer extends GeneralDiagnosticsBehavior {
...
 
+    /**
+     * Register a hardware fault. This convenience method updates the activeHardwareFaults attribute and sends out the
+     * HardwareFaultChanged event. Make sure to clear the fault when it is resolved.
+     * This method requires that the activeHardwareFaults attribute is activated when the cluster gets initialized.
+     *
+     * @param faultType The hardware fault to register.
+     */
+    registerHardwareFault(faultType: GeneralDiagnostics.HardwareFault) {
+        if (this.state.activeHardwareFaults === undefined) {
```
lauckhart, line 84: This test happens 6x in this file alone.  Maybe add an "assertAttributeEnabled" or "requireAttribute" utility method to ClusterBehavior that returns the attribute value or throws?  The message could be "To use this feature, please enable attribute ${attributeName} by setting the value during initialization"

packages/matter.js/src/behavior/system/network/ServerNetworkRuntime.ts
```diff
@@ -25,6 +25,13 @@ import { CommissioningBehavior } from "../commissioning/CommissioningBehavior.js
 import { SessionsBehavior } from "../sessions/SessionsBehavior.js";
 import { NetworkRuntime } from "./NetworkRuntime.js";
 
+type NetworkInterfaceDetails = {
```
lauckhart, line 28: WDYT about adding an optional "interface type" field here and converting the return value from Network.getNetInterfaces() to a list of objects that also includes the interface type?

If we then leave a TODO in NetworkNode it would allow users to override NetworkNode to add their own implementation (or submit a PR).

Might be worthwhile to provide a bare-bones linux-only implementation that looks for /sys/class/net/(interface name)/wireless to differentiate between ethernet and wifi.

lauckhart (approved):

packages/matter.js/src/behavior/system/network/ServerNetworkRuntime.ts
Apollon77: What do you think abot that idea ...

packages/matter.js/src/net/Network.ts
```diff
@@ -9,14 +9,60 @@ import { UdpChannel, UdpChannelOptions } from "./UdpChannel.js";
...
+
+export type NetworkInterfaceDetails = {
+    mac: string;
+    ipV4: string[];
+    ipV6: string[];
+};
+
+export type NetworkInterfaceDetailed = NetworkInterface & NetworkInterfaceDetails;
+
+export type NetworkInterfaceNameToTypeMapping = { name: string; type: InterfaceType }[];
```
lauckhart, line 56: Rather than:

```
 MATTER_NETWORK_INTERFACENAMETOTYPEMAPPING_ETH0=1
```

Could we do:

```
MATTER_NETWORK_INTERFACE_ETH0_TYPE=ethernet
```

Thinking being if we ever want to add additional per-interface config properties then there is room to add here without creating an additional mapping object.

Also in general I've tried to use strings rather than numbers for config values just to make things less cryptic in contexts where the enum definition isn't available (e.g. env var/config file/command line).

lauckhart (approved):

packages/matter.js/src/behavior/system/network/ServerNetworkRuntime.ts
lauckhart: Lgtm

packages/matter.js/src/net/Network.ts
Apollon77: Ok cool. Did not knew the parser can do that fancy stuff. I was more thinking about this is set programmatically. Is that then array with {type:"..."} ??

lauckhart: If you did `vars.get("network.interface")` it should give you the config subtree `{ eth0: { type: "ethernet"}}`

So it'd be a `Record<string, { type: string }>` though of course the type would need validation.  I figure at some point we could validate config with Matter schema 🙂

Apollon77: Will verify

Apollon77: @lauckhart how about this?

PR #880, opened 2024-05-12
Adding On demand Certificate determination

Apollon77:
To allow cost effectiveness for devices build with matter.js this PR adds an on demain option to determine device certificates (for legacy API and also new API). When setting up the certificates for the node you can either provide the certificates as before or you provide an async provider method instead.

This method is then called as soon as certificates are required the first time in the commissioning flow. With this it is possible to dynamically generate the certificates when really needed. Becauseof the fact that while commissioning happens (and failsafe timer is active) the default command timeout is 30s there should be enough time to even request the certificates from remote at exactly that timepoint.

This PR adds this feature to the new API and also to the Legacy API. For the Legacy API this PR contains a breaking change by syncing the "certificate" data properties to be in synv with the DeviceCertification types, which basically means that the name of the properties changes.

### DIFF
--- a/packages/matter.js/src/behavior/definitions/operational-credentials/DeviceCertification.ts
+++ b/packages/matter.js/src/behavior/definitions/operational-credentials/DeviceCertification.ts
@@ -11,6 +11,7 @@ import { ImplementationError } from "../../../common/MatterError.js";
 import { Crypto } from "../../../crypto/Crypto.js";
 import { PrivateKey } from "../../../crypto/Key.js";
 import { SecureSession } from "../../../session/SecureSession.js";
+import { AsyncConstruction } from "../../../util/AsyncConstruction.js";
 import { ByteArray } from "../../../util/ByteArray.js";
 import { ProductDescription } from "../../system/product-description/ProductDescription.js";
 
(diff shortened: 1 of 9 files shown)


packages/matter.js/src/behavior/definitions/operational-credentials/DeviceCertification.ts
```diff
@@ -69,4 +117,6 @@ export namespace DeviceCertification {
         intermediateCertificate: ByteArray;
         declaration: ByteArray;
     }
+
+    export type ProviderFunction = () => Promise<DeviceCertification.Configuration>;
```
lauckhart, line 121: Seems like the type we mostly use is config | provider so maybe could define as e.g.:

```js
    export type Definition = Configuration | (() => Promise<Configuration>)
```

...then the union would only appear once

packages/matter.js/src/behavior/definitions/operational-credentials/OperationalCredentialsServer.ts
```diff
@@ -423,6 +430,12 @@ export class OperationalCredentialsServer extends OperationalCredentialsBehavior
         await this.context.transaction.commit();
     }
 
+    async #assureCertification() {
```
lauckhart, line 433: Given that every call to #certification is prefaced by #assureCertification seems simpler if you combine #assureCertification and #certification into one method. That'd also make it impossible to introduce logic errors in the future by omitting #assureCertification

lauckhart (approved):

packages/matter.js/src/behavior/definitions/operational-credentials/OperationalCredentialsServer.ts
Apollon77: #Sometimes there is multiple accesses to #certification after the assure call and I would then need to make these all Async. I decided for that way because so it is one please which ensure it using await so the other access can be sync.

Apollon77: Thx. I will sleep over it ;-)

PR #892, opened 2024-05-15
Implement AccessControl, Minimal Group Management and a lot of fixes

Apollon77:
This PR enhances matter.js with the following topics:
* Add AccessControl logic, cluster and wire in correct handling in all places where needed for new and Legacy API
* Adds the default ACL entry in a backward compatible way which was not set in the past
* It adjusts Write Request handling to process chunked lists as testing (and project-chip implementation wise) requires it - see also https://github.com/project-chip/connectedhomeip/issues/33487
* Adds vaildation for MEIs
* Adds a Minimal Group Key Management cluster which only supports the IPK (so no groups) but provides all relevant methods for Legacy and new API
* Addresses several isues and enhancements discovered while implementing all this
* Adds chip testing for ACE, ACL and some IDM tests
The PR can be reviewsd on a per-commit basis

Note: Especially for correct wiring on new API the PR is doing some short cuts that should be handled afterwards and adjusted


packages/matter.js/src/fabric/Fabric.ts
```diff
@@ -267,6 +278,66 @@ export class Fabric {
         return Array.from(this.#scopedClusterData.get(cluster.id).keys());
     }
 
+    private getAsGroupSet(
+        operationalGroupSet: OperationalGroupKeySet,
+    ): TypeFromSchema<typeof GroupKeyManagement.TlvGroupKeySetStruct> {
```
lauckhart, line 283: This seems like an identity operation?

packages/matter.js/src/node/server/TransactionalInteractionServer.ts
```diff
@@ -119,18 +153,71 @@ export class TransactionalInteractionServer extends InteractionServer {
...
+    }
+
+    override async handleWriteRequest(
+        exchange: MessageExchange<MatterDevice>,
+        writeRequest: WriteRequest,
+        message: Message,
+    ): Promise<WriteResponse> {
+        // TODO: This is a hack to prevent the ACL from updating while we are in the middle of a write transaction and will
+        //  be removed again once we somehow handle relevant sub transactions
+        this.aclServer.aclUpdateDelayed = true;
```
lauckhart, line 199: This doesn't seem too bad given it's only necessary in very limited circumstances

packages/matter.js/src/node/server/TransactionalInteractionServer.ts
```diff
@@ -119,18 +153,71 @@ export class TransactionalInteractionServer extends InteractionServer {
...
+        path: EventPath,
+        eventFilters: TypeFromSchema<typeof TlvEventFilter>[] | undefined,
+        event: EventServer<any, any>,
+        exchange: MessageExchange<MatterDevice>,
+        fabricFiltered: boolean,
+        message: Message,
+        endpoint: EndpointInterface,
+    ): Promise<EventStorageData<any>[]> {
+        const readEvent = (context: ActionContext) => {
+            // TODO Change this to behavior style, right now it does minimalistic whats needed
```
lauckhart, line 171: What would "behavior style" look like?  Extend behavior API to support reading historical events, then doing ACL check there?

packages/matter.js/src/node/server/TransactionalInteractionServer.ts
```diff
@@ -139,18 +226,29 @@ export class TransactionalInteractionServer extends InteractionServer {
...
         command: CommandServer<any, any>,
         exchange: MessageExchange<MatterDevice>,
         commandFields: any,
         message: Message,
         endpoint: EndpointInterface,
         timed = false,
     ) {
-        const invokeCommand = () => super.invokeCommand(command, exchange, commandFields, message, endpoint, timed);
+        const invokeCommand = (context: ActionContext) => {
+            // TODO Change this to behavior style, right now it does minimalistic whats needed
```
lauckhart, line 244: I was thinking ACL enforcement for commands would live here, we could add a method to check ACLs on the behavior but I don't think I'd wire directly into the command handlers because that would be problematic with overrides.

lauckhart (approved):
Looks good to me, comments are minor

packages/matter.js/src/behavior/supervision/RootSupervisor.ts
```diff
@@ -110,7 +110,11 @@ export class RootSupervisor implements ValueSupervisor {
         if (!persistent) {
             persistent = new Set();
             for (const member of this.#members) {
-                if (member.effectiveQuality.nonvolatile) {
+                if (
+                    member.effectiveQuality.nonvolatile ||
+                    member.effectiveAccess.writable ||
+                    member.effectiveAccess.fabricScoped
```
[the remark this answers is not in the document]
Apollon77: Spec tells that non volatile needs to be persisted but semantically other attributes could also be. So I decided back then that a writable attribute (mostly config stuff or behavior changing) should also be persisted because else a written change is gone on next reboot which makes no sense. Ok fabric scoped is sameish but here writable is mostly there too. So it is "interpreting specs topic". I can later post official spec phrase for persisting for that too.

packages/matter.js/src/fabric/Fabric.ts
Apollon77: What do you mean?

packages/matter.js/src/node/server/TransactionalInteractionServer.ts
Apollon77: Problem is that also other will need that. Like groups cluster. Also here you have a list where we should process the final changes once at the end. Because here you start connecting to other nodes (or disconnecting). I think we need a more generic solution for that.

packages/matter.js/src/node/server/TransactionalInteractionServer.ts
Apollon77: Good question. If you are ok with thatcher all good. ;-)

packages/matter.js/src/node/server/TransactionalInteractionServer.ts
Apollon77: I just thought you planned more because there are methods like autorizeInvoke and such when i remember. We can also do like that  and good.

packages/matter.js/src/behavior/context/server/OnlineContext.ts
```diff
@@ -90,11 +93,31 @@ export function OnlineContext(options: OnlineContext.Options) {
                     transaction,
                     trace,
 
-                    accessLevelFor(_location?: AccessControl.Location) {
-                        // TODO - use AccessControlServer on the RootNodeEndpoint
-                        //const accessControl = accessContext.behavior.node.get(AccessControlServer);
-                        //return accessControl.accessLevelFor((context as ActionContext), accessContext);
-                        return AccessLevel.Administer;
+                    hasAccessFor(desiredAccessLevel: AccessLevel, location?: AccessControl.Location) {
```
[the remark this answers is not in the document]
Apollon77: adjusted

models/src/local/GroupKeyManagementOverrides.ts
```diff
@@ -11,6 +11,14 @@ LocalMatter.children.push({
     name: "GroupKeyManagement",
 
     children: [
+        // Conformance of this field in the struct is "provisional, mandatory" which in fact means "optional"
```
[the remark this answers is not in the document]
Apollon77: added TODO also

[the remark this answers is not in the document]
Apollon77: https://github.com/orgs/project-chip/projects/11/views/1?pane=issue&itemId=64811050

packages/matter.js/src/behavior/supervision/RootSupervisor.ts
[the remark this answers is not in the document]
Apollon77: https://github.com/orgs/project-chip/projects/11/views/1?pane=issue&itemId=64811050 and added TODO

packages/matter.js/src/fabric/Fabric.ts
Apollon77: Ok, I changed it. I added a namespace function OperationalGroupKeySet.asTlvGroupSet ... i think so better? But correct, wrong as method in Fabric

packages/matter.js/src/node/server/TransactionalInteractionServer.ts
Apollon77: I added a TODO

lauckhart: Yes will need for groups too but I think groups & acls is it?  But can add subtransactions although I don't quite understand your first comment, maybe you could elaborate on the "why" in the project task.

PR #902, opened 2024-05-23
Refactor Channel management

Apollon77:
While checking for chip subscription test 3.6 it became clear that the current channel management was not correct according to the specification.

This change is adjusting this to correctly handle multiple channels per Fabric and not just one. We also wire this with the capability Minimal value about concurrent CASE sessions per fabric.

Internally we still use Fabric AND Node, so in fact we have higher number of CASe sessions. The reason is that we need to validate the effect of limiting it when using controller. I left a TODO for this and add one to the project.


packages/matter-node.js/src/storage/StorageBackendDisk.ts
```diff
@@ -9,19 +9,24 @@ import { LocalStorage } from "node-localstorage";
 
 export class StorageBackendDisk extends SyncStorage {
     private readonly localStorage;
+    protected _initialized = false;
```
lauckhart, line 12: Is "_" convention for protected variables?

packages/matter.js/src/behavior/definitions/operational-credentials/OperationalCredentialsServer.ts
```diff
@@ -366,7 +366,13 @@ export class OperationalCredentialsServer extends OperationalCredentialsBehavior
             };
         }
 
-        await fabric.remove(this.session.id);
+        // That's a workaround to make sure that factory reset works correctly in case of removal of the last fabric
+        // TODO find better solution
```
lauckhart, line 370: Looks like a bug I should look at?

lauckhart (dismissed):

packages/matter.js/src/behavior/definitions/operational-credentials/OperationalCredentialsServer.ts
Apollon77: Yes, I add a todo in the project

Apollon77: Ha, I found the solution :-) check the newest commits please ... I think thats it.

The factory was started with the fabric removal, but sessions are expired afterwards and might need some round trips to finish with ack and such. so now the factory reset is triggered when the last session gets removed after fabric removal.

I also adjusted the logic that we do not announce the device as commissionable on last session closure to just expire that again because of factory reset. Have a look please.

packages/matter-node.js/src/storage/StorageBackendDisk.ts
Apollon77: *cough* ... suuuure ;-)

changed

PR #905, opened 2024-05-26
Matter 1.3

lauckhart:
Adds support for Matter 1.2 and Matter 1.3 clusters and device types.  Includes a variety of additional improvements that made sense along the way.

More detail available in individual commit messages.

Still TODO:
- ~~Fix up CHIP tests (don't have these running locally yet but I assume some will fail)~~
- Extended constraint dialect.  There are a few constraints we don't parse yet...  Not a new issue but they're used a bit more now
- Test examples and IRL


packages/matter.js/src/behavior/definitions/bridged-device-basic-information/BridgedDeviceBasicInformationServer.ts
```diff
@@ -51,7 +51,7 @@ export class BridgedDeviceBasicInformationServer extends BridgedDeviceBasicInfor
         }
 
         // Assert owner is an aggregator
-        if (!owner.get(DescriptorServer).hasDeviceType(AggregatorEndpoint.deviceType)) {
+        if (!owner.get(DescriptorServer).hasDeviceType(AggregatorDevice.deviceType)) {
```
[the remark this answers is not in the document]
lauckhart: I was actually thinking I'd special case this so it goes back to its old location even though they classify it as "application" in the docs now. WDYT?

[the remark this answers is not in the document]
Apollon77: This or an alias COuld also work, then it stays backward compat. Also ok for me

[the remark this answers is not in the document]
lauckhart: I reverted to old name, I already had a special case in place for BridgedNode

packages/matter.js/src/behavior/definitions/window-covering/WindowCoveringServer.ts
```diff
@@ -574,8 +574,11 @@ export class WindowCoveringServerLogic extends WindowCoveringServerBase {
         this.#assertMotionLockStatus();
 
         if (this.features.positionAwareLift) {
-            this.state.targetPositionLiftPercent100ths = liftPercent100thsValue;
-            this.#triggerLiftMotion(MovementDirection.DefinedByPosition, this.state.targetPositionLiftPercent100ths);
+            this.state.targetPositionLiftPercent100ths = liftPercent100thsValue ?? null;
```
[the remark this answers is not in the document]
lauckhart: OK I marked the non-100ths field as deprecated

packages/matter.js/src/device/DeviceTypes.ts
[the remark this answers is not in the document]
lauckhart: I marked the DeviceTypes constant as deprecated, but do you think I should just convert the legacy code to use the new EndpointType definitions?

packages/matter.js/src/cluster/definitions/BasicInformationCluster.ts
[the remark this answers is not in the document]
lauckhart: It was an override, I took it out

models/src/local/subject-id.ts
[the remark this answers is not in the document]
lauckhart: Note this isn't new, it's just broken out from Globals.ts.  It represents semantic Matter "subject ID" so is different from our current TS SubjectID...  I assume we'll change that to support GroupID or CAT at some point.

packages/matter.js/src/device/DeviceTypes.ts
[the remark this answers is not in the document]
Apollon77: Whenthey are type compatible then this could work ... and then still add then to AllClusters so that this struct is backward compatible existsing

chip-testing/src/BridgeTestInstance.ts
```diff
@@ -22,6 +22,13 @@ import { Storage } from "@project-chip/matter.js/storage";
 import { ByteArray } from "@project-chip/matter.js/util";
 import { TestInstance } from "./GenericTestApp.js";
 
+// Note - inevitable future confusion
+//
+// See https://github.com/project-chip/connectedhomeip/issues/30467
+//
+// Bug in CHIP tests, should be fixed for 1.4 but not available here for reasons.  Meanwhile give them what they expect
+Object.assign(BridgedDeviceBasicInformation.Cluster, { revision: 2 });
```
Apollon77, line 30: We should remove this, the *1_1 tessts are gone, so no need to anymore

chip-testing/src/BridgeTestInstanceLegacy.ts
```diff
@@ -19,6 +19,13 @@ import { Aggregator, DimmableLightDevice } from "@project-chip/matter.js/device"
 import { ByteArray } from "@project-chip/matter.js/util";
 import { AllClustersTestInstanceLegacy } from "./AllClustersTestInstanceLegacy.js";
 
+// Note - inevitable future confusion
+//
+// See https://github.com/project-chip/connectedhomeip/issues/30467
+//
+// Bug in CHIP tests, should be fixed for 1.4 but not available here for reasons.  Meanwhile give them what they expect
+Object.assign(BridgedDeviceBasicInformation.Cluster, { revision: 2 });
```
Apollon77, line 27: same here

packages/matter.js/package.json
```diff
@@ -86,6 +86,26 @@
                 "default": "./dist/cjs/endpoint/definitions/device/*.js"
             }
         },
+        "./elements": {
```
Apollon77, line 89: Please add to matter.js exports list in package README

packages/matter.js/src/CommissioningServer.ts
```diff
@@ -274,24 +267,27 @@ export class CommissioningServer extends MatterNode {
...
+            localConfigDisabled: false,
+            softwareVersion: 1,
+            softwareVersionString: "v1",
+            capabilityMinima: {
+                caseSessionsPerFabric: 3, // TODO get that limit from Sessionmanager or such or sync with it, add limit? Just a minima?
+                subscriptionsPerFabric: 3, // TODO get that limit from Interactionserver? Respect it? It is just a minima?
             },
-            options.basicInformation,
-        ) as AttributeInitialValues<typeof BasicInformationCluster.attributes>;
+            serialNumber: `node-matter-${Crypto.get().getRandomData(4).toHex()}`,
```
Apollon77, line 283: officially out of scope ... but should we change that to matter.js :-))

packages/matter.js/src/parser/Lexer.ts
Apollon77, line 1: /parser really on the root level of package and not in model?

Apollon77 (commented):
So, last small stuff and please add changelog infos ... I will release 0.9.1 now with ACL then we have a clean base to merge this and also add all changelog stuff.. question here mainly if we just say "1.3" or if you like to add more infgos on nall the awesome changes you did in parser, codegen and such

packages/matter.js/src/parser/Lexer.ts
lauckhart: Yes, nothing parser specific and it'll have considerably more code that many of our src-level directories.  I only included the lexer for this PR...  Needed fixes I'd made in the Lexer & the parser isn't done.

packages/matter.js/src/CommissioningServer.ts
lauckhart: will do

chip-testing/src/BridgeTestInstance.ts
lauckhart: Oops, I actually did but I think I did a hard reset due to other changes before I committed

packages/matter.js/src/parser/Lexer.ts
Apollon77: Alright, we can clean up when we split packages

Apollon77 (approved):

PR #925, opened 2024-06-03
MessageExchange and Session Adjustments/Optimizations for Mattert 1.3

Apollon77:
This PR adjusts and optimizes some things around MessageExchange and Sessions regarding Matter (1.3) specs
* Sync SecureChannel messages with specs (even currently unused ones that will be implemented soon)
* Evicting sessions should use the Session Timestamp and not Active Timestamp as per specs
* Optimize Message handling and drop invalid messages earlier
* Catch Message handling errors ourself instead of letting them throw in dgram/net handlers
* Dynamically calculate the session close timer
* Sync session intervals with Specs and fix bugs where variables were switched
* Add new Matter 1.3 Session Params values including fallback values and default values when we initiate sessions
* Cleanups

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,7 +28,9 @@ The main work (all changes without a GitHub username in brackets in the below li
     -   Feature: Adds all elements (clusters, attributes, events, commands, device types and datatypes) introduced in Matter 1.2 and Matter 1.3.
 -   Matter-Core functionality:
     -   Feature: Increase Data Model revision to 17 (introduced by Matter 1.2)
+    -   Enhancement: Update Session parameters in PASE/CASE to match Matter 1.3 specification
     -   Enhancement: Removes TCP and ICD TXT records from MDNS responses because both currently not supported and optional to reduce the size of the MDNS responses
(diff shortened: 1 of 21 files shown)


CHANGELOG.md
```diff
@@ -28,7 +28,9 @@ The main work (all changes without a GitHub username in brackets in the below li
     -   Feature: Adds all elements (clusters, attributes, events, commands, device types and datatypes) introduced in Matter 1.2 and Matter 1.3.
 -   Matter-Core functionality:
     -   Feature: Increase Data Model revision to 17 (introduced by Matter 1.2)
+    -   Enhancement: Update Session parameters in PASE/CASE to match Matter 1.3 specification
```
Apollon77, line 31: dummy .. Please leave this comment open ... Iwant to merge this PR and not squash it if we need to cherry pick stuff into a 0.9 branch or such :-)

lauckhart: Hmm yes it would be nice if our PRs wouldn't squash by default, kind of throws away useful information

lauckhart (approved):

PR #953, opened 2024-06-13
Handle constraints with 2**62 bound

### LINKED ISSUE #950: Spec scraping issue with "powered numbers" like 2^62
Example: Little bug in MeasurementAccuracyRange.ts generated file

Where:
rangeMin: TlvField(0, TlvInt64.bound({ min: -262, max: 262 })),
and 
rangeMax: TlvField(1, TlvInt64.bound({ min: -262, max: 262 })),

should be something like this  in my opinion:

rangeMin: TlvField(0, TlvInt64.bound({ min: -(2 ** 62), max: 2 ** 62 })),
and 
rangeMax: TlvField(1, TlvInt64.bound({ min: -(2 ** 62), max: 2 ** 62 })),

lauckhart:
Special cased for now as we don't yet support constraint expressions.

Also includes minor validation optimizations.

Fixes #950

### DIFF
--- a/packages/matter.js/src/model/aspects/Constraint.ts
+++ b/packages/matter.js/src/model/aspects/Constraint.ts
@@ -202,8 +202,8 @@ export namespace Constraint {
 
     function parseValue(numOrName: string): FieldValue {
         let value;
-        if (numOrName.startsWith("0x") || numOrName.startsWith("0b")) {
-            value = Number.parseInt(numOrName);
+        if (numOrName.match(/^-?0[xb]/)) {
+            value = Number.parseInt(numOrName.replace(/[_ ]/g, ""));
         } else {
             value = Number.parseFloat(numOrName);
         }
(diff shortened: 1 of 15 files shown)


packages/matter.js/src/model/aspects/Constraint.ts
```diff
@@ -202,8 +202,8 @@ export namespace Constraint {
 
     function parseValue(numOrName: string): FieldValue {
         let value;
-        if (numOrName.startsWith("0x") || numOrName.startsWith("0b")) {
-            value = Number.parseInt(numOrName);
+        if (numOrName.match(/^-?0[xb]/)) {
```
Apollon77, line 205: When I remember correctly an analysis i once saw then in general startsWith is faster then a regex ...

Apollon77 (approved):

packages/matter.js/src/model/aspects/Constraint.ts
lauckhart: I added sign support so it would've become 4x startsWith

Apollon77: kk

PR #955, opened 2024-06-16
Update README.md device compatibility

jpadie:
updated the device support grid and added some more information about composed devices on Alexa.


Apollon77 (approved):
Thank you very much!

PR #959, opened 2024-06-17
Adding Pure-JS Crypto

Apollon77:
To experiment around with it better and also for react-native here is an intermediate Pure-JS Crypto. It bases on crypto-browserify with a patched aes-ccm lib based on https://github.com/browserify/browserify-aes/pull/57

It is done as an own packge because is uses several deprecated packages which generate several warnings and we might not need to "Infec"T the main matter.js package with that. It can be used here if needed.

@lauckhart tests work with CJS and ESM already, just web has issues and tell about "global not found" ... where can I tweak that in?

Still TODOs:
* Readme and other "content" files needs to be adjusted
* Get testing to run
* Is own package and such good approach? Naming?


Apollon77:
When I get the error in https://github.com/project-chip/matter.js/actions/runs/9555385608/job/26338682930#step:5:309 right then it refers to https://github.com/browserify/randombytes/blob/b4eb4126fccc503e375853ca7b3e96aaa34af9e6/browser.js#L15 ... is then globalThis not defined?

lauckhart:
The version of randombytes in NPM does not appear to use globalThis, it uses an undefined variable called "global", presumably expecting a bundler to define it for the browser.

There are a few ways to fix:

- If you're editing the library anyway you could remove the randombytes dependency and instead use crypto.getRandomValues which is available everywhere in modern JS
- If you're bundling as a library in web.ts you can instruct esbuild to define "global" as globalThis when bundling
- You could add globalThis.global = globalThis somewhere in the web bootstrap code
- If you build off of CryptoSubtle.ts in my crypto-js branch it has everything except for encrypt/decrypt.  You could implement those using browserify-aes directly and drop the rest of crypto-browserify
- You could build off CryptoSubtle.ts and just port over a CCM mode implementation like the one in the unmerged browserify-aes PR

lauckhart:
I just remembered you guys were targeting react native... I checked and it looks like react-native-quick-crypto supports AES-CCM. That might be worth exploring. Then sticking with the Node API could be an advantage as I'm not sure if they've even implemented enough of CryptoSubtle for that to be viable.

PR #960, opened 2024-06-19
Introducing Platform implementation for react-native

Apollon77:
This PR introduces the platform implementations for react-native (at least partially).

Included is:
* ble
* crypto
* network (UDP)
As storage we use the StorageBackendMemory, so nothing is persisted as of now. A Storage backend would still be needed mid-term

I also marked it "experimental" for now.

Iwould like to merge after review and then have a nightly to try it out by a dev.


Apollon77:
@lauckhart This test fail is because of the chip changes for 1.4 which will be integrated as fix in #961 ... so after merging #961 we can update this here and it should work

lauckhart (approved):

PR #969, opened 2024-06-25
Update Interactions, BTP, Session for Matter 1.3

Apollon77:
This PR got a bit bigger as planned, but can be reviewed by commit. It contains:
* Split up ValidationErrors to be more specific if they should return "InvalidAction" ot "ConstraintError". Basically invalid/unexpected datatypes are InvalidAction, "out of bound values" are ConstraintErrors
* Adjust some returned errors to be more specific and to the specification (e.g. InvalidAction instead of Failure)
* Fix that a to-be-sent StandaloneAck uses an outstanding Ack number as piggyback
* Added WildcardPathFlagsBitmap to Attribute expansion for read/subscribe Interactions
* Refactored Message size handling - channel knows his payload size and Exchange too (minus Matter message overheads). Removes the hard coded (too low) value of 1024 bytes for a matter message payload. New value is dynamically calculated but normally 1178 bytes. Also preparation for TCP
* Refactored and cleanup CASE and PASE and adding Matter 1.3 session params and corrected handling in some places
* Implemented Multi-Invoke Feature (default for now are 10 invokes till we have a better value)
* Fix: Make sure subscription maxInterval can not exceed the matter defined limit of 60mins
* Synced attMtu handling with chip to always use MTU-3 bytes for BLE connections
* Added BTP Idle timeout as defined in Matter specification
* Adjust MRP behavior with chip and only use/expect MRP ion unreliable channels (UDP). Fixes BLE commissioning
* Contains more optimizations and fixes to Interaction and BLE
* Revoked a change from this PR to send DataReports always with suppressResponse=true because chip does not do it and it breaks python tests (see https://github.com/project-chip/connectedhomeip/issues/34077)


packages/matter.js/src/protocol/interaction/InteractionServer.ts
```diff
@@ -1126,156 +1150,257 @@ export class InteractionServer implements ProtocolHandler<MatterDevice>, Interac
...
-        );
 
-        // TODO support suppressResponse for responses
-        return {
-            suppressResponse: false,
-            interactionModelRevision: INTERACTION_MODEL_REVISION,
-            invokeResponses,
-        };
+                const result = await tryCatchAsync(
+                    async () =>
```
lauckhart, line 1354: Maybe we could overload `tryCatchAsync` take a promise or a function?  Then `async () => await this.invokeCommand(...)` could become `this.invokeCommand(...)` and it'd remove an anonymous stack trace frame.  Even if not this could just be `() => this.invokeCommand(...)`.

But actually, if we just didn't use tryCatchAsync here I don't think it'd be a big hit to readability and it'd remove two stack frames and (I think) at least two promises from every invoke.

I've been going out of my way to avoid stack frames where reasonable in critical code paths...  Mainly so stack truncation removes less useful information but it has perf effect too (admittedly prolly marginal in network code).

packages/matter.js/src/protocol/interaction/InteractionServer.ts
```diff
@@ -1126,156 +1150,257 @@ export class InteractionServer implements ProtocolHandler<MatterDevice>, Interac
...
+                            path,
+                            command,
+                            exchange,
+                            commandFields ?? TlvNoArguments.encodeTlv(commandFields),
+                            message,
+                            endpoint,
+                            receivedWithinTimedInteraction,
+                        ),
+                    StatusResponseError,
+                    async error => {
```
lauckhart, line 1365: And the error handler parameter to tryCatchAsync seems like a good candidate for MaybeAsync given this isn't actually async.

packages/matter.js/src/common/ValidationError.ts
```diff
@@ -7,16 +7,42 @@
...
     public constructor(
         message: string,
         public fieldName?: string,
+        errorCode = StatusCode.ConstraintError,
     ) {
-        super(message, StatusCode.ConstraintError);
+        super(message, errorCode);
 
-        this.message = `(Validation/${StatusCode.ConstraintError}) ${message}`;
+        this.message = `(Validation/${this.code}) ${message}`;
```
lauckhart, line 23: Do you think we should include class name and (if applicable) status code in the error formatter?  I was thinking about that in some of my stuff.  We've got a couple different conventions for these so would be nice to normalize

lauckhart (approved):
Mostly piddly comments about code conventions but logic LGTM to the extent I understood it ;)

packages/matter-node.js-examples/src/examples/DeviceNodeFull.ts
```diff
@@ -297,6 +297,7 @@ const server = await ServerNode.create(RootEndpoint, {
         networks: [{ networkId: networkId, connected: !wifiOrThreadAdded }],
         scanMaxTimeSeconds: wifiOrThreadAdded ? 3 : undefined,
         connectMaxTimeSeconds: wifiOrThreadAdded ? 3 : undefined,
+        supportedWifiBands: wifiOrThreadAdded ? [NetworkCommissioning.WiFiBand["2G4"]] : undefined,
```
Apollon77, line 300: just for Wifi, for Thread add Support­edThread­Features

Apollon77: And ThreadVersion too for Thread

packages/matter.js/src/protocol/interaction/InteractionEndpointStructure.ts
```diff
@@ -33,6 +38,23 @@ import {
...
+const GLOBAL_ATTRIBUTELIST_ID = 0xfffb;
+const GLOBAL_EVENTLIST_ID = 0xfffa;
+const GLOBAL_COMMANDLIST_IDS = [
+    0xfff8, // GeneratedCommandList
+    0xfff9, // AcceptedCommandList
+];
+const GLOBAL_ATTRIBUTES_LIST = [...GLOBAL_COMMANDLIST_IDS, GLOBAL_EVENTLIST_ID, GLOBAL_ATTRIBUTELIST_ID];
+
+// Build a list of cluster IDs that are used for diagnostics to not always filter through model
+const DIAGNOSTICS_CLUSTER_IDS = MatterModel.standard.clusters
```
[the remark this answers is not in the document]
Apollon77: I added an Todo for now, but yes you are right!

packages/matter.js/src/protocol/interaction/InteractionMessenger.ts
```diff
@@ -84,7 +78,7 @@ class InteractionMessenger<ContextT> {
         return this.exchange.send(messageType, payload, options);
     }
 
-    sendStatus(status: StatusCode) {
+    async sendStatus(status: StatusCode) {
```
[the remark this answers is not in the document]
Apollon77: right ... changed back

packages/matter.js/src/common/ValidationError.ts
Apollon77: I adjusted

PR #970, opened 2024-06-26
Docs splitup

Apollon77:
The main Readme got too long and so I decided to split up the documents a bit. With this also other projects like matterbridge can link to it more easiely and we can build up one information source


Luligu:
Thanks

lauckhart (approved):

PR #975, opened 2024-06-27
Add proper name to cluster aliases

### LINKED ISSUE #974: All ConcentrationMeasurement derived clusters (e.g. Pm1ConcentrationMeasurement) are missing the correct name property
Hi,
when you define the CompleteInstance that will be registered with ClusterRegistry.register()

in this way:

export const CompleteInstance = MutableCluster({ ...ConcentrationMeasurement.Complete, id: 0x42c });

has a side effect that the name is taken from ConcentrationMeasurement and not from Pm1ConcentrationMeasurement.

Should be instead:
export const CompleteInstance = MutableCluster({ ...ConcentrationMeasurement.Complete, id: 0x42c, name: "Pm1ConcentrationMeasurement"});

Thanks.

lauckhart:
Fixes #974


Apollon77 (approved):
Thank you!

PR #978, opened 2024-06-28
Finalize Matter 1.3 Core Specs changes

Apollon77:
This PR finalizes the Matter 1.3 core spec changes:
* Update GeneralStatus enum
* Added ProtocolData handling for StatusReport
* Expose Exchange in action context
* Enhances GeneralDiagnosticsServer default implementation
* Added Schema for Specification version (for controller usage)
* Enhances/Fixes some cluster implememntations

TODO ist still to check and potentially enable chop tests for the foloowing clusters:
* DGGEN
* DESC - just manual test, so not relevant
* DGETH - validateds counters, so not relevant for now
* PWRTL - just validates value constrains when values are set, so nothing more for us


packages/matter.js/src/behavior/Behavior.ts
```diff
@@ -95,6 +95,14 @@ export abstract class Behavior {
         return this.#agent.context;
     }
 
+    get exchange() {
```
lauckhart, line 98: If use is the exchange is limited could we leave this out and just do error checking when we read from the context?

lauckhart (approved):

packages/matter.js/src/behavior/Behavior.ts
Apollon77: Alright. Will adjust

PR #983, opened 2024-06-30
Experimental AsyncStorage support for react-native

Apollon77:
Use @react-native-async-storage/async-storage to implement an AsyncStorage based storage.

The whole react-native part is experimental and will be verified by building a react-native app in the next weeks. AFter this we can mark it as stable.

@foxriver76 - please feel free to also review or comment


Apollon77:
Very strange is that - most likely because of a double default export  - the types for AsyncStorage tell that it only exists under AsyncStorage.default. That way it is building but I have big doubts if it will be working tbh. 
Anyone an idea?

Apollon77:
Ok, adjusted now to "without .default" because I assume it is just type issue ... see also https://react-native-async-storage.github.io/async-storage/docs/usage ... so we ignore the errors

packages/matter.js-react-native/src/storage/StorageBackendAsyncStorage.ts
lauckhart, line 1: Given that AsyncStorage is duck compatible with the standard Storage interface except that's it's async, maybe we could add a MaybeAsync version of standard storage that our local atorage implementation uses. Then this class could extend and just override the storage source... Would eliminate the duplicate code.

Alternatively, maybe a base class implementation of our storage that does all of the key fiddling? Just trying to think how to avoid the code copying

lauckhart (approved):

packages/matter.js-react-native/src/storage/StorageBackendAsyncStorage.ts
Apollon77: I agree. I will add a task ... lets first get react-native working ... then I know which cleanups are needed :-) network will also partially duplicate code  in next pr

PR #988, opened 2024-07-01
Adjust React-Native NetworkInterface handling

Apollon77:
... which was missed.

This also required to make the "network interface" relevant methods MaybePromise-style for all

### DIFF
--- a/packages/matter.js-react-native/src/net/NetworkReactNative.ts
+++ b/packages/matter.js-react-native/src/net/NetworkReactNative.ts
@@ -3,12 +3,168 @@
  * Copyright 2022-2024 Matter.js Authors
  * SPDX-License-Identifier: Apache-2.0
  */
-
 import dgram from "react-native-udp";
 
 // @ts-expect-error globalThis is no index structure
 global.dgram = dgram;
 
-import { NetworkNode } from "@project-chip/matter-node.js/net";
+import { Logger } from "@project-chip/matter.js/log";
+import {
+    InterfaceType,
+    Network,
+    NetworkError,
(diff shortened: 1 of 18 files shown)


packages/matter.js-react-native/src/net/NetworkReactNative.ts
lauckhart, line 1: Rather than copying this code from network-node.ts could we instead make a common base class or just have the dgram implementation and any other variance be configurable via constructor parameters?

packages/matter.js-react-native/src/net/UdpChannelReactNative.ts
lauckhart, line 1: Should find way to avoid duplication here too

lauckhart (changes requested):
Lgtm except see comments r.e. code duplication

packages/matter.js-react-native/src/net/NetworkReactNative.ts
Apollon77: Agreed, I have this on refactoring list once it works ... no idea how many iterations we need. Once we are done I will do this. promise
Todo already in project https://github.com/orgs/project-chip/projects/11/views/1?filterQuery=reac&pane=issue&itemId=69408905

packages/matter.js-react-native/src/net/UdpChannelReactNative.ts
Apollon77: see other comment https://github.com/orgs/project-chip/projects/11/views/1?filterQuery=reac&pane=issue&itemId=69408905

lauckhart (approved):

PR #1006, opened 2024-07-09
Bump react-native-quick-crypto and typescript 

Apollon77:
replaces https://github.com/project-chip/matter.js/pull/997 which returned an error on dgram types with this upgrade. And also updated typescript to 5.5.3

### DIFF
--- a/packages/matter.js-react-native/src/net/UdpChannelReactNative.ts
+++ b/packages/matter.js-react-native/src/net/UdpChannelReactNative.ts
@@ -8,10 +8,48 @@ import dgram from "react-native-udp";
 import { Diagnostic, Logger } from "@project-chip/matter.js/log";
 import { MAX_UDP_MESSAGE_SIZE, NetworkError, UdpChannel, UdpChannelOptions } from "@project-chip/matter.js/net";
 import { ByteArray } from "@project-chip/matter.js/util";
-import type { RemoteInfo, Socket, SocketOptions } from "dgram";
 import { NetworkReactNative } from "./NetworkReactNative.js";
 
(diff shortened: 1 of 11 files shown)


packages/matter.js-react-native/src/net/UdpChannelReactNative.ts
```diff
@@ -8,11 +8,49 @@ import dgram from "react-native-udp";
 import { Diagnostic, Logger } from "@project-chip/matter.js/log";
 import { MAX_UDP_MESSAGE_SIZE, NetworkError, UdpChannel, UdpChannelOptions } from "@project-chip/matter.js/net";
 import { ByteArray } from "@project-chip/matter.js/util";
-import type { RemoteInfo, Socket, SocketOptions } from "dgram";
```
lauckhart, line 11: This is a node import. Probably rn crypto or one of its dependencies has an improper dependency on node's types that they fixed so it's no longer available to you unless you add it to tsconfig.

Given that a.) these are intended to be drop-in replacements for node libs and b.) you'll need to use a single type definition to have a shared implementation...  I would just cast the top-level imports to node's types and be done with it.

So e.g. something like:

```typescript
const dgram = rnDgram as unknown as typeof global.dgram;
const crypto = rnCrypto as unknown as typeof global.crypto;
```

lauckhart (approved):

packages/matter.js-react-native/src/net/UdpChannelReactNative.ts
Apollon77: I add this comment as link to the cleanup todo.

PR #1011, opened 2024-07-10
Fix Subscription clearing

Apollon77:
The most current definition of subscriber is a node on a fabric ... and we missed the fabric in the last adjustment. This might be just a special case for testing because of overlapping peerNodeIds but better fix properly


packages/matter.js/src/session/SessionManager.ts
```diff
@@ -373,9 +374,9 @@ export class SessionManager<ContextT> {
             }));
     }
 
-    async clearSubscriptionsForNode(nodeId: NodeId, flushSubscriptions?: boolean) {
+    async clearSubscriptionsForNode(fabricIndex: FabricIndex, nodeId: NodeId, flushSubscriptions?: boolean) {
         for (const session of this.#sessions) {
-            if (session.peerNodeId === nodeId) {
+            if (session.fabric?.fabricIndex === fabricIndex && session.peerNodeId === nodeId) {
```
Apollon77, line 379: Leave open please. I need to do one local test manually tomorrow morning

lauckhart (approved):

PR #1031, opened 2024-07-19
Better handle Async subscription updates

Apollon77:
Addresses Review feedback from former PR

### DIFF
--- a/packages/matter.js/src/protocol/interaction/SubscriptionHandler.ts
+++ b/packages/matter.js/src/protocol/interaction/SubscriptionHandler.ts
@@ -111,7 +111,7 @@ export class SubscriptionHandler {
     private sendingUpdateInProgress = false;
     private sendNextUpdateImmediately = false;
     private sendUpdateErrorCounter = 0;
-    private attributeUpdatePromises = new Array<PromiseLike<void>>();
+    private attributeUpdatePromises = new Set<PromiseLike<void>>();
 
     constructor(options: {
         subscriptionId: number;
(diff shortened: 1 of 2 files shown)


packages/matter.js/src/protocol/interaction/SubscriptionHandler.ts
```diff
@@ -713,7 +713,26 @@ export class SubscriptionHandler {
         );
     }
 
-    async attributeChangeListener<T>(path: AttributePath, schema: TlvSchema<T>, version: number, value: T) {
+    attributeChangeListener<T>(path: AttributePath, schema: TlvSchema<T>, version: number, value: T) {
+        const changeResult = this.attributeChangeHandler(path, schema, version, value);
+        if (MaybePromise.is(changeResult)) {
+            this.attributeUpdatePromises.push(
+                changeResult.then(() => {
```
lauckhart, line 720: If changeResult can throw then this should probably be `Promise.resolve(changeResult).finally(`

packages/matter.js/src/protocol/interaction/SubscriptionHandler.ts
```diff
@@ -713,7 +713,26 @@ export class SubscriptionHandler {
...
+                    const promiseIndex = this.attributeUpdatePromises.indexOf(changeResult);
+                    if (promiseIndex !== -1) {
+                        this.attributeUpdatePromises.splice(promiseIndex, 1);
+                    }
```
lauckhart, line 724: If you use a set for attributeUpdatePromises then this could just be `this.attributeUpdatePromises.delete`

lauckhart (approved):

lauckhart (approved):

PR #1046, opened 2024-07-22
Fix: Ensure all controllerStorage.get calls are properly awaited

Orzech99:
# Fix: Ensure all `controllerStorage.get` calls are properly awaited

## Description

This PR fixes an inconsistency in the way `controllerStorage.get` calls are handled. Previously, the `uniqueid` was awaited but `ip` and `port` were not, which could lead to unexpected behavior due to the asynchronous nature of JavaScript.

### Changes Made

- Ensured all `controllerStorage.get` calls in this example are properly awaited.

## Checklist

- [x] Ensure all async `get` calls are awaited.
- [x] Code compiles and runs without errors.
- [x] All changes are tested and verified.

### DIFF
--- a/packages/matter-node.js-examples/src/examples/ControllerNode.ts
+++ b/packages/matter-node.js-examples/src/examples/ControllerNode.ts
@@ -77,10 +77,10 @@ class ControllerNode {
 
         const controllerStorage = (await storageService.open("controller")).createContext("data");
         const ip = (await controllerStorage.has("ip"))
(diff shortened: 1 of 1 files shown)


Apollon77:
Thank you , yes that was a miss because the example usually uses a "SyncStorage", but in fact correct if someone uses a different (async) Storage implementation

Apollon77 (approved):

PR #1061, opened 2024-07-29
Revert #1052 and fix properly

Apollon77:
We should just disable Scenes support in CI-PICS and we should be fine


lauckhart (approved):

Apollon77:
Ok will not work because top level pics are ignored ;-)

PR #1062, opened 2024-07-29
Try real hard to surface error details

lauckhart:
Some environments do not report `Error#cause` or `AggregateError#errors`.  Or if they do they may not report them when deeply nested (e.g. with Node's default `util.inspect` implementation).

To work around this, on node.js we now monitor uncaught exceptions and report them in matter.js logs using matter.js's formatting.  We also offer error formatting via `MatterError#format()` and provide a custom inspector for node.js.

Includes related improvements:

  - Modifies object formatting so it's available outside of logging
  - Adds more comprehensive tests for error formatting logic

### DIFF
--- a/packages/matter-node.js/src/environment/ProcessManager.ts
+++ b/packages/matter-node.js/src/environment/ProcessManager.ts
@@ -42,6 +42,12 @@ export class ProcessManager implements Destructable {
         this.runtime.started.on(this.startListener);
         this.runtime.stopped.on(this.stopListener);
         this.runtime.crashed.on(this.crashListener);
+
+        if (this.hasUnhandledErrorSupport) {
+            process.addListener("uncaughtExceptionMonitor", event => {
+                Logger.reportUnhandledError(event);
+            });
+        }
     }
 
     close() {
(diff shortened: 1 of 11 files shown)


packages/matter-node.js/src/environment/ProcessManager.ts
```diff
@@ -63,6 +69,10 @@ export class ProcessManager implements Destructable {
         return this.env.vars.get("runtime.exitcode", true);
     }
 
+    protected get hasUnhandledErrorSupport() {
+        return this.env.vars.get("runtime.unhandlederrors", true);
```
Apollon77, line 73: Wer should add that to the in-code docs from NodeJsEnvironment to nbe documented for now? WDYT?

Apollon77 (approved):

Apollon77:
PS: After all got merged I will do a PR to try nodejs 22 in testing ...

lauckhart:
> PS: After all got merged I will do a PR to try nodejs 22 in testing ...

Oh good I was going to mention

packages/matter-node.js/src/environment/ProcessManager.ts
lauckhart: added

PR #1088, opened 2024-08-11
Allow MaxPathsPerInvoke to use default logic

Apollon77:
The change adjusts the MaxPathsPerInvoke attribute to have a minimum of 0 again so the out default logic applies.

### DIFF
--- a/models/src/local/BasicInformationOverrides.ts
+++ b/models/src/local/BasicInformationOverrides.ts
@@ -10,14 +10,15 @@ LocalMatter.children.push({
     tag: "cluster",
     name: "BasicInformation",
 
+    // We override default in Model to 0 in order to use our default settings.
+    // Both API versions use defaults when 0 would exist.
     children: [
         {
             tag: "attribute",
-            name: "ProductAppearance",
-            id: 0x14,
-            type: "ProductAppearanceStruct",
-            conformance: "O",
-            quality: "F",
+            name: "MaxPathsPerInvoke",
+            id: 0x16,
+            constraint: "min 0",
+            default: 0,
         },
     ],
 });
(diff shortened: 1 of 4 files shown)


models/src/local/BasicInformationOverrides.ts
```diff
@@ -10,14 +10,15 @@ LocalMatter.children.push({
...
             tag: "attribute",
-            name: "ProductAppearance",
-            id: 0x14,
-            type: "ProductAppearanceStruct",
-            conformance: "O",
-            quality: "F",
+            name: "MaxPathsPerInvoke",
+            id: 0x16,
+            constraint: "min 0",
+            default: 0,
```
lauckhart, line 21: I think it'd be better to maintain the default in the schema but change it in the cluster when we use it, but not a big deal.

lauckhart (approved):

models/src/local/BasicInformationOverrides.ts
Apollon77: How you mean that?

lauckhart: Don't create the override but rather modify BasicInformationCluster when we use it (via e.g. `BasicInformationCluster.set`).  Just so the model represents the spec's "truth" but we can be more flexible

Apollon77: my problem with this is that I can not determine of the value 1 is from the default or because the dev explicitely wanted to set 1 ... so yes I could so "if ===1 then set to Default aka 10", but i can not determine if the 1 is from dev or from default. Or what do I miss?

Apollon77:
replaced by #1103

PR #1098, opened 2024-08-13
Handful of small improvements & fixes

### LINKED ISSUE #1086: Scraper generates wrong Names when deprecated names are included
See DoorLock cluster.

When the table contains a new and a deprecated name of an attribute with an empty line between them then this is all concatinated, leading to strange names.
This mainly affects Doorlock but in approx 20 cases.

I would favor a scraper/generator fix rather then overrides.

lauckhart:
- Add parent device types to DeviceTypeList for derived clusters
- Fix imported garbage in some DoorLock field names
- Change typing for Behavior#features to include disabled features
- Explicitly define DescriptorServer as an endpoint requirement so attributes are configurable in TS
- Add DescriptorServer#addTag to make adding tags more convenient
- Modify DescriptorServer#addDeviceType to accept device type name to simplify avoidance of cyclical dependencies
- Modify additional behaviors to add device types when DT is implied by behavior's presence

fixes #1086

### DIFF
--- a/models/src/v1.3/spec.ts
+++ b/models/src/v1.3/spec.ts
@@ -11650,9 +11650,22 @@ export const SpecMatter = Matter({
                         "\n" +
                         "† The PIN/RFID Code is an obsolete field name, use PINCode instead.",
                     xref: { document: "cluster", section: "5.2.10.1" },
-                    children: [
-                        Field({ name: "PinCodePinRfidCode", id: 0x0, type: "octstr", conformance: "[COTA & PIN]" })
-                    ]
+
+                    children: [Field({
+                        name: "PinCode", id: 0x0, type: "octstr", conformance: "[COTA & PIN]",
+
+                        details: "If the RequirePINforRemoteOperation attribute is True then PINCode field shall be provided and the " +
+                            "door lock shall NOT grant access if it is not provided." +
+                            "\n" +
+                            "If the PINCode field is provided, the door lock shall verify PINCode before granting access " +
+                            "regardless of the value of RequirePINForRemoteOperation attribute." +
+                            "\n" +
+                            "When the PINCode field is provided an invalid PIN will count towards the WrongCodeEntryLimit and " +
+                            "the UserCodeTemporaryDisableTime will be triggered if the WrongCodeEntryLimit is exceeded. The lock " +
+                            "shall ignore any attempts to lock/unlock the door until the UserCodeTemporaryDisableTime expires.",
+
+                        xref: { document: "cluster", section: "5.2.10.1.1" }
+                    })]
                 }),
 
                 Command({
@@ -11671,9 +11684,11 @@ export const SpecMatter = Matter({
                         "† The PIN/RFID Code is an obsolete field name, use PINCode instead.",
 
                     xref: { document: "cluster", section: "5.2.10.2" },
-                    children: [
-                        Field({ name: "PinCodePinRfidCode", id: 0x0, type: "octstr", conformance: "[COTA & PIN]" })
-                    ]
+                    children: [Field({
+                        name: "PinCode", id: 0x0, type: "octstr", conformance: "[COTA & PIN]",
+                        details: "See PINCode field.",
+                        xref: { document: "cluster", section: "5.2.10.2.1" }
+                    })]
                 }),
 
                 Command({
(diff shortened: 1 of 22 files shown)


packages/matter.js/src/behavior/definitions/descriptor/DescriptorServer.ts
```diff
@@ -48,23 +52,76 @@ export class DescriptorServer extends DescriptorBehavior {
                     revision: partType.deviceRevision,
                 },
             ];
+            for (
```
Apollon77, line 55: See discussion in Discord ... it is unclear which consequences this has based on specs and "duplicate condition" for device types.
So I would propose to comment it out ( so that it is easy to readd based on the clarification on spec level ... for now lets just allow a dev to add it manually

packages/matter.js/src/behavior/definitions/descriptor/DescriptorServer.ts
```diff
@@ -48,23 +52,76 @@ export class DescriptorServer extends DescriptorBehavior {
...
                 }
             }
             list.push(newDeviceType);
         }
-        this.state.deviceTypeList = list;
+    }
+
+    /**
+     * Add semantic tags.  This is a shortcut for deduped insert into the tagList cluster attribute.
+     */
```
Apollon77, line 100: Please add here the info that for this method to work the feature needs to be enabled, so that devs see it at least in the IDE if they look

packages/matter.js/src/behavior/definitions/device-energy-management/DeviceEnergyManagementServer.ts
```diff
@@ -15,4 +14,10 @@ import { DeviceEnergyManagementBehavior } from "./DeviceEnergyManagementBehavior
  * default. You should use {@link DeviceEnergyManagementServer.with} to specialize the class for the features your
  * implementation supports.
  */
-export class DeviceEnergyManagementServer extends DeviceEnergyManagementBehavior {}
+export class DeviceEnergyManagementServer extends DeviceEnergyManagementBehavior {
+    override initialize() {
+        // Technically ElectricalSensor also requires Device EnergyManagementMode (as of Matter 1.3) but no other device
+        // type requires DeviceEnergyManagement
+        this.agent.get(DescriptorServer).addDeviceTypes("ElectricalSensor");
```
Apollon77, line 21: Formally check that a PowerTopology cluster also exists on the endpoint because that one is the mandatory one?

packages/matter.js/src/behavior/definitions/power-topology/PowerTopologyServer.ts
```diff
@@ -26,6 +27,8 @@ const PowerTopologyBase = PowerTopologyBehavior.with(
  */
 export class PowerTopologyServerLogic extends PowerTopologyBase {
     override initialize(): void {
+        this.agent.get(DescriptorServer).addDeviceTypes("ElectricalSensor");
```
Apollon77, line 30: Formally check that a either Electrical Power Measurement or Electrical Energy Measurement cluster also exists on the endpoint because it is only allowed in this combination?

models/src/v1.3/spec.ts
```diff
@@ -11650,9 +11650,22 @@ export const SpecMatter = Matter({
                         "\n" +
                         "† The PIN/RFID Code is an obsolete field name, use PINCode instead.",
                     xref: { document: "cluster", section: "5.2.10.1" },
-                    children: [
-                        Field({ name: "PinCodePinRfidCode", id: 0x0, type: "octstr", conformance: "[COTA & PIN]" })
-                    ]
+
+                    children: [Field({
+                        name: "PinCode", id: 0x0, type: "octstr", conformance: "[COTA & PIN]",
```
Apollon77, line 11655: Thats great ... please also regenerate model :-) because seems thats missing

packages/matter.js/src/behavior/definitions/descriptor/DescriptorServer.ts
lauckhart: OK

packages/matter.js/src/behavior/definitions/descriptor/DescriptorServer.ts
lauckhart: OK

packages/matter.js/src/behavior/definitions/device-energy-management/DeviceEnergyManagementServer.ts
lauckhart: Meant to revert this file

packages/matter.js/src/behavior/definitions/power-topology/PowerTopologyServer.ts
lauckhart: Yes, good idea

Apollon77 (approved):
Awesome, thank you!

PR #1107, opened 2024-08-15
Check for both Wifi and Ethernet in Step 11 during the commissioning process

### LINKED ISSUE #1106: Commissioning of Wifi devices broken in v10
I upgraded my dependencies to v10 recently, and while everything else works fine, i can no longer commission wifi devices.  The process fails on step 11 here:

https://github.com/project-chip/matter.js/blob/e6636a20d810fc9b48201248eea80691a45e8446/packages/matter.js/src/protocol/ControllerCommissioner.ts#L777-L779

If i print out on the console the values of `this.collectedCommissioningData.networkFeatures` I get:

```
{ endpointId: 0, value: { wiFiNetworkInterface: true, threadNetworkInterface: false, ethernetNetworkInterface: false } } 
```

but the code at:

https://github.com/project-chip/matter.js/blob/e6636a20d810fc9b48201248eea80691a45e8446/packages/matter.js/src/protocol/ControllerCommissioner.ts#L773-L775

~~treats `value` as an array using `some`, and not an object as its structured in  `networkFeatures`.~~

If i comment out the `throw` statement here, then commissioning works as expected.

digitaldan:
Fixes #1106

### DIFF
--- a/packages/matter.js/src/protocol/ControllerCommissioner.ts
+++ b/packages/matter.js/src/protocol/ControllerCommissioner.ts
@@ -763,17 +763,18 @@ export class ControllerCommissioner {
             this.commissioningOptions.threadNetwork === undefined
         ) {
             // Check if we have no networkCommissioning cluster or an Ethernet one
-            const anyEthernetInterface =
+            const anyNetworkInterface =
                 this.collectedCommissioningData.networkFeatures.length === 0 ||
                 this.collectedCommissioningData.networkFeatures.some(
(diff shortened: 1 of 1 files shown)


Apollon77:
Please see my comment on the issue. This change makes no sense because it would accept any not connected WiFi or thread cluster as ok. This check is there to verify if we can commission it when WiFi or thread is not connected.

Apollon77:
According to discussion in #1106 and the (in my eyes) better solution in #1109 I close here. Thank you anyway for your support and PR!

PR #1122, opened 2024-08-23
Include attr/event callbacks when resubscribing a paired node.

### LINKED ISSUE #1121: Paired Node looses attribute callbacks on re-subscribing
Hi, I noticed that after a few days of running a modified version of the shell app some nodes would send attribute changed messages, but my callbacks were not being called. I tracked this down to this line:

https://github.com/project-chip/matter.js/blob/81c0c32f0c2abf360aceb980189bd7387ef2dfc2/packages/matter.js/src/device/PairedNode.ts#L375-L378

which is subscribing again when it detects a missed subscription event, but is not passing in the callbacks originally setup with the paired node.  

I have a PR that fixes this for me and maybe better describes the issue, but feel free to disregard/close it if there is a better solution ;-)

digitaldan:
Fixes #1121

### DIFF
--- a/packages/matter.js/src/device/PairedNode.ts
+++ b/packages/matter.js/src/device/PairedNode.ts
@@ -374,12 +374,7 @@ export class PairedNode {
             updateTimeoutHandler: async () => {
                 logger.info(`Node ${this.nodeId}: Subscription update not received ...`);
                 try {
-                    const { attributeChangedCallback, eventTriggeredCallback } = this.options;
-                    await this.subscribeAllAttributesAndEvents({
-                        ignoreInitialTriggers: true,
-                        attributeChangedCallback: data => attributeChangedCallback?.(this.nodeId, data),
-                        eventTriggeredCallback: data => eventTriggeredCallback?.(this.nodeId, data),
-                    });
+                    await this.subscribeAllAttributesAndEvents({ ...options, ignoreInitialTriggers: false });
                 } catch (error) {
                     logger.info(`Node ${this.nodeId}: Error resubscribing to all attributes and events`, error);
                     // TODO resume logic right now retries and discovers for 60s .. prolong this but without command repeating
--- a/packages/matter.js/src/device/PairedNode.ts
+++ b/packages/matter.js/src/device/PairedNode.ts
@@ -374,7 +374,12 @@ export class PairedNode {
             updateTimeoutHandler: async () => {
                 logger.info(`Node ${this.nodeId}: Subscription update not received ...`);
                 try {
-                    await this.subscribeAllAttributesAndEvents();
+                    const { attributeChangedCallback, eventTriggeredCallback } = this.options;
+                    await this.subscribeAllAttributesAndEvents({
+                        ignoreInitialTriggers: true,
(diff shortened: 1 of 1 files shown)


packages/matter.js/src/device/PairedNode.ts
```diff
@@ -374,7 +374,12 @@ export class PairedNode {
             updateTimeoutHandler: async () => {
                 logger.info(`Node ${this.nodeId}: Subscription update not received ...`);
                 try {
-                    await this.subscribeAllAttributesAndEvents();
+                    const { attributeChangedCallback, eventTriggeredCallback } = this.options;
+                    await this.subscribeAllAttributesAndEvents({
+                        ignoreInitialTriggers: true,
+                        attributeChangedCallback: data => attributeChangedCallback?.(this.nodeId, data),
```
Apollon77, line 380: Why you change here and next line the callback method? In fact what's missing is thatched original options are also set when resubscribing. So passing the callbacks from options just like in options of the subscribeAllAttributesAndEvents method should be sufficient or?

packages/matter.js/src/device/PairedNode.ts
```diff
@@ -374,7 +374,12 @@ export class PairedNode {
             updateTimeoutHandler: async () => {
                 logger.info(`Node ${this.nodeId}: Subscription update not received ...`);
                 try {
-                    await this.subscribeAllAttributesAndEvents();
+                    const { attributeChangedCallback, eventTriggeredCallback } = this.options;
+                    await this.subscribeAllAttributesAndEvents({
+                        ignoreInitialTriggers: true,
```
Apollon77, line 379: That's a good question depending on how long it stayed disconnected we could miss data. So better miss them all as change callbacks or get change-calls that are for the same value. Maybe second is better or? Better more then missing? Then this should be false.

packages/matter.js/src/device/PairedNode.ts
```diff
@@ -374,7 +374,12 @@ export class PairedNode {
             updateTimeoutHandler: async () => {
                 logger.info(`Node ${this.nodeId}: Subscription update not received ...`);
                 try {
-                    await this.subscribeAllAttributesAndEvents();
+                    const { attributeChangedCallback, eventTriggeredCallback } = this.options;
```
Apollon77, line 377: This.options or better using what comes in in this method as params?

Apollon77 (changes requested):
Hey, I commented ... I am just on mobile currently so I might not have the full picture of the file in my head right now ... but I hope my comments make sense.

packages/matter.js/src/device/PairedNode.ts
digitaldan: yeah, i totally spaced this was actually calling itself and already had the options  passed in.

packages/matter.js/src/device/PairedNode.ts
digitaldan: Yeah, my oversight.

Apollon77 (approved):

Apollon77:
Thank you. Fix will be in next nightly tomorrow.

PR #1188, opened 2024-09-12
Update ECOSYSTEMS.md

sammachin:
add flic as an ecosystem, will add more test results when I get to creating the devices


Apollon77:
Thank you. Could you please also add a line in https://github.com/project-chip/matter.js/blob/main/README.md#compatibility-to-ecosystems

Apollon77 (approved):

PR #1190, opened 2024-09-14
Controller reconnect optimizations

Apollon77:


packages/matter.js/src/protocol/interaction/SubscriptionHandler.ts
```diff
@@ -855,13 +862,13 @@ export class SubscriptionHandler {
...
-        } catch (e) {
-            if (StatusResponseError.is(e, StatusCode.InvalidSubscription, StatusCode.Failure)) {
+        } catch (error) {
+            if (StatusResponseError.is(error, StatusCode.InvalidSubscription, StatusCode.Failure)) {
                 logger.info(`Subscription ${this.subscriptionId} cancelled by peer.`);
                 await this.cancel(false, true);
             } else {
+                StatusResponseError.accept(error);
                 await this.cancel(false);
-                throw e;
```
lauckhart, line 864: It seems like this could hide a significant amount of error information, maybe there's a more specific StatusResponseError to test for?

packages/nodejs/src/net/UdpChannelNode.ts
```diff
@@ -129,7 +130,10 @@ export class UdpChannelNode implements UdpChannel {
         return new Promise<void>((resolve, reject) => {
             this.socket.send(data, port, host, error => {
                 if (error !== null) {
-                    const netError = new NetworkError(error.message);
+                    const netError =
+                        error instanceof Error && error.message.includes("EHOSTUNREACH")
```
lauckhart, line 134: Testing error.code === "EHOSTUNREACH" seems cleaner than searching the error message

Apollon77: good idea ... 

I decided for 

> "code" in error && error.code === "EHOSTUNREACH"

for simplicity ... alternatively I would have needed to check how to get the correct typing for this special Nodejs ErrnoException type

packages/matter.js/src/protocol/interaction/SubscriptionHandler.ts
Apollon77: I would add a logline in this case but formally (after spec) no other status error should happen then the two that are already handled 3 lines above (InvalidSubscription, Failure).

lauckhart (approved):

PR #1210, opened 2024-09-19
Add Lifecycle finalized events triggered before starting the node

Apollon77:
We should finalize/migrate ACL before we start the node. Also useful for other cases when we need to have these in the final data to make sure wwe do not change after someone already tries to connect.


packages/node/src/node/NodeLifecycle.ts
```diff
@@ -55,6 +56,13 @@ export class NodeLifecycle extends EndpointLifecycle {
         return this.#online;
     }
 
+    /**
+     * Emits when the node is finalized and goes online next
+     */
+    get finalized() {
```
lauckhart, line 62: I'm wondering if we can avoid adding the event, it doesn't seem very general purpose.  Why do we initialize AccessControlManager when the node comes online rather than when AccessControlServer initializes?

Is it just because FabricManager isn't available until the node is online, and we need that to perform the upgrade from 0.7?  In my ClientNode branch I pulled FabricManager initialization into node startup, so if that's the problem maybe I could push that so AccessManager initialization could happen sooner.

lauckhart (changes requested):

Apollon77:
We skip that for now and do when w have a use case

PR #1219, opened 2024-09-20
Store 1000 history lines for shell app

Apollon77:
... because it is just convenient :-)

Also removes unused dependency


lauckhart (approved):
Probably future TODOs - 1.) limit history file size, 2.) mask secrets from history file

Also I vote for 1000 lines 😉

Apollon77:
The file size is limited on each start by cutting to the maximum lines after initial read and rewriting file once. I think this should do it for now or? Improved to 1000

PR #1229, opened 2024-09-22
Typescript and Package updates

Apollon77:
* Updates to TS 5.6
* Adjusts all packages ... basic idea is that all is in what reguklar users need in deps, else general stiff is in the main package ... we need to see if that works

### DIFF
--- a/packages/main/package.json
+++ b/packages/main/package.json
@@ -53,7 +53,8 @@
         "@matter.js/react-native": "*"
     },
     "devDependencies": {
-        "@matter.js/tools": "*"
+        "@matter.js/tools": "*",
+        "@matter.js/codegen": "*"
     },
     "files": [
         "dist/**/*",
--- a/packages/main/package.json
+++ b/packages/main/package.json
@@ -53,9 +53,7 @@
         "@matter.js/react-native": "*"
     },
     "devDependencies": {
-        "@matter.js/tools": "*",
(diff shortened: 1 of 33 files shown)


packages/matter.js/package.json
```diff
@@ -53,15 +53,7 @@
         "@matter.js/node": "*"
     },
     "devDependencies": {
-        "@matter.js/tools": "*",
-        "@matter.js/codegen": "*",
```
lauckhart, line 57: This dependency was legit, src/build.config.ts uses `generateSubmoduleProxy` from codegen

packages/main/package.json
```diff
@@ -53,9 +53,7 @@
         "@matter.js/react-native": "*"
     },
     "devDependencies": {
-        "@matter.js/tools": "*",
-        "@matter.js/codegen": "*",
```
lauckhart, line 57: This one is legit

lauckhart (approved):
Cool thanks.  The two public packages that with dev deps on codegen were legit, it's used in build to generate the "proxies" for submodule re-exports

packages/matter.js/package.json
Apollon77: yes found that out too now ;) re-adding

packages/main/package.json
Apollon77: kk

PR #1255, opened 2024-10-02
Controller Refactoring and Optimizations

Apollon77:
tbd


packages/protocol/src/interaction/InteractionClient.ts
```diff
@@ -123,16 +133,26 @@ export class SubscriptionClient implements ProtocolHandler {
 }
 
 export class InteractionClient {
-    // TODO Add storage for these data to be able to optimize follow up subscriptions by only request updated data
-    private readonly subscribedLocalValues = new Map<string, any>();
-    private readonly subscribedClusterDataVersions = new Map<string, number>();
+    readonly #cachedData: NodeCachedData = {
+        attributeValues: new Map(),
```
lauckhart, line 137: Caching feels like a separate concern, could there be a separate component with the same read interface except extended with the caching features? Would be nice to have a class just focused on protocol

lauckhart (approved):

packages/general/src/util/PromiseQueue.ts
```diff
@@ -0,0 +1,105 @@
...
+     * Add a promise to the queue. It returns a promise that ca be awaited.
+     */
+    add<T>(executor: () => Promise<T>): Promise<T> {
+        const { promise, resolver, rejecter } = createPromise<T>();
+
+        logger.debug("Add promise to queue on place", this.#queue.length + 1);
+        this.#queue.push({
+            func: () =>
+                executor()
+                    .then(value => {
```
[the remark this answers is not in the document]
Apollon77: How you mean?

packages/protocol/src/interaction/InteractionClient.ts
Apollon77: In fact this caching interface should be more a "NodeEndpointStore" as next step. To also utilize that on controller restarts. So I would like to leave for now but I as a todo. Because as mentioned in discord question is how to store this as next. 

So basically RootStore/EndpointStoreService?

packages/general/src/util/PromiseQueue.ts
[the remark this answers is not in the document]
lauckhart: `executor().then(resolver).catch(rejecter)`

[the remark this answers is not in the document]
Apollon77: Ahh that way. Kk.

packages/protocol/src/interaction/InteractionClient.ts
lauckhart: Sure, for ClientNode I was just planning to store "last known" values using the same NodeStore as the server (I renamed it from "ServerNodeStore" to "NodeStore" in one of the refactoring PRs).

There are three scenarios where (I think) the backing store is basically just node -> endpoint -> cluster -> attribute:

* Device where we "master" the data and have the canonical values
* Bridge where our values are the "last known" reported by the bridged device
* Client where our values are the "last known" reported by the peer node

I'm thinking for client we may just want a small extension to have a place for caching operational information (e.g. the stuff in OperationalPeer).

Apollon77: In fact this current "cache" is what's needed for proper client operation. So exactly as you say. I just did not do it in this pr. If ok I would add a todo to adjust this. In fact yes i could now split out this cache into own class but then we remove it once we do it "right". And I hoped I can also use NodeStore approach in PairedNode context ;-)

packages/general/src/util/PromiseQueue.ts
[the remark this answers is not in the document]
Apollon77: done

packages/general/src/util/PromiseQueue.ts
```diff
@@ -0,0 +1,105 @@
...
+            this.#runningCount++;
+            func()
+                .catch(() => {}) // already catched internally, can not happen
+                .finally(() => {
+                    logger.debug("Promise processed ... Still running:", this.#runningCount - 1);
+                    if (this.#delay > 0) {
+                        // Keep the queue blocked for the delay time
+                        Time.sleep("Queue delay", this.#delay)
+                            .then(() => this.#runNext())
+                            .catch(() => {}); // not relevant
```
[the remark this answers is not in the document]
Apollon77: done

packages/nodejs-shell/src/shell/cmd_nodes.ts
```diff
@@ -10,6 +10,15 @@ import { CommissioningControllerNodeOptions, NodeStateInformation } from "@proje
 import type { Argv } from "yargs";
 import { MatterNode } from "../MatterNode";
 
+const NodeStateString = {
```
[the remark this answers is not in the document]
Apollon77: cool, works

packages/matter.js/src/device/PairedNode.ts
```diff
@@ -343,32 +673,83 @@ export class PairedNode {
         logEndpoint(rootEndpoint, options);
     }
 
+    #determineSubscriptionParameters() {
```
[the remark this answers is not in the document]
Apollon77: Done, kind of that way?

packages/nodejs-shell/src/shell/cmd_nodes.ts
```diff
@@ -4,20 +4,15 @@
...
 
-const NodeStateString = {
-    [NodeStateInformation.Connected]: "Connected",
-    [NodeStateInformation.Disconnected]: "Disconnected",
-    [NodeStateInformation.Reconnecting]: "Reconnecting",
-    [NodeStateInformation.WaitingForDeviceDiscovery]: "Waiting for device discovery",
-    [NodeStateInformation.StructureChanged]: "Structure changed",
-    [NodeStateInformation.Decommissioned]: "Decommissioned",
-};
+function capitalizeFirstLetter(str: string) {
```
lauckhart, line 13: There's "capitalize" in String.ts 🙂

packages/matter.js/src/device/PairedNode.ts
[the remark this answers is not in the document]
lauckhart: Yeah, looks good

packages/protocol/src/interaction/InteractionClient.ts
lauckhart: PeerStore is another case of "need persistence from higher level tier", might be able to do something similar here but doesn't need to be in this PR.

NodeStore might need a couple of tweaks for client use cases but I think it'd work fine for PairedNode, shouldn't be anything specific to the "Node" API.

I used the attribute names as keys instead of the IDs for ServerNode. Which won't work for unknown attributes (can't remember if you are loading those). But that's not required, ID would work too

packages/nodejs-shell/src/shell/cmd_nodes.ts
Apollon77: lol ... yeeaahh

packages/protocol/src/interaction/InteractionClient.ts
Apollon77: Lets talk next days how we can best do it ... For the data enrichment we also need to have convenient ways to access the stored data ...

Apollon77: PS: unkown Attributes are getting an name like unknownAttribute_0x123456 in the current structures, so by utilizing their IDs ... maybe could also be used ... But I have now in all the structures also the pure IDs because easier to work with then the strings.

PR #1258, opened 2024-10-04
Allow to discover VendorId + ProductId together optionally

Apollon77:
... and a mini shell optimization


lauckhart (approved):
Cool, I thought it was weird that wasn't supported

PR #1267, opened 2024-10-06
Fix exports for cjs and Example enhancements

Apollon77:
... and adds a Measured Socket example ... and it adds a convenience method for EEM which handles also event triggering


packages/node/src/behaviors/electrical-energy-measurement/ElectricalEnergyMeasurementServer.ts
```diff
@@ -4,15 +4,95 @@
...
  * default. You should use {@link ElectricalEnergyMeasurementServer.with} to specialize the class for the features your
  * implementation supports.
  */
-export class ElectricalEnergyMeasurementServer extends ElectricalEnergyMeasurementBehavior {}
+export class ElectricalEnergyMeasurementServerLogic extends ElectricalEnergyMeasurementServerBase {
+    /**
+     * Convenience method to set the measurement values for the cluster in an atomic way. This method updates the
+     * attributes and also sends the events for the imported and exported energy values.
+     */
+    setMeasurement(measurement: AtomicElectricalEnergyMeasurement) {
```
lauckhart, line 47: Could we react to the attribute$Changed instead?

To avoid triggering events more than once I would add change handlers that start a "send update" timer (if not already started)

When the timer fires then look at the previous cumulative value to calculate the periodic value

A few benefits to this approach:

* Simplifies API a bit
* Ensures events fire regardless of how developer updates attributes
* Prevents events from emitting until the changes have been successfully committed
* Could use timer to coalesce multiple changes of same attribute into one notification (e.g. only notify every 10s even if dev updates attribute every 100ms)

Apollon77: I had the same thought at the beginning but did this approach because of the following rationales:
* reacting to attributes changes in the default implementation would not that easiely allow the developer to do own implementation (or he needs to override "Initialize" which would potentially have other effects and felt not ideal). So if we do that we would need an extra state to disable default behavior
* The main thing on events in my eyes is that they represent "a measurement" means updates or up to 4 values. So yes we could assume things and have a "timer approach" which waits Xms or such for all attributes to be set (assumed the dev is not doing it within one transaction), but this is already again assumptions and adds logic that can have side effects.
* I decided to ignore the "do not update more often then 1s" and such and move that to 1.4 when that should handled by new logic and quality anyway. So for now the dev is responsible.
* I would also not start calculating any values (like periodic out of cumulatives and time or such) ... The dev should still be responsible to send out data relevant to the featureset he enabled.

Out of all these thoughts I came to this idea of having this method for a dev to push. "a measurement" to the cluster and we set and handle things - otherwise he is responsible to do it ... at least so long to the background of my thoughts

lauckhart: I think it's good to have a single method like the one you created to "update current energy values".  I'm more looking at how we wire it under the hood so the cluster works correctly regardless of which API the developer chooses to adopt.

R.e. "need extra an state to disable default behavior", I would make two main extension points: 1.) the function that sends the updates would be a protected method, so dev could override however they see fit, and 2.) if that's insufficient dev can extend *Behavior instead of *Server and do everything themselves.

R.e. the second/third point, I was assuming one transaction...  Timer was mainly to coalesce from the individual attribute events so e.g. setTimeout(0).  Allowing for delaying updates would be an additional feature but maybe doesn't make sense if we have another way of doing it

R.e. the last bullet, why don't you want to calculate the periodic values?  Semantically I'm not sure there's a big difference between "you give me periodic and I calculate cumulative" and "you give me cumulative and I calculate periodic".  Functionally though the second one guarantees dev can't break the cluster by accidentally bypassing functionality.

Apollon77: Ok, as discussed in Discord a "onChange" approach will not work because it is "per measurement" and also "per period end" ... sooo ...lets stay with the proposed way for now ... We can enhance when we have more usecases

lauckhart (approved):

PR #1278, opened 2024-10-12
Some refactoring, additional work on client node and new "cli tool"

lauckhart:
Should be reviewable either by commit or by package.  I can create separate PRs if you'd prefer.

High-level summary:

- A couple minor build fixes
- Removes MatterDevice from ServerNode and moves back into matter.js as only legacy code paths use it now
- Refactors root certificate management a bit and adds an independent component for managing fabrics we control
- Adds a new REPL that offers a hybrid shell/JS environment
- Moves some storage concerns from node store into other components and creates subclasses with client- and server-specific functionaltiy
- More progress on ClientNode but still WIP (got distracted by REPL)


packages/node/src/node/storage/ClientNodeStore.ts
```diff
@@ -0,0 +1,77 @@
...
+
+    /**
+     * Load a store for each peer that was previously persisted.
+     */
+    static async load(peerStorage: StorageContext) {
+        const addresses = await peerStorage.contexts();
+        const stores = Array<ClientNodeStore>();
+
+        for (const addrStr of addresses) {
+            const addrComponents = addrStr.match(/^([0-9]+)-([0-9a-f])/i);
```
Apollon77, line 39: maybe $ at the end? or can there be other content after it?

Apollon77 (approved):
Wow impressive work ... Awesome to have MatterDevice "out of the new API" ... that's cool and a huge step forward to easiely cleanup the legacy API.

Also the cli-tool is cool ... yes docs missing, but can be done later

I just found one mini nit thing in a regex ...
And please add CHANGELOG entries for the relevant (mainly RootCertificate class rename and the cli-tool additionion or what you zhink makes sense too

packages/node/src/node/storage/ClientNodeStore.ts
lauckhart: Need to check but I think you're probably right

lauckhart:
> Also the cli-tool is cool ... yes docs missing, but can be done later

Yeah it's kind of just a toy until I get the Node stuff in there anyway, but will get docs in order soon

packages/node/src/node/storage/ClientNodeStore.ts
Apollon77: When I saw the code some lines below then it should be exactly this ... so with a $ could get easier for the regex engine. other question tbh is if not a "split" is faster ... the number conversions afterwards should make sure it was valid ... https://measurethat.net/Benchmarks/Show/7822/0/strmatch-vs-strsplit

lauckhart: I didn't mention RootCertificateManager rename because the old export (from matter.js) still works...  Just to stay consistent with other components we renamed.  Could add these in but maybe in separate PR

PR #1294, opened 2024-10-17
Fix test issues 

Apollon77:


lauckhart (approved):

packages/node/test/node/ServerNodeTest.ts
```diff
@@ -316,14 +316,14 @@ describe("ServerNode", () => {
         // Simulate receiving the response to the removeFabric request which normally closes the underlying session
         // delayed
         await contextOptions.exchange.session.destroy(false, false);
-
```
Apollon77, line 319: A comment to keep unmerged to do some more runs for test.

PR #1297, opened 2024-10-19
Maybe locks not needed?

Apollon77:


Apollon77:
As discussed offline was good to find these places and we leave them in even if not needed with normal speed storage.

PR #1316, opened 2024-10-26
Lifecycle shutdown fixes

Apollon77:
This PR:
* introduces a "goingOffline" lifecycle event and moves some logic using it
* tweak shutdown by closing sessions at right place before we go offline because else sessions stay open
* also clear resumption records Map when clearing storage


packages/node/src/behaviors/basic-information/BasicInformationServer.ts
```diff
@@ -63,6 +63,9 @@ export class BasicInformationServer extends Base {
         if (lifecycle.online !== undefined) {
             this.reactTo(lifecycle.online, this.#online);
         }
+        if (lifecycle.goingOffline !== undefined) {
```
lauckhart, line 66: What's the purpose of this test?

lauckhart (approved):

PR #1338, opened 2024-10-30
Install all required dependencies on project init

lauckhart:
Also includes @types/node on project init.

Makes @project-chip/matter.js a dev dependency of @matter/nodejs because it's only needed for tests.

### DIFF
--- a/packages/create/src/build.config.ts
+++ b/packages/create/src/build.config.ts
@@ -43,12 +43,14 @@ export async function before({ project }: Project.Context) {
             }
             const source = await readFile(file, "utf-8");
 
-            for (const [, pkgName] of source.matchAll(/from "(@[^/"]+\/[^/"]+|[^/"]+)[^"]*"/g)) {
+            // Quick hack to pull out imports, assumes modules formatted by prettier.  More than sufficient for current
+            // needs
+            for (const [, pkgName] of source.matchAll(/import .* from "(@[^/"]+\/[^/"]+|[^/"]+)[^"]*";/g)) {
                 if (dependencies[pkgName]) {
                     continue;
                 }
 
-                if (pkgName.startsWith("@matter/")) {
+                if (pkgName.startsWith("@matter/") || pkgName.startsWith("@project-chip/")) {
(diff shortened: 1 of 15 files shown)


packages/create/src/build.config.ts
```diff
@@ -35,7 +41,27 @@ export async function before({ project }: Project.Context) {
             if (!entrypoint && filename.indexOf("/") === -1) {
                 entrypoint = filename;
             }
-            await cp(file, createPkg.resolve("dist/templates/", name, filename));
+            const source = await readFile(file, "utf-8");
+
+            for (const [, pkgName] of source.matchAll(/from "(@[^/"]+\/[^/"]+|[^/"]+)[^"]*"/g)) {
```
Apollon77, line 46: nit, maybe to be more specific also add a space at the beginning and a semicolon at the end?

Apollon77: and does that work with "@matter/main/types" ? because would try to add this package as dep in L51? I think you need to cut a bit more or?

packages/create/src/build.config.ts
```diff
@@ -35,7 +41,27 @@ export async function before({ project }: Project.Context) {
...
+                if (dependencies[pkgName]) {
+                    continue;
+                }
+
+                if (pkgName.startsWith("@matter/")) {
```
Apollon77, line 51: should we also handle the @project-chip/matter.js dep here and add?

Apollon77 (commented):
.. and I think you need to generate package lock anew or?

packages/create/src/build.config.ts
lauckhart: Addressed the nit and added comment referencing half-ass nature.  Careful or I'll end up doing it correctly with typescript api and this code will get much uglier. ;)

R.e. second question, fine as is, it only extracts prefixes of the form "foo" or "@foo/bar".

packages/create/src/build.config.ts
lauckhart: yes, fixed.  It worked before but only because the ble package (incorrectly) had a dependency on @project-chip/matter.js

PR #1340, opened 2024-10-30
Adds additional features to generated projects

lauckhart:
* Adds .gitignore so you can add directly to git
* Adds .vscode/launch.json so you can do "code ." then hit f5 to debug


packages/create/src/new-project.ts
```diff
@@ -41,6 +41,33 @@ const TSCONFIG = {
...
 
+const VS_CODE_LAUNCH = {
+    version: "0.2.0",
+    configurations: [
+        {
+            type: "node",
+            request: "launch",
+            name: "Launch Program",
+            skipFiles: ["<node_internals>/**"],
+            program: "${workspaceFolder}/dist/SensorDeviceNode.js",
```
Apollon77, line 52: intended?

Apollon77 (approved):

packages/create/src/new-project.ts
lauckhart: nope

PR #1347, opened 2024-11-01
Documentation generator

lauckhart:
This tool generates documentation using typedoc.  Output is kind of a mess but I think about as close as we'll get to useful without a more custom solution.

Also fix a few issues the doc generator warned about.

Removes the "--workspaces" CLI option for building.  If you're in a sub-package of a monorepo, the tool works on that.  Otherwise it works on the monorepo as a whole.


Apollon77 (approved):
Ok for me but ou might want to make CodeQL happy ;-)

PR #1384, opened 2024-11-11
Contributor onboarding

lauckhart:
Adds "contributor" template for initializing local matter dev environment.  Doesn't really do much, just pulls and installs, but does walk the user through steps and setting up origin and main branches.

Also adds basic VS code configuration, including recommended extensions and a launch file for each test suite, example and generator.  Originally I generated as part of init but decided it should just go into git instead.  Still generated, just statically.

Also modifies test runner to execute tests in the same process instead of running npm repeatedly in each process.  With smaller projects this saves a significant amount of time and will be more significant when we fix and/or drop IntegrationTest.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,9 @@ The main work (all changes without a GitHub username in brackets in the below li
 
 ## __WORK IN PROGRESS__
 
+-   @matter/node
+    - Enhancement: The `with` functions on endpoint and cluster behavior types now alias to `withBehaviors` and `withFeatures` respectively to make their function more explicit
+
 -   @matter/nodejs
     - Feature: New export @matter/nodejs/config allows for fine-grained configuration of Node.js bootstrap logic
 


CHANGELOG.md
```diff
@@ -14,6 +14,16 @@ The main work (all changes without a GitHub username in brackets in the below li
 -   @matter/nodejs
     - Feature: New export @matter/nodejs/config allows for fine-grained configuration of Node.js bootstrap logic
 
+-   @matter/create
+    - Feature: Add command line option "--verbose" to enable informational NPM messages during initialization
+    - Feature: Add template "contributor" to bootstrap dev environment for working on matter.js itself
```
Apollon77, line 19: Should be also add this to the main README explicitely or into CONTRIBUTING.md or such?

PS: And maybe also add the withFeatures/withBehaviors to changelog :-)

Apollon77 (approved):

CHANGELOG.md
lauckhart: K I added the changelog entry and will wait on updating README/CONTRIBUTING pending testing once released.

PR #1394, opened 2024-11-13
Remove $Changed from logical flow of emitter

lauckhart:
Previously we were providing the ActionContext of the emitter to $Changed handlers.  The transaction was destroyed in this case so was not really useful.  Now we instead just provide metadata about the subject that triggered the change.

Also, the emitter was awaiting $Changed handlers if they were async.  There are advantages and disadvantages to doing this:

* Advantage is that the logical flow is more deterministic and we track the promise as a normal part of node activity
* The disadvantage is that the emitter blocks until an independent process completes

This commit makes it so $Changed handlers operate independently of the emitter.  Tests pass but effects will be subtle so will need to keep an eye on this.

### DIFF
--- a/packages/node/src/behavior/AccessControl.ts
+++ b/packages/node/src/behavior/AccessControl.ts
@@ -124,7 +124,7 @@ export namespace AccessControl {
     /**
      * Information about the subject that triggered a change.
      */
-    export interface Subject {
+    export interface Actor {
         /**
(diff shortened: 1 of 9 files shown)


Apollon77:
see review comments in Discord - also reason for the failed testcase

lauckhart:
Closing because this is covered in upcoming "Q" quality implementation

PR #1399, opened 2024-11-15
Ignore persisted values when features change

### LINKED ISSUE #1161: Update cluster features in storage/consider only current and not stored features
Suggestion that conformance errors are not automatically fatal errors in all cases.  As an example use case:

set an onoff lighting device with colour support and colour temperature support and provide defaults for the mandatory attributes.
then remove colour temperature support.

the deployment will fail for conformance because an attribute is in the device storage that is contrary to the conformance requirements.  But this could be silently ignored instead of resulting in a fatal error.

lauckhart:
Affects individual behaviors on endpoints with persisted values.  Storage must be tagged with features, so storage created by previous versions of code will not benefit from this enhancement.

Fixes #1161

### DIFF
--- a/packages/node/src/behavior/state/managed/Datasource.ts
+++ b/packages/node/src/behavior/state/managed/Datasource.ts
@@ -231,9 +231,10 @@ function configure(options: Datasource.Options): Internals {
     let featuresKey: undefined | string;
     if (options.supervisor.featureMap.children.length) {
         featuresKey = [...options.supervisor.supportedFeatures].join(",");
-        if (storedValues?.[FEATURES_KEY] !== undefined && storedValues[FEATURES_KEY] !== featuresKey) {
+        const storedFeaturesKey = storedValues?.[FEATURES_KEY];
+        if (storedFeaturesKey !== undefined && storedFeaturesKey !== featuresKey) {
             logger.warn(
-                `Ignoring persisted values for ${options.path} because features changed from "${values.featuresKey}" to "${featuresKey}"`,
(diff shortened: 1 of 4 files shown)


packages/node/src/behavior/state/managed/Datasource.ts
```diff
@@ -223,11 +226,28 @@ interface CommitChanges {
...
     const values = new options.type() as Val.Struct;
 
+    let storedValues = options.store?.initialValues;
+
+    let featuresKey: undefined | string;
+    if (options.supervisor.featureMap.children.length) {
+        featuresKey = [...options.supervisor.supportedFeatures].join(",");
+        if (storedValues?.[FEATURES_KEY] !== undefined && storedValues[FEATURES_KEY] !== featuresKey) {
+            logger.warn(
+                `Ignoring persisted values for ${options.path} because features changed from "${values.featuresKey}" to "${featuresKey}"`,
```
Apollon77, line 236: should this log not also use "storedValues?.[FEATURES_KEY]" ?!

Apollon77 (approved):

Apollon77:
@lauckhart And please add this into Changelog in any case. thank you!

packages/node/src/behavior/state/managed/Datasource.ts
lauckhart: yes, fixed

lauckhart:
> @lauckhart And please add this into Changelog in any case. thank you!

wtf, the changelog entry is there, a rebase or something pushed into a separate commit

lauckhart:
oh, I probably forgot to push after I added then the pull of the merge created a separate commit

PR #1403, opened 2024-11-16
Collection of minor adjustments and refactorings

Apollon77:
see commits


packages/protocol/src/interaction/ServerSubscription.ts
```diff
@@ -568,7 +568,7 @@ export class ServerSubscription extends Subscription {
             }
 
             this.sendUpdateErrorCounter++;
-            logger.error(
+            logger.info(
```
Apollon77, line 571: do not log stack trace

PR #1412, opened 2024-11-20
Fix collection

Apollon77:
This PR addresses some smaller fixes, see commits
* round mireds-to-kelvin in central place
* network interface names longer 32 chars are cutted now (Seems that WSL sometimes does 34 chars)
* Fix async storage added in 0.11.0 to be backwards compatible to node-localstorage in case of special chars. Could be breaking if older 0.11 versions were used initially with special chars IDs (should be uncommon), but I decided that backward compat to <0.11 is more important for now
* Troubleshooting enhancements

### DIFF
--- a/packages/nodejs/src/storage/StorageBackendDiskAsync.ts
+++ b/packages/nodejs/src/storage/StorageBackendDiskAsync.ts
@@ -86,7 +86,10 @@ export class StorageBackendDiskAsync extends MaybeAsyncStorage {
             throw new StorageError("Key must not be an empty string.");
         }
         const contextKey = this.getContextBaseKey(contexts);
-        return `${contextKey}.${key}`;
+        const rawName = `${contextKey}.${key}`;
+        return encodeURIComponent(rawName)
+            .replace(/[!'()]/g, escape)
+            .replace(/\*/g, "%2A");
     }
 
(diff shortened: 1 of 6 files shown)


packages/nodejs/src/storage/StorageBackendDiskAsync.ts
```diff
@@ -161,8 +164,9 @@ export class StorageBackendDiskAsync extends MaybeAsyncStorage {
         const files = await readdir(this.#path);
 
         for (const key of files) {
-            if (key.startsWith(contextKeyStart) && !key.includes(".", len)) {
-                keys.push(key.substring(len));
+            const decodedKey = decodeURIComponent(key);
```
lauckhart, line 167: Confirming, this'll encode the vast majority of characters.  You think that's the way to go just to ensure we never have to worry about filesystem oddities?

lauckhart (approved):

packages/nodejs/src/storage/StorageBackendDiskAsync.ts
Apollon77: it is in fact just mimic the sync lib e used before https://github.com/lmaccherone/node-localstorage/blob/master/LocalStorage.coffee#L152 ... and yes with this we should be safe

PR #1478, opened 2024-12-05
New model algorithms and APIs for scope analysis

### LINKED ISSUE #1455: ModeTag validation uses wrong "ModeTags" instance
This issue is the essence out of https://github.com/project-chip/matter.js/discussions/1444#discussioncomment-11412712

Repro use the attacked example file.

My findings are that two things are "wrong":
1.) In element ModeBase the datatype ModeTagStruct has a child "Value" which is as scraped an enum16 but do not have any children. It needs to be changed to type "ModeTag" to point to the "right" types

2.) When then a derived cluster (e.g. EvseMode) is used then the ValueValidator - when working on the attribute supportedModes and all "below that structure like then also  "ModeTag" type is use the definition of MopdBase, so all "derived overwritten types" are ignored because the validator iterated down the schema - and this schema is "ModeBase".
I also tried to "copy" all the datatypes and the relevant attribute to the EnergyEvseMode.ts element file for tests, but it was still not taking these "overwritten" things

Eample script
```javascript
import { Endpoint, ServerNode } from "@matter/main";
import { EnergyEvseModeServer } from "@matter/main/behaviors";
import { EnergyEvseMode } from "@matter/main/clusters";
import { OnOffPlugInUnitDevice } from "@matter/main/devices";

// Create the "node".  In Matter a "node" is a standalone device
const node = await ServerNode.create();

// Define the endpoint for the device.  This is the "thing/device" that will appear on the network.
const measuredSocketEndpoint = new Endpoint(OnOffPlugInUnitDevice.with(EnergyEvseModeServer), {
    id: "socket",

    energyEvseMode: {
        currentMode: EnergyEvseMode.ModeTag.Manual,
        startUpMode: EnergyEvseMode.ModeTag.Manual,
        supportedModes: [
            {
                label: "Manual",
                mode: 0,
                modeTags: [
                    {
                        value: EnergyEvseMode.ModeTag.Manual,
                    },
                ],
            },
            {
                label: "Quick",
                mode: 1,
                modeTags: [
                    {
                        value: EnergyEvseMode.ModeTag.TimeOfUse,
                    },
                ],
            },
        ],
    },
});

// Create the light "endpoint".  In Matter an "endpoint" is a component of a node
await node.add(measuredSocketEndpoint);

// Run our server
await node.start();
```

lauckhart:
Previously we identified implicit base models of datatypes and other model children, but we only did this from the perspective of the child (the "extension"), not from the perspective of the base.  This worked fine until Matter 1.3 when the specification began relying more heavily on the "extended cluster" construct.

This commit adds the notion of a "scope" to the model.  Instead of querying a model for applicable members, you now query the scope.  The scope is aware of extensions, so this allows you to retrieve the correct set of members in contexts (such as datatype fields defined in the base cluster) where there is otherwise no explicit reference to the extension.

CHANGELOG.md contains additional details.

Fixes #1455

@Apollon77 - this one is a somewhat large and has a couple of breaking changes, up to you whether we put in a dot release

### DIFF
--- a/packages/types/src/globals/WildcardPathFlags.ts
+++ b/packages/types/src/globals/WildcardPathFlags.ts
@@ -58,5 +58,5 @@ export const WildcardPathFlags = {
     /**
      * Skip all clusters with the Diagnostics (K) quality during wildcard expansion.
      */
-    wildcardSkipDiagnosticsClusters: BitFlag(8),
+    wildcardSkipDiagnosticsClusters: BitFlag(8)
 };
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,15 @@ The main work (all changes without a GitHub username in brackets in the below li
 
 -   @project-chip/matter.js
     - Enhancement: Considers a node in reconnection state that should be decommissioned as already factory reset
+    - Fix: Do not try to convert color mode details if they are not defined
+    - Fix: Clusters generated for extensions of base clusters such as Alarm Base and Mode Base now include full details of extended types; in particular extended enums such as Mode Tag were previously insufficiently defined
+    - BREAKING: In `ContentLauncher` cluster `ParameterEnum` is renamed to `Parameter` and `Parameter` is renamed to `ParameterStruct`
+
+-   @matter/model
+    - Feature: New `Scope` component analyzes scope of a model, caches analysis results, and implements algorithms that require analysis to perform efficiently
+    - Enhancement: Models that define datatypes now inherit from common `ScopeModel` base class
+    - Fix: Extended enums and other types now report the full set of members via `Scope#membersOf`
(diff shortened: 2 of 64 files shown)


packages/types/src/globals/WildcardPathFlags.ts
```diff
@@ -31,9 +31,9 @@ export const WildcardPathFlags = {
     wildcardSkipAttributeList: BitFlag(2),
 
     /**
-     * Reserved
+     * Skip the EventList global attribute during wildcard expansion.
```
Apollon77, line 34: Please revert this rename because should be that way since we official do 1.4 ... but we removed events already ... so yes it is a "manual ignreo change"

Apollon77: Ok I do an override in another PR ... all good

Apollon77:
On hold for 0.12!!

Apollon77:
We decided to bring it into 0.11

Apollon77 (approved):

PR #1484, opened 2024-12-05
WindowCovering default implementation enhancement

Apollon77:
We introduce a new state flag to disable the operationalState and position value handling by the default implementation. If a real device is used then this device should know these details better nd should be used.


packages/node/src/behaviors/window-covering/WindowCoveringServer.ts
```diff
@@ -713,6 +723,13 @@ export namespace WindowCoveringServerLogic {
 
         /** Does the device supports maintenance mode? */
         supportsMaintenanceMode: boolean = true;
+
+        /**
+         * Disable OperationalMode and position value management.
+         * This requires the device developer to set all these states (operationalMode, percentage and
+         * absolute values according to the feature set) according to the Matter specification!
+         */
+        disableOperationalModeHandling: boolean = false;
```
lauckhart, line 732: Is this more of a property of the implementation than state? 

WDYT about either a property you can override (could be static if it would never be conditional, or an instance property), or a protected method such as initializeOperationalStateManagement that could be overridden to disable default behavior?

OTOH if you're thinking a "real" device could implement without otherwise extending WindowCoveringServer then we could leave on State. But it seems like the device would need to implement some of the commands.

lauckhart (approved):

packages/node/src/behaviors/window-covering/WindowCoveringServer.ts
Apollon77: Ahh you mean an internal state part that is then send on initialize? Yes could make sense too. Will check tomorrow.

Apollon77: moved to internal

PR #1502, opened 2024-12-11
Event persistence

lauckhart:
Note: Depends on #1499 so ignore first commit

Adds the term "occurrence" to differentiate between "classes of events that I can subscribe to" and "historical instances of said event".

Splits the functionality of EventHandler into storage (EventStore) and logic (OccurrenceManager).  Adds volatile and non-volatile EventStore implementations.  Extends the algorithm for pruning old events a bit.

By default OccurrenceManager operates in volatile mode.

Primarily needed now for testing, but created with public API as reliable event delivery seems like table stakes for nodes playing a role in security.


Apollon77 (approved):
I added Changelog entry and resolved conflicts after other PR got merged

PR #1504, opened 2024-12-11
GeneralDiagnostics fixes for TC_DGGEN_2_4

lauckhart:
- Omit posixTimeMs as required by TC_DGGEN_2_4; spec language is poor but I don't think this is required formally
- TC_DGGEN_2_4 also fails if uptime is greater than system time; our old way of computing would fail because of behavior initialization order
- As part of fixing previous added "system start time" and "process start time" concepts to time implementation; they default to current time but nodejs overrides with host/proc information


packages/node/src/behaviors/general-diagnostics/GeneralDiagnosticsServer.ts
```diff
@@ -103,8 +103,11 @@ export class GeneralDiagnosticsServer extends Base {
     override timeSnapshot() {
         const time = Time.nowMs();
         return {
-            systemTimeMs: time - this.internal.bootUpTime,
-            posixTimeMs: time,
+            systemTimeMs: time - Time.startup.systemMs,
+
+            // TC_DGGEN_2_4.py fails us if we set this without TimeSynchronizationCluster support.  Spec is worded
+            // poorly but my read of "SHALL only if" is not the same as "SHALL if".  But conforming to tests for now
```
Apollon77, line 109: In fact yes One spec sentence is wired but in the attribute description it is described clear:
Bildschirmfoto 2024-12-11 um 14 00 34

Why you do not check for existence of the TimeSync cluster and populate the value correctly as defined?

lauckhart: That sentence is the one that confuses me... "shall only be null if" means “may not be null unless", not "must be null if".

So it MAY be null but is not required to be.  And why would they prevent you from providing the value just because it can't be changed?

I didn't add check for time sync because it didn't occur to me, I figured would be changed if time sync added. I'll add

lauckhart: OK I added the test for TimeSynchronizationCluster.  I left the comment in but elaborated a bit more.  I still think the test suite is wrong based on formal interpretation of that language.  So at very least it's a spec issue.

Apollon77: lets merge that way because seems chip understood it that way. So lets see

Apollon77 (approved):

PR #1505, opened 2024-12-11
Tweak two error codes

lauckhart:
As required by TestAccessControlCluster.yaml, unsure of formal requirement

### DIFF
--- a/packages/node/src/behaviors/access-control/AccessControlServer.ts
+++ b/packages/node/src/behaviors/access-control/AccessControlServer.ts
@@ -109,11 +109,23 @@ export class AccessControlServer extends AccessControlBehavior {
         for (const entry of fabricAcls) {
             const { privilege, subjects, targets, authMode } = entry;
             if (subjects !== null && subjects.length > this.state.subjectsPerAccessControlEntry) {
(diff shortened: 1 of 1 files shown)


lauckhart:
@Apollon77 looks like we have patched files that have different error codes...  I don't see any background info but LMK if this is something you've discussed w/ chip folks.  We can either adjust our code for now to work w/ official test suites (per this PR) or I can patch the test before running in new harness.

Personally if this is officially a CHIP bug I'd probably lean toward emulating the broken behavior when (or if) they fix the tests.

Apollon77:
I would like to keep our implementation because thats a chip bug

Spec states:
> An attempt to add an Access Control Entry when no more entries are available SHALL result in a
RESOURCE_EXHAUSTED error being reported and the ACL attribute SHALL NOT have the entry
added to it. See access control limits.

Chip issue link: https://github.com/project-chip/connectedhomeip/issues/33578

lauckhart:
I patch the tests at runtime rather than copying the entire YAML. But easy enough to do.

My concern is this will be a friction point though for anybody attempting certification. OTOH chip may then fix it. I'll go ahead and implement the patch

lauckhart:
Oh right, I just realized this isn't a cert test so won't cause friction either

PR #1506, opened 2024-12-11
Updates for Operational Credentials cert tests

lauckhart:
- Modify a couple of Operational Credentials errors to return NocResponse with cluster status code rather than throwing StatusResponseError.  Not sure why this is preferable but required by cert tests
- Change status code to that expected by cert tests, although they seem a little arbitrary
- Allow root cert to be set in an update timed op as CHIP does but fail subsequent UpdateNoc.  This seems nonsensical but is required by TC_OPCREDS_3_4

### DIFF
--- a/packages/node/src/behaviors/operational-credentials/OperationalCredentialsServer.ts
+++ b/packages/node/src/behaviors/operational-credentials/OperationalCredentialsServer.ts
@@ -119,7 +119,7 @@ export class OperationalCredentialsServer extends OperationalCredentialsBehavior
 
         if (isForUpdateNoc && this.session.isPase) {
             throw new StatusResponseError(
-                "csrRequest for UpdateNoc received on a PASE session.",
+                "csrRequest for UpdateNoc received on a PASE session",
                 StatusCode.InvalidCommand,
             );
         }
@@ -128,7 +128,7 @@ export class OperationalCredentialsServer extends OperationalCredentialsBehavior
         const failsafeContext = commissioner.failsafeContext;
         if (failsafeContext.fabricIndex !== undefined) {
             throw new StatusResponseError(
-                `csrRequest received after ${failsafeContext.forUpdateNoc ? "UpdateNOC" : "AddNOC"} already invoked.`,
+                `csrRequest received after ${failsafeContext.forUpdateNoc ? "UpdateNOC" : "AddNOC"} already invoked`,
                 StatusCode.ConstraintError,
             );
         }
@@ -205,7 +205,7 @@ export class OperationalCredentialsServer extends OperationalCredentialsBehavior
 
         if (failsafeContext.fabricIndex !== undefined) {
             throw new StatusResponseError(
-                `addNoc received after ${failsafeContext.forUpdateNoc ? "UpdateNOC" : "AddNOC"} already invoked.`,
+                `AddNoc received after ${failsafeContext.forUpdateNoc ? "UpdateNOC" : "AddNOC"} already invoked`,
                 StatusCode.ConstraintError,
             );
         }
@@ -213,20 +213,20 @@ export class OperationalCredentialsServer extends OperationalCredentialsBehavior
         if (!failsafeContext.hasRootCert) {
             return {
                 statusCode: OperationalCredentials.NodeOperationalCertStatus.InvalidNoc,
-                debugText: "Root certificate not found.",
+                debugText: "Root certificate not found",
             };
         }
 
         if (failsafeContext.csrSessionId !== this.session.id) {
             return {
                 statusCode: OperationalCredentials.NodeOperationalCertStatus.MissingCsr,
-                debugText: "CSR not found in failsafe context.",
+                debugText: "CSR not found in failsafe context",
             };
         }
 
         if (failsafeContext.forUpdateNoc) {
             throw new StatusResponseError(
-                `addNoc received after csr request was invoked for UpdateNOC.`,
+                `addNoc received after csr request was invoked for UpdateNOC`,
                 StatusCode.ConstraintError,
             );
         }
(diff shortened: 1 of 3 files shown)


packages/node/src/behaviors/operational-credentials/OperationalCredentialsServer.ts
```diff
@@ -313,22 +313,29 @@ export class OperationalCredentialsServer extends OperationalCredentialsBehavior
             );
         }
 
-        if (timedOp.forUpdateNoc) {
+        if (timedOp.forUpdateNoc === false) {
             throw new StatusResponseError(
-                `addNoc received after csr request was invoked for UpdateNOC.`,
+                `UpdateNoc received after csr request was invoked for UpdateNOC.`,
```
Apollon77, line 318:
```suggestion
                `UpdateNoc received after csr request was invoked for AddNOC.`,
```

packages/node/src/behaviors/operational-credentials/OperationalCredentialsServer.ts
```diff
@@ -313,22 +313,29 @@ export class OperationalCredentialsServer extends OperationalCredentialsBehavior
...
-                "Trusted root certificate added in this session which is now allowed for UpdateNOC.",
-                StatusCode.ConstraintError,
-            );
+        if (timedOp.rootCertSet) {
+            return {
+                // TC_OPCREDS_3_4 expects MissingCsr here...  Previously we used InvalidNoc which seems like it makes
+                // more sense
+                statusCode: OperationalCredentials.NodeOperationalCertStatus.MissingCsr,
+                fabricIndex: this.session.fabric?.fabricIndex,
+                debugText: "Trusted root certificate added in this session which is now allowed for UpdateNOC.",
```
Apollon77, line 329:
```suggestion
                debugText: "Trusted root certificate added in this session which is not allowed for UpdateNOC.",
```

packages/node/src/behaviors/operational-credentials/OperationalCredentialsServer.ts
```diff
@@ -395,7 +402,10 @@ export class OperationalCredentialsServer extends OperationalCredentialsBehavior
     override addTrustedRootCertificate({ rootCaCertificate }: OperationalCredentials.AddTrustedRootCertificateRequest) {
         const failsafeContext = this.#failsafeContext;
 
-        if (failsafeContext.hasRootCert) {
+        // TC_OPCREDS_3_4 fails if we don't allow set of the root certificate in updates, even though that's illegal and
```
Apollon77, line 405: where exactly in the python test this is set before?

https://github.com/project-chip/connectedhomeip/blob/master/src/python_testing/TC_OPCREDS_3_4.py

I only see addTrustedRoot Cert call once?

packages/node/src/behaviors/operational-credentials/OperationalCredentialsServer.ts
```diff
@@ -313,22 +313,29 @@ export class OperationalCredentialsServer extends OperationalCredentialsBehavior
...
                 StatusCode.ConstraintError,
             );
         }
 
-        if (timedOp.hasRootCert) {
-            throw new StatusResponseError(
-                "Trusted root certificate added in this session which is now allowed for UpdateNOC.",
-                StatusCode.ConstraintError,
-            );
+        if (timedOp.rootCertSet) {
```
Apollon77, line 323: I need a full test execution log please because this check should not be triggered at all because who should set an trusted root before? I needs to understand why this happens because in fact this should not be triggered

packages/node/src/behaviors/operational-credentials/OperationalCredentialsServer.ts
Apollon77: So see above ... we need to debug what happens here and who sets this cert why? Then we can see if it is an issue on our side or in the test or where ever

Apollon77 (changes requested):

packages/node/src/behaviors/operational-credentials/OperationalCredentialsServer.ts
```diff
@@ -205,28 +205,28 @@ export class OperationalCredentialsServer extends OperationalCredentialsBehavior
...
                 statusCode: OperationalCredentials.NodeOperationalCertStatus.MissingCsr,
-                debugText: "CSR not found in failsafe context.",
+                debugText: "CSR not found in failsafe context",
             };
         }
 
         if (failsafeContext.forUpdateNoc) {
             throw new StatusResponseError(
-                `addNoc received after csr request was invoked for UpdateNOC.`,
+                `addNoc received after csr request was invoked for UpdateNOC`,
```
Apollon77, line 229:
```suggestion
                `AddNoc received after csr request was invoked for UpdateNOC`,
```

packages/node/src/behaviors/operational-credentials/OperationalCredentialsServer.ts
```diff
@@ -308,32 +308,39 @@ export class OperationalCredentialsServer extends OperationalCredentialsBehavior
...
-        if (timedOp.hasRootCert) {
-            throw new StatusResponseError(
-                "Trusted root certificate added in this session which is now allowed for UpdateNOC.",
-                StatusCode.ConstraintError,
-            );
+        if (timedOp.rootCertSet) {
+            return {
+                // TC_OPCREDS_3_4 expects MissingCsr here...  Previously we used InvalidNoc which seems like it makes
+                // more sense
+                statusCode: OperationalCredentials.NodeOperationalCertStatus.MissingCsr,
```
Apollon77, line 327: This is against Spec and against Test plan. So it is an SDK AND test error as it seems. Clarification ongoing

Apollon77 (changes requested):

packages/node/src/behaviors/operational-credentials/OperationalCredentialsServer.ts
lauckhart: OK, I modified to conform to specification.  See if it looks correct now, if so I'll handle in testing branch after merge.

Apollon77 (approved):

PR #1507, opened 2024-12-11
Synchronize node lifecycle transitions

lauckhart:
This places protections around node lifecycle transitions with many async steps.  Previously it was possible to trigger these simultaneously in ways that cause crashes due to interference between them.  They may still be invoked in parallel but will execute sequentially.  Protected methods are available for composing the operations when the mutex is already held.


packages/general/src/util/Mutex.ts
```diff
@@ -65,6 +65,21 @@ export class Mutex implements PromiseLike<unknown> {
         }
     }
 
+    /**
+     * Enqueue work with an awaitable result.
+     */
+    produce<T>(task: () => PromiseLike<T>, cancel?: () => void): Promise<T> {
```
Apollon77, line 71: Should we then kill PromiseQueue? ;-) because was/is exactly this :-)

Apollon77 (approved):

packages/general/src/util/Mutex.ts
lauckhart: Hmm, I was thinking intent of PromiseQueue is for it support multiple workers...  So not a mutex but a work queue.  Could maybe have a shared implementation since a work queue with one worker is effectively a mutex but I think it makes sense to have both primitives

Apollon77: Ahh ok , yes makes sense. then fine!

PR #1520, opened 2024-12-15
Update typedocs of SwitchServer

t0bst4r:
It looks like the typedocs were coopied from WindowCovering, but wasn't updated afterwards.


Apollon77 (approved):
Thanks

PR #1540, opened 2024-12-20
CHIP test harness

lauckhart:
This is an extension to @matter.js/testing that adds support for executing CHIP certification tests.  This differs from the native test infrastructure in @project-chip/connectedhomeip and @project-chip/certification-tool in that:

- It provides a unified interface for executing and managing tests regardless of underlying implementation
- It separates management of the test subject ("DUT" in CHIP parlance) from the test implementation
- The test controller executes within a JS VM with standard testing infrastructure where feasible
- The test controller may execute alongside JS test subjects in the same VM
- The test controller is a single component with no dependencies other than the subject and test implementations
- Tests are configured automatically as much as possible, including test selection via PICS and configuration via other metdata
- Other than a JS VM testing is platform independent

This commit adds the test harness itself but does not include logic specific to matter.js.

Includes:

- Management of primary container for executing tests
- Management of utility containers for Avahi and dbus containers
- Generic "test" and "subject" abstractions
- `Test` implementations for CHIP's "YAML" and "Python" test types that perform commissioning and test execution
- Generic "test descriptor" abstraction for metadata associated with a specific test
- Test descriptor generation for matter.js and CHIP tests
- Additional PICS file logic and CHIP PICS expression evaluation
- Adds "matter-test inspect" for inspecting the state of known tests
- Implements necessary back channels for tests, one based on XML RPC and one that uses UNIX FIFOs


Apollon77 (approved):
Awesome. Can not await to use

PR #1549, opened 2024-12-22
matter.js CHIP test implementations

lauckhart:
Implements CHIP tests for matter.js.  A few of these tests do not pass yet, noted with TODOs where applicable.  Tests run with "npm run ci-test", not during normal testing via "npm run test".

Includes changes to DUT implementations in chip-testing/ to make the test instances compatible with new infrastructure.  This code is shared with original method of CHIP testing so that may break temporarily.

Also has a few minor tweaks and fixes to help tests succeed.


packages/model/src/standard/elements/RelativeHumidityMeasurement.ts
```diff
@@ -23,7 +23,7 @@ export const RelativeHumidityMeasurement = Cluster(
     Attribute(
         {
             name: "MeasuredValue", id: 0x0, type: "uint16", access: "R V", conformance: "M",
-            constraint: "minMeasuredValue to maxMeasuredValue", quality: "X P",
+            constraint: "minMeasuredValue to maxMeasuredValuep", quality: "X P",
```
Apollon77, line 26: Is this right or typo?

packages/node/src/behaviors/level-control/LevelControlServer.ts
```diff
@@ -246,10 +246,14 @@ export class LevelControlServerLogic extends LevelControlLogicBase {
         withOnOff: boolean,
         options: TypeFromPartialBitSchema<typeof LevelControl.Options> = {},
     ) {
+        if (rate === 0) {
```
Apollon77, line 249: This change and the OnOff one is in fact 1.4 changes. Dahlias we move out here and do separate pr when we do 1.4?

Apollon77 (approved):

packages/model/src/standard/elements/RelativeHumidityMeasurement.ts
lauckhart: typo

packages/node/src/behaviors/level-control/LevelControlServer.ts
lauckhart: discussed in discord, leaving as is

PR #1573, opened 2024-12-27
Catch UDP errors that are thrown directly

Apollon77:


packages/nodejs/src/net/NodeJsUdpChannel.ts
```diff
@@ -132,18 +132,22 @@ export class NodeJsUdpChannel implements UdpChannel {
...
+                                ? new RetransmissionLimitReachedError(error.message)
+                                : new NetworkError(error.message);
+                        netError.stack = error.stack;
+                        reject(netError);
+                        return;
+                    }
+                    resolve();
+                });
+            } catch (error) {
+                reject(new NetworkError((error as Error).message));
```
lauckhart, line 149: Can you give more background on this? Wondering what kind of error can be thrown directly... And maybe this should use the same logic as we do in the error handler above to convert to NetworkError?

Apollon77: Sure, it seems when DGram is already stopped then send directly throws an error and do not invoke the callback as used for "normal" errors, that's why the additional try-catch to get these kind of errors also to reject.

In fact it is about this error ... I see that sometimes when stopping with bad timing:

```
2024-12-21 12:57:29.423 - debug: matter.0 (2638) 2024-12-21 12:57:29.419 INFO ServerSubscription Error sending subscription update message (error count=1): Not running
at healthCheck (node:dgram:922:11)
at Socket.send (node:dgram:649:3)
at /opt/iobroker/node_modules/@matter/nodejs/src/net/NodeJsUdpChannel.ts:135:25
at new Promise ()
at NodeJsUdpChannel.send (/opt/iobroker/node_modules/@matter/nodejs/src/net/NodeJsUdpChannel.ts:134:16)
at UdpConnection.send (/opt/iobroker/node_modules/@matter/general/src/net/UdpInterface.ts:65:28)
at MessageChannel.send (/opt/iobroker/node_modules/@matter/protocol/src/protocol/ExchangeManager.ts:83:29)
at MessageExchange.send (/opt/iobroker/node_modules/@matter/protocol/src/protocol/MessageExchange.ts:425:28)
at processTicksAndRejections (node:internal/process/task_queues:95:5)
at InteractionServerMessenger.sendDataReportMessage (/opt/iobroker/node_modules/@matter/protocol/src/interaction/InteractionMessenger.ts:359:13)
```

Apollon77: Ok, what about this way? :-)

lauckhart (approved):

PR #1583, opened 2024-12-29
Add expression support to constraint AST

lauckhart:
Supports the very limited dialect required by spec as of 1.4.  Requires parser update for actual use.

### DIFF
--- a/packages/model/src/aspects/Constraint.ts
+++ b/packages/model/src/aspects/Constraint.ts
@@ -4,7 +4,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import { camelize, isObject } from "@matter/general";
+import { camelize } from "@matter/general";
 import { FieldValue } from "../common/index.js";
 import { Aspect } from "./Aspect.js";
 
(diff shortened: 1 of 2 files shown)


packages/model/src/aspects/Constraint.ts
```diff
@@ -89,12 +89,41 @@ export class Constraint extends Aspect<Constraint.Definition> implements Constra
...
+                            const lhs = valueOf(value.lhs);
+                            const rhs = valueOf(value.rhs);
+                            if (typeof lhs === "number" && typeof rhs === "number") {
+                                return lhs - rhs;
+                            }
+                            return undefined;
+                        }
+                        break;
+                }
                 if (type === FieldValue.reference && typeof name === "string") {
```
Apollon77, line 127: Is this supposed to be removed because now duplicated?

Apollon77 (approved):

packages/model/src/aspects/Constraint.ts
lauckhart: Well yes, it should have been removed because it *should* have been a duplicate, but there was a bug in the version above.  Fixed now

PR #1586, opened 2024-12-30
Shutdown and promise optimizations

Apollon77:
This PR optimized the node shutdown and some promises

If mainly makes sure that the network is shut down after sessions are closed and subscriptions flushed.

### DIFF
--- a/packages/protocol/src/mdns/MdnsServer.ts
+++ b/packages/protocol/src/mdns/MdnsServer.ts
@@ -15,6 +15,7 @@ import {
     DnsRecordType,
     isDeepEqual,
     Logger,
+    MatterAggregateError,
     MAX_MDNS_MESSAGE_SIZE,
     Network,
     Time,
@@ -252,7 +253,7 @@ export class MdnsServer {
     }
 
     async announce(announcedNetPort?: number) {
-        await Promise.allSettled(
+        await MatterAggregateError.allSettled(
             (await this.#getMulticastInterfacesForAnnounce()).map(async ({ name: netInterface }) => {
                 const records = await this.#records.get(netInterface);
                 for (const [portType, portTypeRecords] of records) {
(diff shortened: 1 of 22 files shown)


packages/protocol/src/mdns/MdnsServer.ts
```diff
@@ -252,7 +252,7 @@ export class MdnsServer {
     }
 
     async announce(announcedNetPort?: number) {
-        await Promise.all(
+        await Promise.allSettled(
```
lauckhart, line 255: This will hide errors, should it throw an aggregate error if there are failures?  Even if there shouldn't be errors I don't like the idea of them disappearing

packages/protocol/src/mdns/MdnsServer.ts
```diff
@@ -267,7 +267,7 @@ export class MdnsServer {
     }
 
     async expireAnnouncements(announcedNetPort?: number, type?: AnnouncementType) {
-        await Promise.all(
+        await Promise.allSettled(
```
lauckhart, line 270: same here

packages/protocol/src/session/SessionManager.ts
```diff
@@ -533,13 +533,14 @@ export class SessionManager {
...
+        const closePromises = this.#sessions.map(async session => {
             await session?.end(false);
             this.#sessions.delete(session);
-        }
+        });
         for (const session of this.#insecureSessions.values()) {
-            await session?.end();
+            closePromises.push(session?.end());
         }
+        await Promise.allSettled(closePromises);
```
lauckhart, line 543: Errors on close seem less problematic but would probably be worth handling here too

lauckhart (approved):

packages/protocol/src/mdns/MdnsServer.ts
lauckhart: Maybe we should have a utility so we can do e.g. `await MatterAggregateError.throwSettlementErrors(...)` or something

lauckhart (approved):

PR #1600, opened 2024-12-31
Full scope search for named references

lauckhart:
Adds support for default values that are named references.  For any property with a default value that is a named reference, a managed struct will return the referenced value until set explicitly.

Adds logic that searches for properties by name in any owner of a managed value.  This allows conformance expressions, constraints and default values to reference values by name so long as they are visible in any object in the ownership hierarchy.  Previously we only supported referencing values that were siblings.

I believe this is the last feature we need to support 100% of the data model semantics used in the Matter specifications (as of Matter 1.4).

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -30,8 +30,6 @@
                 "extendDefaults": true
             }
         ],
-        // turn on errors for missing imports
-        "import/no-unresolved": "error",
 
         "@typescript-eslint/require-await": "off",
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ The main work (all changes without a GitHub username in brackets in the below li
 ## __WORK IN PROGRESS__
 
 -   @matter/node
+    - Feature: Constraint and conformance expressions may now reference values by name in any owner of a constrained value
     - Enhancement: Each new PASE session now automatically arms the failsafe timer for 60s as required by specs
     - Enhancement: Optimizes Node shutdown logic to close sessions and subscriptions before shutting down the network
     - Fix: Fixes withBehaviors() method on endpoints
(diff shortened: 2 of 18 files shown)


.eslintrc.json
```diff
@@ -30,8 +30,6 @@
                 "extendDefaults": true
             }
         ],
-        // turn on errors for missing imports
-        "import/no-unresolved": "error",
```
Apollon77, line 34: Why you changed this?

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ The main work (all changes without a GitHub username in brackets in the below li
 ## __WORK IN PROGRESS__
 
 -   @matter/node
+    - Feature: Constraint and conformance expressions may now reference values by name in any owner of a constrained value
```
Apollon77, line 15: This also adds new validations to existing 1.3 cases right? If yes we might add a "potentially breaking" to description because we could validate stuff that was not validated before which could break existing code. So just text adjustment here in changelog.

Apollon77 (approved):

.eslintrc.json
lauckhart: It's a duplicate key, turned off lower in the file. I believe it's then off by default but I left the "off" version in to preserve the comment

CHANGELOG.md
lauckhart: No, it's not used anywhere for conformance or constraints, only for default values.

It's possible there are a couple of usages that we've deleted with overrides but I didn't look for them yet.  Was planning on checking as part of 1.4 work.  It's also possible there are places we can use in overrides even though they aren't in the spec formally.

But I mainly updated conformance and constraints for completeness since I had to do the hard part anyway for default values.

Even for defaults it's barely used, but I'm close to zero validation errors (after this here will only be 2 left in 1.4 branch).

PR #1606, opened 2025-01-05
Eslint upgrade

Apollon77:
This PR upgrades eslint to v9 including config and relevant deps.

It also addresses all new eslint errors reported by this versions. Please check the ocmmits

It also includes upgrade of @noble/curves which changed types and required an adjustment to access P256 curve "p" parameter ... I validated that the new place returns the same value as needed.

@lauckhart Please especially check https://github.com/project-chip/matter.js/pull/1606/commits/4c8f050b7d72dd7211ed3217954c0e7107a3f16e that in fact the new behavior is like the old one ... tests are all passing


packages/nodejs-ble/src/BlenoBleServer.ts
```diff
@@ -64,15 +64,17 @@ function initializeBleno(server: BlenoBleServer, hciId?: number) {
...
+            server
+                .handleC2SubscribeRequest(maxValueSize, updateValueCallback)
+                .catch(e => logger.warn("Error happened in when handling C2 subscribe", e));
         }
 
-        override async onUnsubscribe() {
+        override onUnsubscribe() {
             logger.debug("C2 unsubscribe");
-            await server.close();
+            server.close().catch(e => logger.warn("Error happened in when closing server for C2 unsubscribe", e));
```
lauckhart, line 77: Typo "in when"

codegen/src/clusters/NumberConstants.ts
```diff
@@ -21,8 +21,28 @@ import {
...
+    Model,
+    nodeId,
+    percent,
+    percent100ths,
+    posixMs,
+    subjectId,
+    systimeMs,
+    systimeUs,
+    vendorId,
+} from "#model";
```
lauckhart, line 45: Is this a bug in eslint? Or is there a reason we can't import default to access names exports?

packages/general/src/util/Error.ts
```diff
@@ -12,6 +12,13 @@ function considerAsError(error: unknown): error is Error {
     return (error as Error).message !== undefined;
 }
 
+export function asError(e: any, prefixText?: string): Error {
```
lauckhart, line 15: I think the prefix text here is confusing since it only applies if the cause is not an error.  Either the caller wants to add a prefix, in which case we should do it unconditionally, or it gets added at a higher level, in which case we should leave the message as whatever non-error was thrown

So I'd say either take it out or make it apply unconditionally

packages/general/src/util/Promises.ts
```diff
@@ -55,7 +56,7 @@ export function anyPromise<T>(promises: ((() => Promise<T>) | Promise<T>)[]): Pr
                 .catch(reason => {
                     numberRejected++;
                     if (!wasResolved && numberRejected === promises.length) {
-                        reject(reason);
+                        reject(asError(reason));
```
lauckhart, line 59: For some of these low-level utility classes perhaps we should leave the error alone... So add ignore eslint rule...

My thinking is that a.) the caller will need to do this check anyway, b.) the stack will be more useful if the caller adds it, and c.) a library user that for some reason throws non-errors will not expect the type of the throw to change

packages/general/src/util/Observable.ts
```diff
@@ -184,11 +185,7 @@ export class BasicObservable<T extends any[] = any[], R = void> implements Obser
                 try {
                     result = observer(...payload);
                 } catch (e) {
-                    if (e instanceof Error) {
-                        this.#errorHandler(e, observer);
-                    } else {
-                        this.#errorHandler(new Error(`${e}`), observer);
-                    }
+                    this.#errorHandler(asError(e, "Error processing Observables"), observer);
```
lauckhart, line 188: I don't think there's should be a prefix here, rather just assume that if someone threw a non-error the value will be meaningful to them

packages/general/src/util/Error.ts
```diff
@@ -12,6 +12,13 @@ function considerAsError(error: unknown): error is Error {
...
+    if (e instanceof Error) {
```
lauckhart, line 16: It occurs to me that if an error is thrown from another vm context, `instanceof` won't work... Also this doesn't really respect duck typing...

So probably we should accept any value with a `message` property that is a string and cast to Error.

Otherwise toString() might convert useful error information to "[object Object]"

Mainly just mention because this is so low level and fits to do it right

lauckhart (dismissed):

codegen/src/clusters/NumberConstants.ts
Apollon77: It seems that eslint can not "see" the names with multiple * exports are imported. So he says that the member is not in the imported namespace. I could revert and disable this one rule?

But because you access each single element alone anyway I thought it is equivalent

packages/general/src/util/Promises.ts
Apollon77: Alright. I then disable the eslint check in these cases.

packages/general/src/util/Error.ts
Apollon77: removed

packages/general/src/util/Error.ts
Apollon77: adjusted

codegen/src/clusters/NumberConstants.ts
lauckhart: It is equivalent, a little verbose but it's fine. But I would disable that rule regardless, we don't need eslint to be a second buggy type checker 🙂

lauckhart (approved):

codegen/src/clusters/NumberConstants.ts
Apollon77: lets disable when we stumble again over it

PR #1615, opened 2025-01-08
Add instance specific MDNS expires

Apollon77:
This PR mainly adds support to selectively expire MDNS records for just specific operational instances when a fabric gets removed.
It also fixes a special case to ensure a advertisement window is not restarted when  a window is opened during fabric removal.

It contains 2 more mini logging tweaks


packages/general/src/util/DeepEqual.ts
```diff
@@ -6,7 +6,7 @@
 
 // TODO - implement more efficient specialization for Array and ArrayBuffer
 // TODO - currently will hang on self-referential data structures
-export function isDeepEqual(a: any, b: any) {
+export function isDeepEqual(a: any, b: any, ignoreUndefiendProperties = false): boolean {
```
lauckhart, line 9: undefiened propertys :)

lauckhart (approved):

PR #1617, opened 2025-01-10
Stores MDNS host information for faster reuse with new SRV announcements

Apollon77:
Host records were not persisted when they were used already or such, so we needed to re-ask them. in commissioning processes the hostname likely stays the same when the device switches from commissioning to operative mode, so we could reuse the host information easiely.

The PR also adds expiry logic for active query data and details and for these stored data.

it also adds a bit more logging to be likely removed before we publish 0.12 but allows better validation of the fix with the affected users.

### DIFF
--- a/packages/protocol/src/mdns/MdnsScanner.ts
+++ b/packages/protocol/src/mdns/MdnsScanner.ts
@@ -70,10 +70,16 @@ type OperationalDeviceRecordWithExpire = Omit<OperationalDevice, "addresses"> &
         addresses: Map<string, MatterServerRecordWithExpire>; // Override addresses type to include expiration
     };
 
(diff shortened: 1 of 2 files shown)


packages/protocol/src/mdns/MdnsScanner.ts
```diff
@@ -103,6 +109,7 @@ export class MdnsScanner implements Scanner {
     }
 
     readonly #activeAnnounceQueries = new Map<string, { queries: DnsQuery[]; answers: StructuredDnsAnswers }>();
+    readonly #discoveredIpRecords = new Map<string, StructuredDnsAddressAnswers>();
```
lauckhart, line 112: Comments on these guys that describe the key as interface name would be helpful

lauckhart (approved):

PR #1628, opened 2025-01-13
Add installation notes for required packages when installing noble and bleno.

windhorn:
This library was required when trying to install bleno and noble in an Ubuntu Server 24.04.1 environment.
It was not mentioned in README.md, so I added it.


Apollon77 (approved):
Thank you

PR #1658, opened 2025-01-24
Trigger event listeners before attribute listeners for Subscriptions

Apollon77:
In fact happened events usually lead to a new "final attribute status" which is also reported. To mimic this a bit better we should trigger event listeners before attribute listeners. This could halp a lot with e.g. Switch events where events report presses, but the switch position is reported as "0".


CHANGELOG.md
```diff
@@ -9,6 +9,10 @@ The main work (all changes without a GitHub username in brackets in the below li
 	## __WORK IN PROGRESS__
 -->
 
+## __WORK IN PROGRESS__
+-   @matter/protocol
```
Apollon77, line 13: Open a dummy comment ... I send this to the current Controller implementors to request feedback ...

> Do you have any opinion if this is breaking or makes sense? I found it problematic getting the switch button events but then the "currentPosition" was reported as "0" before these events. if you consider breaking I would make it configurable ... else just is different behavior. WDYT?

lauckhart (approved):

PR #1677, opened 2025-02-02
Minor log fixes

kennylevinsen:
1. IPv6 addresses in a URL should be printed in brackets to separate IP from port.
2. An empty MatterError message in ServerSubscription left a confusing blank log message, so fall back to printing the whole error.


Apollon77 (approved):

Apollon77:
Thank you!

Issue #1678, opened 2025-02-02
TypeError in example DeviceNode.ts


rolfscherer:
Hello matter.js team,

First of all, thank you for your great work.

When I try to create a bundle for a project using esbuild, I encounter the following error:

```
TypeError: Cannot read properties of undefined (reading 'K')
    at Object.n (/home/rolf/matter/matter.js/packages/model/src/models/Aspects.ts:36:54)
    at ve.set quality (/home/rolf/matter/matter.js/packages/model/src/models/ClusterModel.ts:50:17)
    at new e (/home/rolf/matter/matter.js/packages/model/src/models/Model.ts:476:32)
    at new Yl (/home/rolf/matter/matter.js/packages/model/src/models/ScopeModel.ts:18:8)
    at new ve (/home/rolf/matter/matter.js/packages/model/src/models/ClusterModel.ts:166:9)
    at Object.<anonymous> (/home/rolf/matter/matter.js/packages/model/src/standard/elements/models.ts:122:36)
```
This can also be reproduced with a sample program with the following commands (in the example directory):

```
npm run bundle-device
npm run matter-device-bundled
```

The behavior is consistent on both Windows and Linux.
Is there a possibility that my installation is faulty?
I have tried this with version "@matter/main": "0.12.2" as well as with the main branch.
Thanks for your feedback.

Best regards, Rolf

rolfscherer:
I figured it out. You still need to add the --keep-names parameter to esbuild.

`"bundle-device": "esbuild src/device-onoff/DeviceNode.ts --bundle --platform=node --conditions=esbuild --external:@stoprocent/bleno --external:@stoprocent/bluetooth-hci-socket --sourcemap --minify --keep-names --outfile=build/bundle/DeviceNode.cjs",`

Or with a config file:

`
import { build } from 'esbuild';

build({
  entryPoints: ['src/server.ts'],
  bundle: true,
  platform: 'node',
  conditions: ['esbuild'],
  external: ['@stoprocent/bleno', '@stoprocent/bluetooth-hci-socket'],
  sourcemap: true,
  minify: true,
  keepNames: true,
  outfile: 'build/bundle/server.cjs',
}).catch(() => process.exit(1));

`

Apollon77:
@rolfscherer thank you for figuring this out. Can you do a PR or should I adopt into a PR?

rolfscherer:
PR added https://github.com/project-chip/matter.js/pull/1679

[...]

PR #1679, opened 2025-02-02
Added the --keep-names flag in esbuild.

rolfscherer:
Added the --keep-names flag in esbuild script in package.json
See also the issue https://github.com/project-chip/matter.js/issues/1678


lauckhart (approved):

lauckhart:
Thanks!

PR #1683, opened 2025-02-05
Minor ServerSubscription.flush cleanup

kennylevinsen:
The caller of flush always awaits the `currentUpdatePromise` if one is set, so `flush` doesn't need to do this and does in turn not need to be async.

### DIFF
--- a/packages/protocol/src/interaction/ServerSubscription.ts
+++ b/packages/protocol/src/interaction/ServerSubscription.ts
@@ -878,7 +878,7 @@ export class ServerSubscription extends Subscription {
         }
     }
 
-    async flush() {
+    async #flush() {
         this.#sendDelayTimer.stop();
         if (this.#outstandingAttributeUpdates.size > 0 || this.#outstandingEventUpdates.size > 0) {
             logger.debug(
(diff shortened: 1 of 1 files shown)


Apollon77:
Thank you very much for digging into it! The "real" fix of the root cause is already included in #1682 ... Flush needs to await when there is any update in progress. The problem was a cyclic await when the session was destroyed and ended (not even flushed) the subscription where it was also awaited

Apollon77:
it is not needingly redundant .. flush is only awaiting the "own" sending promise if there is anything left to send. but if there is a send currently running but no new data were added then this need to be awaited in close.

Apollon77:
The "currentpromise" variable gets cleared when promise resolves this means that there would be no second await because the currentPromise is undefined so the "if" guards this. 

Yes maybe not the most clear code flow ;-) Thanks for digging into this!

kennylevinsen:
> would be no second await because the currentPromise is undefined so the "if" guards this.

That's why its redundant (only one of the two blocks in immediate succession can ever do something, and the latter alone would have catch all the needed cases), but it's fine. Code is subjective, and maintainers decide the style.

Cheers.

PR #1687, opened 2025-02-07
0.13: Matter 1.4 Protocol

Apollon77:
This PR contains the protocol level changes for Matter 1.4 which is mainly session parameters and Discovery capabilities.

@lauckhart please do not approve right now but feel free to review and post LGTM or such ... we should merge together with your PR

We likely need a third one then afterwards with the application cluster relevant changes


packages/matter.js/src/CommissioningServer.ts
```diff
@@ -768,7 +767,7 @@ export class CommissioningServer extends MatterNode {
      * Return the pairing information for the device
      */
     getPairingCode(
-        discoveryCapabilities?: TypeFromBitSchema<typeof DiscoveryCapabilitiesBitmap>,
+        discoveryCapabilities?: TypeFromPartialBitSchema<typeof DiscoveryCapabilitiesBitmap>,
```
Apollon77, line 770: Dummy comment to prevent merging

packages/protocol/src/mdns/MdnsScanner.ts
```diff
@@ -1319,6 +1319,9 @@ export class MdnsScanner implements Scanner {
         } else if (result.T === 1) {
```
lauckhart, line 1319: I think this is redundant with the next check now

packages/protocol/src/mdns/MdnsScanner.ts
```diff
@@ -1319,6 +1319,9 @@ export class MdnsScanner implements Scanner {
...
             // Value 1 is reserved and should be handled as 0 according to Matter spec
             result.T = 0; // TCP not supported
+        } else if (result.T !== 2 && result.T !== 4 && result.T !== 6) {
```
lauckhart, line 1322: I'd write `!(result.T & 6)`

Apollon77: I think the best is to simply clear bit 0 always, then yes that all should be the same

Apollon77: so likely 

```!((result.T & ~1) & 6) ```

Apollon77: see commit

packages/protocol/src/mdns/MdnsScanner.ts
Apollon77: removed

lauckhart (approved):

PR #1699, opened 2025-02-10
Node read optimization

lauckhart:
This offers a new path for performing attribute reads against the Node API.  As discussed it is NOT fully wired yet so we can discuss first, but this PR should be ready for merge.

There's significant new code here.  Some of this is noise, which I detail below, but here are the main bits of new functionality to focus on:

1. packages/protocol/src/action/** - this is a new API for performing interactions that is designed for simplicity and performance:
    - action/protocols.ts - these define a `ProtocolNode` JS contract for exposing a Matter data model in a form that efficiently maps to the wire protocol
    - action/server/AttributeResponse.ts - implements the attribute subset of a Read Request Action against ProtocolNode
    - action/Interactable.ts and request/* - an API I created to simplify construction of Matter actions.  The goal was to make it easy to initiate Matter interactions for e.g. batch purposes, but with a more convenient JS API.  Much of this is stubbed due to focus on read in this commit but I do use some of the type definitions
3. packages/node/src/node/server/ProtocolService.ts - implements ProtocolNode for a @matter/node Node
4. packages/behavior/state/managed/values/StructManager.ts - I updated struct manager to generate ID-based getters and setters.  This means that where state values are associated with a Matter element that has an ID you can use `state[id]` and `state.name` interchangeably.  I had done this to support attributes that are not in the schema but it also allows us to map protocol queries against state objects without performing name lookup

It became apparent as I was working through this that the code was going to be too @matter/node specific unless I performed some code reorganization.  Unfortunately that adds noise to the commit but I think it's worthwhile.

Moved modules include:

- node/src/behavior/state/transaction -> general/src/transaction
- node/src/behavior/state/Val.ts -> protocol/src/action
- node/src/behavior/errors.ts -> protocol/src/action
- node/src/behavior/AccessControl.ts -> protocol/src/action/server

Also unfortunate... I encountered a Mocha bug while developing this that indicated there were import cycles.  I added a "matter-build cycles" command to detect cycles and fixed all cycles in the node package, but this ended up changing a bunch of extra files that are now hard to disentangle from the commit.


packages/node/src/node/server/TransactionalInteractionServer.ts
```diff
@@ -396,7 +366,12 @@ export class TransactionalInteractionServer extends InteractionServer {
         timed = false,
     ) {
         const invokeCommand = (context: ActionContext) => {
-            if (!context.authorizedFor(command.invokeAcl, { cluster: path.clusterId } as AccessControl.Location)) {
+            if (
+                context.authorityAt(command.invokeAcl, {
+                    endpoint: EndpointNumber(1),
```
Apollon77, line 371: I think thats reason for test fail?

packages/protocol/src/action/request/Invoke.ts
```diff
@@ -0,0 +1,93 @@
...
+ */
+export function Invoke(definition: Invoke.Definition): Invoke {
+    let { interactionModelRevision, suppressResponse, timed: timedRequest } = definition;
+    const { commands } = definition;
+
+    if (!commands?.length) {
+        throw new MalformedRequestError(`Invocation requires at least one command`);
+    }
+
+    interactionModelRevision ??= FALLBACK_INTERACTIONMODEL_REVISION;
```
Apollon77, line 27: you could add setting the fallback values in the deconstruction above then they could stay const

packages/protocol/src/action/request/MalformedRequestError.ts
```diff
@@ -0,0 +1,12 @@
...
+ * Copyright 2022-2025 Matter.js Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { MatterError } from "#general";
+
+/**
+ * Thrown when an action request does not adhere to the Matter specification.
+ */
+export class MalformedRequestError extends MatterError {}
```
Apollon77, line 12: Directly map to StatusResponseError for simplicity or do you plan error type conversions?

packages/protocol/src/action/request/Read.ts
```diff
@@ -0,0 +1,356 @@
...
+
+    /**
+     * Update "real" ReadRequest fields from our convenience attribute "selector".
+     */
+    function reifyAttributeSelector(selector: Read.AttributeSelector) {
+        const cluster = Specifier.clusterOf(selector);
+        const { endpoint } = selector;
+
+        // Install data version filter if the endpoint reports it has complete version information
+        if (typeof endpoint === "object" && typeof endpoint.versions === "object" && typeof cluster === "object") {
```
Apollon77, line 105: can that be null? (endpoint or cluster)? pot is the isObject tooling method

packages/node/src/node/server/TransactionalInteractionServer.ts
lauckhart: Umm, yes, seems likely

packages/protocol/src/action/request/MalformedRequestError.ts
lauckhart: This is a client-side error, it's thrown by the builder API if there's something illegal we can't prevent with types.  I updated the comment and made it extend `ImplementationError` instead

packages/protocol/src/action/request/Read.ts
lauckhart: Well, they can't be null according to the types.  Might be good to validate JS clients too... I tweaked so it won't fail here on null but it might elsewhere.

I think we need to revisit `isObject()` because it fails to detect any object with a `toString()` method, so would likely fail here if endpoint is a @matter/node Endpoint.

packages/protocol/src/action/request/Invoke.ts
Apollon77: See discord. Your adjustment is not what I meaned and will not work. Please adjust again.

packages/protocol/src/action/request/Read.ts
Apollon77: Hmmmm.... need to check that but ok.

PR #1711, opened 2025-02-16
ClientNode implementation round 3

lauckhart:
Protocol-level plumbing factors out various bits of controller- and client-related logic into more composable APIs:

  - Decouples PeerSet from InteractionClient and adds a (mostly stubbed) client Interactable
  - Modifies client networking to expose incoming data reports as a stream
  - Exposes physical device properties and subscription client as independent APIs
  - Extracts interface and logic to support working with partially-decoded data reports on the client
  - Adds hooks to InteractionServer to support delegation of subscription responses to a client implementation

Also fleshes out operational networking for client nodes.  Operational discovery is in place but actual interaction is blocked pending client Interactable implementation.


packages/protocol/src/interaction/InteractionMessenger.ts
```diff
@@ -713,56 +706,77 @@ export class IncomingInteractionClientMessenger extends InteractionMessenger {
...
+
+        if (result === undefined) {
+            // readDataReports should have thrown
+            throw new InternalError("No data reports loaded during read");
+        }
+
+        return result;
+    }
+
+    async *readDataReports() {
```
Apollon77, line 741: Please add a clear note as JSDoc that this is returning the DataReport content as by message content and that the last attribute in there could be still chunked over into the next message if it is a list ... We also need to consider this when we really "stream updates", unless we really want to emit this "partially" with replaces and adds

packages/protocol/src/interaction/InteractionServer.ts
```diff
@@ -227,9 +227,11 @@ export interface InteractionContext {
  * Translates interactions from the Matter protocol to matter.js APIs.
  */
 export class InteractionServer implements ProtocolHandler, InteractionRecipient {
+    readonly id = INTERACTION_PROTOCOL_ID;
     #context: InteractionContext;
     #nextSubscriptionId = Crypto.getRandomUInt32();
     #isClosing = false;
+    #clientHandler?: ProtocolHandler;
```
Apollon77, line 234: see above if not "multi protocol" is a better and more modular approach

packages/protocol/src/interaction/InteractionServer.ts
```diff
@@ -256,13 +254,28 @@ export class InteractionServer implements ProtocolHandler, InteractionRecipient
...
-    async onNewExchange(exchange: MessageExchange) {
+    async onNewExchange(exchange: MessageExchange, message: Message) {
         // Note - changes here must be copied to TransactionalInteractionServer as it does not call super() to avoid
         // the stack frame
         if (this.#isClosing) return; // We are closing, ignore anything newly incoming
+
+        // An incoming data report as the first message is not a valid server operation.  We instead delegate to a
+        // client implementation if available
+        if (message.payloadHeader.messageType === MessageType.SubscribeRequest && this.#clientHandler) {
+            return this.#clientHandler.onNewExchange(exchange, message);
```
Apollon77, line 265: see above

packages/protocol/src/interaction/SubscriptionClient.ts
```diff
@@ -0,0 +1,107 @@
...
+    delete(id: number) {
+        this.#listeners.delete(id);
+        const timer = this.#timeouts.get(id);
+        if (timer !== undefined) {
+            timer.stop();
+            this.#timeouts.delete(id);
+        }
+    }
+
+    async onNewExchange(exchange: MessageExchange) {
```
Apollon77, line 80: do you miss the initial message here that is send in when coming from InteractionServer.call? (Or i missed something in the code)

packages/protocol/src/peer/InteractionQueue.ts
```diff
@@ -0,0 +1,22 @@
...
+ * Copyright 2022-2025 Matter.js Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { Environment, Environmental, PromiseQueue } from "@matter/general";
+
+const CONCURRENT_QUEUED_INTERACTIONS = 4;
+const INTERACTION_QUEUE_DELAY_MS = 100;
+
+export class InteractionQueue extends PromiseQueue {
```
Apollon77, line 12: maybe better name ClientInteractionQueue? maybe we somewhen also need one for server

packages/protocol/src/peer/PeerSet.ts
```diff
@@ -358,8 +354,11 @@ export class PeerSet implements ImmutableSet<OperationalPeer>, ObservableSet<Ope
             // This ends discovery without triggering promises
             mdnsScanner?.cancelOperationalDeviceDiscovery(this.#sessions.fabricFor(address), address.nodeId, false);
         }
-        this.#clients.forEach(client => client.close());
-        this.#clients.clear();
+
+        for (const { address } of this.#peers) {
+            await this.disconnect(address, false);
+        }
+
```
Apollon77, line 361: clearing this.#peers is included somewhere else?

packages/protocol/src/peer/PhysicalDeviceProperties.ts
```diff
@@ -0,0 +1,80 @@
...
+const logger = Logger.get("PhysicalDeviceProperties");
+
+const DEFAULT_SUBSCRIPTION_FLOOR_DEFAULT_S = 1;
+const DEFAULT_SUBSCRIPTION_FLOOR_ICD_S = 0;
+const DEFAULT_SUBSCRIPTION_CEILING_WIFI_S = 60;
+const DEFAULT_SUBSCRIPTION_CEILING_THREAD_S = 60;
+const DEFAULT_SUBSCRIPTION_CEILING_THREAD_SLEEPY_S = 180;
+const DEFAULT_SUBSCRIPTION_CEILING_BATTERY_POWERED_S = 600;
+
+export interface PhysicalDeviceProperties {
```
Apollon77, line 18: Imaybe PeerDevice instead PhysicalDevice? But I am sure you thought about that a lot :-) so also ok with Pysical

packages/protocol/src/peer/PhysicalDeviceProperties.ts
```diff
@@ -0,0 +1,80 @@
...
+            description = "Node";
+        }
+
+        const { isBatteryPowered, isIntermittentlyConnected, threadConnected, isThreadSleepyEndDevice } =
+            properties ?? {};
+
+        if (isIntermittentlyConnected) {
+            if (minIntervalFloorSeconds !== undefined && minIntervalFloorSeconds !== DEFAULT_SUBSCRIPTION_FLOOR_ICD_S) {
+                logger.info(
+                    `${description}: Overwriting minIntervalFloorSeconds for intermittently connected device to 0`,
```
Apollon77, line 53: i know, issue from before but aybe log the used constant DEFAULT_SUBSCRIPTION_FLOOR_ICD_S instead of 0 ... more future proof

packages/node/src/behavior/system/commissioning/CommissioningClient.ts
```diff
@@ -111,8 +112,17 @@ export class CommissioningClient extends Behavior {
             commissioningOptions.finalizeCommissioning = this.finalizeCommissioning.bind(this);
         }
 
-        const address = await commissioner.commission(commissioningOptions);
-        this.state.peerAddress = address;
+        const identityService = this.endpoint.env.get(IdentityService);
+        const address = identityService.assignNodeAddress(node, fabric.fabricIndex, options.nodeId);
+
+        try {
+            await commissioner.commission(commissioningOptions);
```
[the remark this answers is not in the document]
lauckhart: Looks that way, I moved address assignment up and used the assigned ID in `commissioningOptions`.

Keep in mind the operational stuff is not complete so some of this code is untested

packages/node/src/behavior/system/controller/ControllerBehavior.ts
```diff
@@ -97,6 +113,44 @@ export class ControllerBehavior extends Behavior {
...
+        this.env.delete(FabricAuthority);
+        this.env.delete(ScannerSet);
+    }
+
+    #nodeOnline() {
+        // Configure network connections
+        const netInterfaces = this.env.get(NetInterfaceSet);
+        const netTransports = this.env.get(TransportInterfaceSet);
+        for (const transport of netTransports) {
+            if ("openChannel" in transport) {
```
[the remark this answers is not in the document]
lauckhart: kk

packages/node/src/behavior/system/network/NetworkClient.ts
```diff
@@ -0,0 +1,91 @@
...
+
+    protected async startup() {
+        const { startupSubscription } = this.state;
+        if (startupSubscription === null) {
+            return;
+        }
+
+        const message = Subscribe({
+            isFabricFiltered: true,
+            minIntervalFloorSeconds: DEFAULT_MIN_INTERVAL_FLOOR_SECONDS,
```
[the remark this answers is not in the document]
lauckhart: OK

packages/node/src/node/server/TransactionalInteractionServer.ts
```diff
@@ -126,12 +127,18 @@ export class TransactionalInteractionServer extends InteractionServer {
...
-    override async onNewExchange(exchange: MessageExchange) {
+    override async onNewExchange(exchange: MessageExchange, message: Message) {
         // When closing, ignore anything newly incoming
         if (this.#newActivityBlocked || this.isClosing) {
             return;
         }
 
+        // An incoming data report as the first message is not a valid server operation.  We instead delegate to a
+        // client implementation if available
+        if (message.payloadHeader.messageType === MessageType.SubscribeRequest && this.clientHandler) {
```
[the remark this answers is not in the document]
lauckhart: Right, fixed

packages/protocol/src/interaction/InteractionClient.ts
```diff
@@ -320,7 +304,12 @@ export class InteractionClient {
             });
         }, executeQueued);
 
-        if (dataVersionFilters !== undefined && dataVersionFilters.length > 0 && enrichCachedAttributeData) {
+        if (
+            dataVersionFilters !== undefined &&
+            dataVersionFilters.length > 0 &&
+            enrichCachedAttributeData &&
+            result.attributeReports
```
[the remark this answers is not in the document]
lauckhart: OK, previously I had left `attributeReports` undefined in `DecodedDataReport` so this would have been incorrect.  But it's defined now so I think I just need to remove that test as it's a no-op.

packages/protocol/src/interaction/InteractionMessenger.ts
lauckhart: OK, but why here and not `readAggregateDataReport` (formally `readDataReports`)?  This is just an extraction to convert from arrays to async iterator, otherwise it works like it used to

packages/node/src/behavior/system/controller/ControllerBehavior.ts
```diff
@@ -97,6 +113,44 @@ export class ControllerBehavior extends Behavior {
...
+                netInterfaces.add(transport as NetInterface);
+            }
+        }
+        if (this.state.ble) {
+            netInterfaces.add(Ble.get().getBleCentralInterface());
+        }
+
+        // This is necessary to receive data reports for subscriptions
+        this.env.get(InteractionServer).clientHandler = this.env.get(SubscriptionClient);
```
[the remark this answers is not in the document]
lauckhart: It is not always in env but it is when the node is online and this will throw if something goes awry.  Is that a bad idea?

I'll respond to second question in InteractionServer.ts

packages/protocol/src/interaction/InteractionServer.ts
lauckhart: I'm already unsure the utility of the protocol ID map in ExchangeManager makes sense, so I actually lean in the other direction - remove it.

My thinking is that pretty much everywhere in the code we use it, the protocol ID is hard-coded.  It's not like this is a generic dispatch of messages from the wire based on protocol ID.  So if we know that a place in the code needs something to handle "secure channel protocol", could we configure with a "secure channel protocol handler" rather than a "bag of protocol handlers that may or may not include the one I need"?

In the case of `#clientHandler`, I was originally going to wire as you describe; make `onNewExchange` return `Promise<boolean>` or somesuch...  But then it became clear that there is exactly one message type that needs to be routed to a separate component.  So rather than make the abstraction I was already questioning even more complicated I figured I'd try just wiring together explicitly and see how that works...  Makes the logic explicit...  There's one thing that who handles these messages, and there's one type of message that he can may delegate to someone else.

Anyway I'm not too concerned either way but maybe we could address in a subsequent refactoring.

packages/protocol/src/interaction/InteractionServer.ts
lauckhart: going to close as dup

packages/protocol/src/interaction/SubscriptionClient.ts
lauckhart: You mean when it delegates data report messages?  Looks OK to me but I haven't tested that wiring so could be that I got something wrong

packages/protocol/src/peer/InteractionQueue.ts
lauckhart: It's in a directory called "peer" so is already associated logically with client implementation, but there's not really anything client-specific about the implementation so I'd tend to leave the name more generic until either there's a conflict, or move to a more general place if it becomes more generally applicable

packages/protocol/src/peer/PeerSet.ts
lauckhart: Might be, where do you see it?  If the peers are cleaned up elsewhere they shouldn't be in the PeerSet anymore so this doesn't preclude cleaning them up as part of another process.  This just ensures we don't leak any if we use in a context where that logic doesn't apply.

packages/protocol/src/peer/PhysicalDeviceProperties.ts
lauckhart: Yeah `PeerDeviceProperties` doesn't seem appropriate when this only includes a subset of information we know about a peer, and this is another case of "we only use for peers right now but nothing is really specific to client-side implementation".

packages/protocol/src/peer/PhysicalDeviceProperties.ts
lauckhart: Lol OK this feels a bit like GHA punishing me for touching files with regexps 🙂

packages/node/src/behavior/system/commissioning/CommissioningClient.ts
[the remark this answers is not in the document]
Apollon77: All fine. a TODO might be enough to mark still open work ... or handle it now. up to you and both fine. ut we should try to keep the intel

packages/protocol/src/interaction/InteractionMessenger.ts
Apollon77: readAggregateDataReport() always returns a full complete report because it collects all messages internally ... right?

packages/protocol/src/interaction/InteractionServer.ts
Apollon77: Thats in fact all the think about Server and client ... when bringing all together I also have no issue with an "SecureChannelProtocol" class which handles all types and just throw if e.g. it is a server one and no server is started and vise versa

packages/protocol/src/interaction/SubscriptionClient.ts
Apollon77: yes The delegationcode calls onNewExchange(exchange, message)

packages/protocol/src/peer/InteractionQueue.ts
Apollon77: kk

packages/protocol/src/peer/PeerSet.ts
Apollon77: I just asked because the old code cleared the set after having closed all entries, the new code does not

packages/protocol/src/peer/PhysicalDeviceProperties.ts
Apollon77: opk, then we leave as you proposed, fine

packages/protocol/src/peer/PhysicalDeviceProperties.ts
Apollon77: lol ... I just noticed it and ... :-) thanks a lot

packages/protocol/src/interaction/InteractionServer.ts
Apollon77: As said in Discord. We can move the idea to combine classes to the "kill legacy" step ... I think then it would be a good idea

packages/protocol/src/interaction/SubscriptionClient.ts
Apollon77: WOuld be cool if you could look , else it is again par of unused yet codeflow

Apollon77 (approved):
Ok, see open comments

packages/protocol/src/peer/PeerSet.ts
lauckhart: Ah, got it.  `this.#clients` was cleared previously but should have been removed as it's owned by `InteractionClientProvider` now.  PeerSet just creates an `ExchangeProvider` which encapsulates connection management.

packages/protocol/src/interaction/SubscriptionClient.ts
lauckhart: Oh, I see.  I provided the message because it's part of the type signature for `onNewExchange` now.  `InteractionServer` needs it to delegate to `clientHandler` but `SubscriptionClient` doesn't use it.

packages/node/src/behavior/system/controller/ControllerBehavior.ts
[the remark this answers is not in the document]
lauckhart: Oh.  Yes, there's always an `InteractionServer` when the node is online as currently written.  This would be how reverse commissioning works in the case of fabric sync.  I'm not sure if there are other uses cases but I wasn't planning on making it start conditionally, rather make the node act like any other node and prevent unauthorized access via ACLs.

LMK if you think that makes sense.

packages/protocol/src/peer/PeerSet.ts
lauckhart: The point of this BTW was to decouple `PeerSet` from `InteractionClient` so I could use an alternate interaction implementation.

packages/node/src/behavior/system/controller/ControllerBehavior.ts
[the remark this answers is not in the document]
Apollon77: alright ... ok for now ... we see when we finalize it - and yes in future there needs to be an interaction server ... all good

packages/protocol/src/interaction/SubscriptionClient.ts
Apollon77: LGTM after checking deeper

PR #1739, opened 2025-02-25
Wait for lock on subscriptions cancellation

Apollon77:


packages/node/src/behavior/system/subscription/SubscriptionBehavior.ts
```diff
@@ -177,17 +177,23 @@ export class SubscriptionBehavior extends Behavior {
...
             const subscriptionIndex = this.state.subscriptions.findIndex(({ subscriptionId }) => id === subscriptionId);
             if (subscriptionIndex !== -1) {
-                this.state.subscriptions.splice(subscriptionIndex, 1);
+                return this.#removeSubscriptionIndex(subscriptionIndex);
             }
         }
     }
 
+    async #removeSubscriptionIndex(index: number) {
+        await this.context.transaction.begin();
```
lauckhart, line 192: If this is working great, you may need to add the resource too to actually grab the lock. Something like `this.transaction.addResource(this)`

lauckhart (approved):

packages/node/src/behavior/system/subscription/SubscriptionBehavior.ts
Apollon77: adjusted

PR #1741, opened 2025-02-26
0.13: Remove Legacy Device building API

Apollon77:
This change removes the legacy device building API and related packages. The Controller API isd still available and exported, but all exports are limited to this.


.github/workflows/chip-tool-tests.yml
```diff
@@ -222,37 +222,6 @@ jobs:
               --tv-app ../chip-testing/dist/esm/TvTestApp.js \
             '
 
-  chip-tests-legacy:
```
Apollon77, line 225: Dummy comment to prevent merging

lauckhart (approved):

PR #1789, opened 2025-03-18
Adjust Logger methods

Apollon77:
mainly adding a new method which does not throw to detect more easiely if a Logger exists for a given identifier
@JimBuzbee

### DIFF
--- a/packages/general/src/log/Logger.ts
+++ b/packages/general/src/log/Logger.ts
@@ -131,10 +131,18 @@ export class Logger {
     }
 
     /**
+     * Check if a logger with the matching identifier exists.
+     * @param identifier The identifier of the logger
+     */
+    public static hasLoggerForIdentifier(identifier: string) {
+        return Logger.logger.some(logger => logger.logIdentifier === identifier);
+    }
+
+    /**
      * Get the logger with the matching identifier.
      * @param identifier The identifier of the logger
      */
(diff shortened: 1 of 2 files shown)


lauckhart (approved):

packages/general/src/log/Logger.ts
```diff
@@ -130,11 +130,19 @@ export class Logger {
...
+    public static hasLoggerForIdentifier(identifier: string) {
+        return Logger.logger.some(logger => logger.logIdentifier === identifier);
+    }
+
     /**
      * Get the logger with the matching identifier.
      * @param identifier The identifier of the logger
      */
-    public static getLoggerforIdentifier(identifier: string) {
+    public static getLoggerForIdentifier(identifier: string) {
```
lauckhart, line 145: Crappy gh didn't save my comment because I hit "approve" before "start a review", then crappy automerge (why?) merged but I was going to say that, since this is a breaking change already... ;)

Maybe we could move away from the c-like API for managing log destinations and get rid of the confusing overuse of the term "logger" (both the logging API and the logging back ends seem to be called loggers).

I'm thinking we could instead have an API that is `Logger.defaultDestination: Logger.Destination` and `Logger.destinations: Set<Logger>` instead.  That'd provide both mutability and introspection and get rid of the concept of identifiers altogether.

PR #1790, opened 2025-03-18
Improve async storage reliability

Apollon77:
This PR adjusts writing the storage files as temp file first and then do a rename. It also makes sure that all renames are finished when closing or clearing the storage.

I do this as Draft PR to get early review comments, but need to finish testing this in detail tomorrow

### DIFF
--- a/packages/nodejs/src/storage/StorageBackendDiskAsync.ts
+++ b/packages/nodejs/src/storage/StorageBackendDiskAsync.ts
@@ -163,6 +163,8 @@ export class StorageBackendDiskAsync extends MaybeAsyncStorage {
         const fd = await open(this.#path, "r");
         try {
             await fd.sync();
+        } catch (error) {
+            logger.warn(`Failed to fsync storage directory ${this.#path}`, error);
         } finally {
             await fd.close();
         }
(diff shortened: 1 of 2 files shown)


packages/nodejs/src/storage/StorageBackendDiskAsync.ts
```diff
@@ -134,14 +135,40 @@ export class StorageBackendDiskAsync extends MaybeAsyncStorage {
...
+     * fsync on a directory ensures there are no rename operations etc. which haven't been persisted to disk.
+     * We do this as best effort to ensure that all writes are persisted to disk.
+     */
+    async #fsyncStorageDir() {
+        const fd = await open(this.#path, "r");
+        try {
+            await fd.sync();
+        } catch (error) {
+            // Windows will cause `EPERM: operation not permitted, fsync`
+            // for directories, so lets catch this generically
```
lauckhart, line 163: Did you try on windows with "w" flag for opening?

Assuming that doesn't work, I think you could branch the code... For windows instead of opening directory, after rename open the new file in write mode, then sync that fd.

lauckhart (approved):

packages/nodejs/src/storage/StorageBackendDiskAsync.ts
Apollon77: Adjusted

Apollon77:
it in fact has a small performance decrease which is ok for devices because the persisted data are not that many and need to be reliable. Issue is more with Controller side when first connecting to a device and all data are first initialized. These writes take a biut longer but are also more reliable. we need to see how much it really affects it

Issue #1764, opened 2025-03-06
Matter 1.4 finalization topics


Apollon77:
- [ ] Q quality
- [ ] Address open comments from https://github.com/project-chip/matter.js/pull/1657
- [ ] Re-enable Test_TC_CC_3_2, Test_TC_CC_3_3, Test_TC_CC_4_3, Test_TC_CC_5_3, Test_TC_CC_6_3, Test_TC_CC_7_3

PR #1804, opened 2025-03-21
Add Quiter guards for PowerSource cluster

Apollon77:
addresses parts of #1764

### DIFF
--- a/packages/node/src/behaviors/power-source/PowerSourceServer.ts
+++ b/packages/node/src/behaviors/power-source/PowerSourceServer.ts
@@ -5,13 +5,38 @@
  */
 
 import { DescriptorServer } from "#behaviors/descriptor";
+import { ClusterType } from "@matter/types";
+import { PowerSource } from "@matter/types/clusters/power-source";
 import { PowerSourceBehavior } from "./PowerSourceBehavior.js";
+
+const PowerSourceLevelBase = PowerSourceBehavior.with(PowerSource.Feature.Battery, PowerSource.Feature.Rechargeable);
 
 /**
  * This is the default server implementation of {@link PowerSourceBehavior}.
  */
-export class PowerSourceServer extends PowerSourceBehavior {
+export class PowerSourceServerLogic extends PowerSourceLevelBase {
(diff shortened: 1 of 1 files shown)


packages/node/src/behaviors/power-source/PowerSourceServer.ts
```diff
@@ -5,13 +5,38 @@
...
         (await this.agent.load(DescriptorServer)).addDeviceTypes("PowerSource");
+
+        // According to specs changes to these attributes should not occur more often than every 10 seconds
+        if (this.events.batPercentRemaining$Changed !== undefined) {
+            this.events.batPercentRemaining$Changed.quiet.config = {
+                minimumEmitIntervalMs: 10_000,
+            };
+        }
+        if (this.events.batTimeRemaining$Changed !== undefined) {
+            this.events.batTimeRemaining$Changed.quiet.config = {
```
lauckhart, line 28: QuietObservable implements the config interface so you can do foo.quiet.minimumCommitIntervalMS = 10_000

packages/node/src/behaviors/power-source/PowerSourceServer.ts
```diff
@@ -5,13 +5,38 @@
...
 /**
  * This is the default server implementation of {@link PowerSourceBehavior}.
  */
-export class PowerSourceServer extends PowerSourceBehavior {
+export class PowerSourceServerLogic extends PowerSourceLevelBase {
     override async initialize() {
         (await this.agent.load(DescriptorServer)).addDeviceTypes("PowerSource");
+
+        // According to specs changes to these attributes should not occur more often than every 10 seconds
+        if (this.events.batPercentRemaining$Changed !== undefined) {
```
lauckhart, line 22: Maybe loop or function to make this less redundant?

lauckhart (approved):

packages/node/src/behaviors/power-source/PowerSourceServer.ts
Apollon77: Both adjusted

PR #1807, opened 2025-03-21
Cleanup for logging API

lauckhart:
Logically separates logging API (called "loggers") from log writers (formerly also called "loggers", now called "log destinations").

Log destinations are managed via a simple object API.  They include a new "add" method that by default formats then writes, but overrides can process `Diagnostic.Message` directly to skip matter.js formatting.

Deprecates some methods and adds compatibility layers as necessary for most user-facing API surface.


packages/general/src/log/LogDestination.ts
```diff
@@ -0,0 +1,113 @@
...
+}
+
+export namespace LogDestination {
+    /**
+     * Defaults for {@link LogDestination} fields.
+     */
+    export const defaults: LogDestination = {
+        name: "default",
+
+        level: LogLevel.DEBUG,
```
Apollon77, line 80: did we not had INFO as default before?

Apollon77 (approved):

packages/general/src/log/LogDestination.ts
lauckhart: No, if you provided an invalid log level to `LogLevel` previously it would select INFO but I made that throw instead.  The actual levels set in `Boot.init` (for default destinations) and in `addLogger` (for tertiary destinations) always defaulted to DEBUG unless explicitly overridden.

That said, we should probably change to INFO as the default soon.

PR #1819, opened 2025-03-24
Ignore missing IPv4 addresses if enabled

Apollon77:
Missing IPv4 addresses on network interfaces are now ignored even if IPv4 is not disabled via configuration

addresses https://github.com/Luligu/matterbridge/issues/244

### DIFF
--- a/packages/nodejs/src/net/NodeJsUdpChannel.ts
+++ b/packages/nodejs/src/net/NodeJsUdpChannel.ts
@@ -12,7 +12,7 @@ import {
     Logger,
     MAX_UDP_MESSAGE_SIZE,
     NetworkError,
-    NoIPv4AddressAvailableError,
+    NoAddressAvailableError,
     repackErrorAs,
     UdpChannel,
     UdpChannelOptions,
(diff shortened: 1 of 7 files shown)


packages/nodejs/src/net/NodeJsUdpChannel.ts
```diff
@@ -71,11 +72,11 @@ export class NodeJsUdpChannel implements UdpChannel {
             if (type === "udp4") {
                 multicastInterface = NodeJsNetwork.getMulticastInterfaceIpv4(netInterface);
                 if (multicastInterface === undefined) {
-                    throw new NetworkError(`No IPv4 addresses on interface: ${netInterface}`);
+                    throw new NoIPv4AddressAvailableError(`No IPv4 addresses on interface "${netInterface}"`);
                 }
             } else {
                 if (netInterfaceZone === undefined) {
-                    throw new NetworkError(`No IPv6 addresses on interface: ${netInterface}`);
+                    throw new NetworkError(`No IPv6 addresses on interface "${netInterface}", but IPv6 is required`);
```
lauckhart, line 79: I'd change the error to “NoAddressAvailable" and throw for ipv6 too.  And move the "but ipv6 is required" bit into matter-specific code

My reasoning is that there are benefits to keeping Matter semantics out of platform adapters as much as possible because:

1. This makes the adapters suitable for future use cases with different (such as non-matter) requirements
2. It reduces the complexity of the adapters, moving logic into platform-independent code where it doesn't need to be replicated for every platform

lauckhart (approved):

PR #1825, opened 2025-03-27
Remove OnOffServer "Lighting " feature enablement

Apollon77:
In fact the default enablement of Lighting makes issues with other device types where this cluster can be in. So better fix this special case now.

### DIFF
--- a/packages/node/src/behaviors/on-off/OnOffServer.ts
+++ b/packages/node/src/behaviors/on-off/OnOffServer.ts
@@ -9,24 +9,24 @@ import { GeneralDiagnostics } from "#clusters/general-diagnostics";
 import { OnOff } from "#clusters/on-off";
 import { RootEndpoint } from "#endpoints/root";
 import { MaybePromise, Time, Timer } from "#general";
+import { ClusterType } from "#types";
 import { OnOffBehavior } from "./OnOffBehavior.js";
 
-const Base = OnOffBehavior.with(OnOff.Feature.Lighting);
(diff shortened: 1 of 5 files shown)


packages/node/src/behaviors/on-off/OnOffServer.ts
```diff
@@ -9,24 +9,24 @@ import { GeneralDiagnostics } from "#clusters/general-diagnostics";
...
+ * override the methods you need to change or add mandatory commands.
  *
- * The "OffOnly" feature is automatically supported because the commands are disabled by conformance.
+ * The "OffOnly" and "Lighting" features are automatically supported because the commands are disabled by conformance.
  * The default implementation do not contain any logic for the DeadFrontBehavior feature because this is very use case
  * specific, so this needs to be implemented by the device implementor as needed.
  */
-export class OnOffServer extends Base {
-    declare protected internal: OnOffServer.Internal;
+export class OnOffServerLogic extends OnOffLogicBase {
```
lauckhart, line 28: The "Logic" suffix feels a little weird to me, though I know we use it in other clusters... What if we use the convention "CompleteOnOffServer", "OnOffBaseServer", "OnOffWithAllServer" or "FullFeaturedOnOffServer"?

They're all a little contorted but the "Logic" suffix makes it seem like the class is something other than a server.

lauckhart (approved):

packages/node/src/behaviors/on-off/OnOffServer.ts
Apollon77: Good topic. I will change that in a separate PR for all where we do it like that

PR #1827, opened 2025-03-27
Accept UDP traffic from unknown interfaces

Apollon77:
for Matter messages. Move the check to also consider traffic from known network interfaces for MDNS.

@digitaldan Wanne try it? :-)

For me locally in one network zone all works. What about your multi zone setup?


lauckhart (approved):

Apollon77:
I will merge after test info from @digitaldan

digitaldan:
Thanks for the quick fix!   Things are working now, tested on both Linux and Mac.

PR #1839, opened 2025-03-29
Revert ScenesManagement Override

Apollon77:
This Override is currently inactive till it is decided how names will be.

        In https://github.com/project-chip/connectedhomeip/pull/38003 SDK was adjusted to be like spec why this
        override is obsolete when it stays that way.

        Discussion is still open in see https://github.com/CHIP-Specifications/connectedhomeip-spec/issues/11386

So for now lets revertto "as chip does it" and see how it goes.

Reverting this fixes the daily interop tests

### DIFF
--- a/models/src/local/ScenesManagementOverrides.ts
+++ b/models/src/local/ScenesManagementOverrides.ts
@@ -11,6 +11,15 @@ LocalMatter.children.push({
     name: "ScenesManagement",
 
     children: [
+        /*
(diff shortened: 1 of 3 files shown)


lauckhart (dismissed):

Apollon77:
@lauckhart Please re-review. thanks

lauckhart (approved):

Issue #1844, opened 2025-03-30
ServerNode sometimes fails to close in v13


digitaldan:
Hi again ;-) 

Another issue i am encountering on the V13 builds is that trying to close a "ServerNode" can hang indefinitely , so for example:

```typescript
this.server = await ServerNode.create({....})
...add endpoints...
this.server.start();
```
and then later try and close it like:

`await this.server?.close();`

The close command will hang indefinitely and never completes.  The only logs that are printed are:

```logs
12:32:12.067 [DEBUG] [nternal.client.MatterWebsocketService] - DeviceNode: Closing bridge
12:32:12.067 [TRACE] [nternal.client.MatterWebsocketService] - none: 2025-03-30 12:32:12.067 NOTICE Node *bridge-0* going offline
12:32:12.067 [TRACE] [nternal.client.MatterWebsocketService] - OccurrenceManager: Recorded event #11003: {"eventId":1,"clusterId":40,"endpointId":0,"epochTimestamp":1743363132067,"priority":2,"payload":"undefined"}
12:32:12.080 [TRACE] [nternal.client.MatterWebsocketService] - DeviceCommissioner: Commissioning mode ended, stop announcements.
12:32:12.080 [TRACE] [nternal.client.MatterWebsocketService] - MdnsServer: Expiring records instanceName: 121C84BE4A3C83BC._matterc._udp.local port: 5540 netInterface: en0
12:32:12.080 [TRACE] [nternal.client.MatterWebsocketService] - MdnsServer: Expiring records instanceName: 121C84BE4A3C83BC._matterc._udp.local port: 5540 netInterface: en1
12:32:12.080 [TRACE] [nternal.client.MatterWebsocketService] - MdnsServer: Expiring records instanceName: 121C84BE4A3C83BC._matterc._udp.local port: 5540 netInterface: en8
12:32:12.080 [TRACE] [nternal.client.MatterWebsocketService] - MdnsServer: Expiring records instanceName: 121C84BE4A3C83BC._matterc._udp.local port: 5540 netInterface: awdl0
12:32:12.080 [TRACE] [nternal.client.MatterWebsocketService] - MdnsServer: Expiring records instanceName: 121C84BE4A3C83BC._matterc._udp.local port: 5540 netInterface: llw0
[...]

PR #1857, opened 2025-04-02
Add UDP send guard

Apollon77:
This UDP send gurt ensures that UDP send calls get rejected after maximum 1-2s when not returning from send call.

could address #1844

### DIFF
--- a/packages/nodejs/src/net/NodeJsUdpChannel.ts
+++ b/packages/nodejs/src/net/NodeJsUdpChannel.ts
@@ -6,6 +6,7 @@
 
 import {
     ChannelType,
+    createPromise,
     Diagnostic,
     isIPv4,
     isIPv6,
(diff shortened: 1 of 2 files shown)


packages/nodejs/src/net/NodeJsUdpChannel.ts
```diff
@@ -23,6 +25,8 @@ import { NodeJsNetwork } from "./NodeJsNetwork.js";
 
 const logger = Logger.get("NodejsChannel");
 
+const UDP_SEND_TIMEOUT_CHECK_INTERVAL_MS = 1_000; // UDP should be send out in some ms so if we needed 1s+, we have a problem
```
lauckhart, line 28: This is a pretty short timeout of DNS is involved in the send... I think that doesn't affect us currently but maybe this should be configurable?

Apollon77: We usually send to IPs directly and not to names, so should not be an issue? I can also go higher

PR #1866, opened 2025-04-03
Updates and fixes

lauckhart:
* Previously we were losing track of various advertisement promises either by leaving them with a non-async observable or by simply catching and logging errors.  We had a mutex to track promises but IIRC did not place all promises into it just to be conservative.  Now we do.  This should prevent spurious errors related to operating on closed advertisers that regularly cropped up in test logs.

* Changes how messages are logged when attribute reads fail.  Previously we logged all exceptions with a stack trace.  Now if it's a StatusResponseError we log the message but not the stack trace.  If it's any other type of error we do not log at all but rethrow (which we already did) an implementation error with the original error as the cause.

* Fix error messages and promise handling for async observers of non-async behavior events.  This was already implemented but the relevant context fields weren't set.

* The test suite now waits 5 seconds for the process to exit once testing is complete.  If the interval elapses and the process is still alive it exits with a message and error code.  This will make CI jobs fail faster when tests are unruly.

* Adjusted test failure formatting to suck less in GH CI

### DIFF
--- a/packages/protocol/src/interaction/InteractionServer.ts
+++ b/packages/protocol/src/interaction/InteractionServer.ts
@@ -7,6 +7,7 @@
 import {
     Crypto,
     Diagnostic,
+    ImplementationError,
     InternalError,
     Logger,
     MatterError,
@@ -52,7 +53,7 @@ import { Message, SessionType } from "../codec/MessageCodec.js";
 import { EndpointInterface } from "../endpoint/EndpointInterface.js";
 import { MessageExchange } from "../protocol/MessageExchange.js";
 import { ProtocolHandler } from "../protocol/ProtocolHandler.js";
-import { NoAssociatedFabricError, SecureSession, assertSecureSession } from "../session/SecureSession.js";
+import { assertSecureSession, NoAssociatedFabricError, SecureSession } from "../session/SecureSession.js";
 import {
     decodeAttributeValueWithSchema,
     decodeListAttributeValueWithSchema,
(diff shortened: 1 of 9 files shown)


packages/protocol/src/interaction/InteractionServer.ts
```diff
@@ -540,14 +541,15 @@ export class InteractionServer implements ProtocolHandler, InteractionRecipient
...
-                    logger.error(
-                        `Error while reading attribute from ${
-                            exchange.channel.name
-                        } to ${this.#endpointStructure.resolveAttributeName(path)}:`,
-                        error,
-                    );
+                    const what = `reading ${this.#endpointStructure.resolveAttributeName(path)} from ${exchange.channel.name}`;
+
+                    if (!(error instanceof StatusResponseError)) {
+                        const wrappedError = new ImplementationError(`Unhandled error ${what}`);
```
Apollon77, line 547: Would InternalError be better? ImplementationError was more thought like "hey dear developer, check your code". But when read fails it is more like "our side"?

packages/testing/src/cli.ts
```diff
@@ -174,6 +176,14 @@ export async function main(argv = process.argv) {
             console.log();
         }
     }
+
+    // Do not hang indefinitely if tests do not clean up after themselves properly.  Instead print error and exit with
+    // non-zero status code
+    const timeout = setTimeout(() => {
+        console.error(`Error: Tests passed but process did not exit cleanly after ${SHUTDOWN_TIMEOUT_MS / 1000}s.`);
+        process.exit(101);
```
Apollon77, line 184: Idea: could we trigger logging node diagnostics in such a case? to directly "see" whats up?

Apollon77 (approved):

packages/protocol/src/interaction/InteractionServer.ts
lauckhart: Yes maybe so... I used ImplementationError because there could in theory be errors from e.g. dynamic attributes, but does seem more likely is our fault.

Maybe we should add a new UnhandledError or something we can use in cases where we don't know whose fault the error is?

packages/testing/src/cli.ts
lauckhart: That would be good, I wanted to dump diagnostic information but I don't know of anything that would be useful, do you?

What we really need is wtfnode but that would require us to run all tests with wtf enabled. Which would maybe work I just worry about perf/stability because it's so invasive.

packages/protocol/src/interaction/InteractionServer.ts
lauckhart: Actually though I should trigger wtf conditionally if it's enabled via command line flags

lauckhart: Replied to wrong thread there 🙄

Apollon77: Ok, lets leave ImplementationError

packages/testing/src/cli.ts
Apollon77: I would log the same as SIGUSR2 would do for the start

lauckhart: Will address when it comes up next

PR #1869, opened 2025-04-04
Adjust Test name parsing

Apollon77:
... to allow also names like CADMIN_1_3_4.py that contain 2 tests

### DIFF
--- a/chip/support/generate-test-descriptor
+++ b/chip/support/generate-test-descriptor
@@ -47,13 +47,13 @@ tests = []
 def load_yaml():
     """Load metadata from YAML tests.
 
-    This is fairly straightforward as the tests are inherently declarative    
+    This is fairly straightforward as the tests are inherently declarative
     """
     categories = {}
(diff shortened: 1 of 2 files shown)


chip/support/generate-test-descriptor
```diff
@@ -229,7 +229,7 @@ load_python()
 shutil.rmtree("/tmp/metadata-garbage")
 
 # Normalize/prettify names and create semantic version for sorting
-semantic_name_pattern = re.compile(r"(.*)_(\d+)_(\d+)(?:_(.*))?")
+semantic_name_pattern = re.compile(r"(.*)_(\d+)_(\d+(?:_\d+))(?:_(.*))?")
```
lauckhart, line 232: Were you able to test this? That capturing group is parsed as an integer below, and I think the next one should retrieve this case as it accepts anything following “_“.

It may be that the structure of the test descriptor needs to change... I'm not sure if/how this script and/or the test harness handles multiple tests in one file

PR #1885, opened 2025-04-10
Fix for command handling in derived clusters

lauckhart:
* ClusterModel properties for attributes, commands, events and datatypes now include all members in scope

* CommandModel direction helpers isRequest and isResponse now use effective (scoped) direction

* Regenerate clusters and endpoints

* (unrelated) Group vscode launch configs


Apollon77 (approved):
Awesome, thank you!

PR #1915, opened 2025-04-17
Remove Sync Storage  and other dependency cleanups

Apollon77:
Sync storage was just used by legacy codepath that got removed by 0.13 anyway, so we can also cleanup the sync storage class, this also reduces production dependencies.

Additionally cleanup some @noble/curves dep definitions where not needed


CHANGELOG.md
```diff
@@ -34,6 +34,7 @@ The main work (all changes without a GitHub username in brackets in the below li
     - Feature: Automatically handle basicInformation uniqueId Property as defined by specification if not set by the developer
 
 -   @matter/nodejs
+    - Breaking: The StorageBackendDisk class got removed including the "node-localstorage" dependency. Use StorageBackendDiskAsync instead.
```
lauckhart, line 37: Can we rename StorageBackendDiskAsync->StorageBackendDisk and make StorageBackendDiskAsync a deprecated alias?

Apollon77: Alright I adjust that

PR #1945, opened 2025-04-23
Fix Wifi naming in Thread configuration output

markus-becker-tridonic-com:


Apollon77 (approved):
Hehe nice copy&paste error. Thanks for fixing!

PR #1987, opened 2025-05-05
Async crypto

lauckhart:
Converts the majority of crypto methods to be async.  This will be required for implementations such as web crypto.


Apollon77 (approved):

Apollon77:
@lauckhart seems there is another code change needed in the controller logic for the test: 2025-05-05 20:13:23.957 ERROR   20:13:23.929 - PAIR OUT  : [LegacyControllerCommandHandler] 2025-05-05 20:13:23.893 ERROR  LegacyCont~ndHandler root key pair is still initializing

PR #1997, opened 2025-05-08
Log RecoverableCommissionError as warning

spudwebb:


Apollon77 (approved):
Thanks. Yes makes sense

PR #2000, opened 2025-05-09
Subscription Wiring to Attribute/EventResponse

Apollon77:
This PR is finalizing the Read rewiring by also wiring Subscription logic and processing.

Details see commits


packages/protocol/src/action/protocols.ts
```diff
@@ -52,6 +53,14 @@ export interface NodeProtocol extends CollectionProtocol<EndpointProtocol> {
      */
     eventHandler: OccurrenceManager;
 
+    /**
+     * Event when data on the node changes.
+     */
+    stateChanged: Observable<[endpointId: EndpointNumber, clusterId: ClusterId, changes: AttributeId[]], MaybePromise>;
```
lauckhart, line 59: Crap, this breaks quieter events.

I think this is still a good optimization but we should probably have the ProtocolService identify Q events and split the logic... For Q, filter from the datasource and add a listener to the online observable.

On the plus side, good reason to room LVL and CC tests even if cluster didn't change... Ws could just break cc up until maybe three separate groups so they can run in parallel.

packages/protocol/src/action/protocols.ts
```diff
@@ -108,10 +122,20 @@ export interface ClusterTypeProtocol extends AddressableElementProtocol<ClusterI
      */
     attributes: CollectionProtocol<AttributeTypeProtocol>;
 
+    /**
+     * Map of attribute names to IDs.
+     */
+    attributeNameToId: Map<string, AttributeId>;
```
lauckhart, line 128: Is this used anywhere? If we need it we could add the name to the ClusterTypeProtocol

packages/protocol/src/action/protocols.ts
```diff
@@ -108,10 +122,20 @@ export interface ClusterTypeProtocol extends AddressableElementProtocol<ClusterI
...
+
     /**
      * Event metadata.
      */
     events: CollectionProtocol<EventTypeProtocol>;
+
+    /**
+     * Map of event names to IDs.
+     */
+    eventNameToId: Map<string, EventId>;
```
lauckhart, line 138: Maybe add a TODO noting that we need this temporarily while events come in through the OccurrenceManager.

Longer term I think we should convert OccurrenceManager to serialize IDs.

Then decouple OccurrenceManager from @matter/node... Instead have another Observable on NodeProtocol for events, and then have NodeProtocol pass them to OccurrenceManager.

That would simplify the contract between @matter/node and @matter/protocol and also be more efficient.

packages/protocol/src/action/server/FilteredAttributeResponse.ts
```diff
@@ -0,0 +1,90 @@
...
+};
+export type AttributeResponseFilter = {
+    [endpointId: EndpointNumber]: ClusterFilter;
+};
+
+/**
+ * FilteredAttributeResponse is a specialized version of AttributeResponse that processes a read/subscribe request
+ * with a filter applied to the attributes. Only the attributes that match the filter will be processed.
+ */
+export class FilteredAttributeResponse<
```
lauckhart, line 24: Maybe call it "AttributeSubscriptionResponse" since AttributeResponse is also filtered?

lauckhart (approved):
LGTM except for Q issue and the other minor comments... But approving since tests won't pass until Q thing is fixed anyway

packages/protocol/src/action/protocols.ts
Apollon77: likely eventToName is not needed ... i an remove that ...

packages/protocol/src/action/protocols.ts
Apollon77: Yes used because the state only has the property names that were changed, so we convert that in protocolService to be ids rather than numbers. name is already in clustertype protocol - but not fast to access

packages/protocol/src/action/protocols.ts
Apollon77: eventToName removed

packages/protocol/src/action/protocols.ts
Apollon77: attributeNameToId stays because used by ProtocolService - and likely too for the change event adjustment events

packages/protocol/src/action/server/FilteredAttributeResponse.ts
Apollon77: good idea

packages/protocol/src/action/protocols.ts
lauckhart: ProtocolService's requirements shouldn't affect the ClusterProtocol... That should only contain fields required by @matter/orotocol...  Which should only be dealing with IDs except for diagnostics, right?

Apollon77: I found a solution that should work

Apollon77: and also can remove that now :-) I just filter on a different place

packages/node/src/node/server/ProtocolService.ts
```diff
@@ -299,6 +297,30 @@ class ClusterState implements ClusterProtocol {
         this.type = type;
         this.#datasource = backing.datasource;
         this.#endpointId = backing.endpoint.number;
+
+        const attributeNameToIdMap = new Map<string, AttributeId>();
+        // For quieter attributes, we need to use the online events to get real state changes
+        for (const attr of type.attributes) {
+            attributeNameToIdMap.set(attr.name, attr.id);
+            if (attr.quieter) {
+                (this.#datasource.events[`${attr.name}$Changed`] as unknown as QuietEvent).online.on(() =>
```
lauckhart, line 306: For these events (and the `stateChanged` one below) we should probably register them using an `ObserverGroup`...  So we can do `this.#observers.close()` to unregister them in `deleteCluster`

Apollon77: Yes had on Todo. Done

packages/node/src/behavior/supervision/RootSupervisor.ts
```diff
@@ -48,6 +50,7 @@ export class RootSupervisor implements ValueSupervisor {
     #root: ValueSupervisor;
     #memberNames?: Set<string>;
     #persistentNames?: Set<string>;
+    #attributeIdsToNames?: Map<string, AttributeId>; // Whenever we need more than Attributes and Fields, we need to generalize
```
lauckhart, line 53: oops I said the name backwards, this is attributeNamesToIds :)

lauckhart (approved):

PR #2018, opened 2025-05-16
Significant model optimization

lauckhart:
Some details in changelog.  Primary benefit is removing human-facing stuff from the operational model.


Apollon77 (approved):
Awesome

PR #2043, opened 2025-05-21
Migrating "logEndpoint" to log via Disgnostics

Apollon77:
So now it is the same :-)

Looks like that for device

Bildschirmfoto 2025-05-21 um 10 40 28

And that for controller:

Bildschirmfoto 2025-05-21 um 10 41 10

Additionally the PR:
* Introduces a "getLocal()" on AttributeCLients to already prepare enw CLientNode where we also have a dedicated "local" value way
* Intoduce egtting endpoint protocol for one endpoint


lauckhart (approved):
I feel like we should have more PRs with emojis

PR #2092, opened 2025-06-01
Enhance CommissioningError by Status Enum

Apollon77:
... and check if maximum fabrics reached directly on start

### DIFF
--- a/packages/protocol/src/peer/ControllerCommissioningFlow.ts
+++ b/packages/protocol/src/peer/ControllerCommissioningFlow.ts
@@ -140,51 +140,41 @@ type CollectedCommissioningData = {
     fabricIndex?: FabricIndex;
 };
 
-export enum CommissioningErrorCode {
-    /** An unspecified error occurred during commissioning. Please check the error message for details. */
-    UnspecifiedError = 1,
+/**
+ * Error that throws when Commissioning fails and a process cannot be continued, and no more specific error
+ * information is available.
+ */
+export class CommissioningError extends MatterError {}
 
-    /** The number of fabrics that can be commissioned is already reached. */
-    MaximumCommissionedFabricsReached = 2,
+/** The number of fabrics that can be commissioned is already reached. */
+export class MaximumCommissionedFabricsReachedError extends CommissioningError {}
 
-    /** The commissioning process could not be finished within the maximum allowed commissioning timeframe. */
-    CommissioningTimeout = 3,
+/** The commissioning process could not be finished within the maximum allowed commissioning timeframe. */
+export class CommissioningTimeoutError extends CommissioningError {}
 
-    /** The device is already commissioned into the fabric. */
-    DeviceAlreadyCommissionedToThisFabric = 4,
+/** Error that throws when the device is already commissioned to this fabric. */
+export class DeviceAlreadyCommissionedToThisFabricError extends CommissioningError {}
 
-    /** The device is already commissioned to another fabric with the same label. */
-    FabricLabelConflict = 5,
+/** Error that throws when the device is already commissioned to another fabric with the same label. */
+export class FabricLabelConflictError extends CommissioningError {}
 
-    /** The device has no connected network and no credentials for Wifi or Thread network were configured for controller. */
-    WifiOrThreadNetworkCredentialsNotConfigured = 6,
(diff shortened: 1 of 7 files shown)


packages/protocol/src/peer/ControllerCommissioningFlow.ts
```diff
@@ -1225,11 +1312,18 @@ export class ControllerCommissioningFlow {
...
-        );
+        let transitionResult: InteractionClient | undefined;
+        try {
+            transitionResult = await this.transitionToCase(
+                this.interactionClient.address,
+                // Assume concurrent connections are supported if not know (which should not be the case when we came here)
+                isConcurrentFlow,
+            );
+        } catch (error) {
+            const commError = repackErrorAs(error, CommissioningError);
```
lauckhart: Changing the error class potentially drops diagnostic information:

* original error class name/support for typeof
* extra fields on SuppressedError and AggregateError
* extra fields on non-standard Error extensions
* current stack trace (which is likely truncated from the original error, or possibly even different in the case of promises)

I think we also don't copy the new Error#cause property, although that's a quick fix

So maybe we should be throwing the new error with the original error as the cause? Except maybe in limited circumstances where we receive a specific error and know the stack doesn't provide additional useful details

packages/protocol/src/peer/ControllerCommissioningFlow.ts
```diff
@@ -140,10 +140,53 @@ type CollectedCommissioningData = {
     fabricIndex?: FabricIndex;
 };
 
-/** Error that throws when Commissioning fails and process cannot be continued. */
-export class CommissioningError extends MatterError {}
+export enum CommissioningErrorCode {
```
lauckhart: Is there a reason to use an enum here vs. creating a specialized version of CommissioningError for each different cause?

Would be nice to avoid creating secondary differentiation mechanism

One reason is that it would simplify error tests from e.g.

    `error instanceof CommissioningError && error.code === CommissioningErrorCode.CommissioningTimeout`

to

     `error instanceof CommissioningTimeoutError`

lauckhart (approved):

packages/protocol/src/peer/ControllerCommissioningFlow.ts
Apollon77: I adjusted it in this pace and also adjusted "repack" method also when not used here now because of the second parameter. Please have a look again

packages/protocol/src/peer/ControllerCommissioningFlow.ts
Apollon77: We have such mechanisms already StatusResponse Error and such which also transports the code.
I was more thinking about the users which maybe can use a "switch case" easier with a code to translate the code into something meaningful for the user. And it might be easier for controller using code to have one  check in the catch tha
n multiple.

But effectively it just "felt" better for that case compared to 11(+) new classes. I am open to adjust it ...
@digitaldan @spudwebb @sammachin @Pierre-Gilles Any opinion from you as main Controller devs?

lauckhart: I think it makes sense with StatusResponseError because we're using an enum defined by the spec. I still generally create specialized versions when I throw StatusResponseError too, often just having the derived class set the status code

You can still switch on the constructor with individual error classes 

In general I've kind of stopped using enums except when dealing with externally defined numeric identifiers... They're kind of unnatural in JS and add overhead of object lookup, and unnecessary in TS because you can use "union of strings" types

digitaldan: My use case will be to catch the error and then send it back over my websocket channel for the client to handle.  I will have to create some type of mapping (numeric code, error name, etc...)  on both sides to handle this.  So an enum works nicely since i can easily map that, alternatively if you go the route of specialized error classes, it would be nice if they all conformed to an interface that has a "getErrorCode" or something like that, although i'm pretty flexible and will work with whatever is deemed best from a architecture POV.  

But i'm very excited about surfacing proper error codes, will make troubleshooting  so much easier!

Apollon77: Alright I would then tend to agree to Greg and will rework to several error classes

lauckhart (approved):
LGTM

PR #2104, opened 2025-06-02
Add Groups Server (and partial Client) implementation

Apollon77:
This PR adds support for Matter Groups on Device/Server side and partial (mainly untested for now) support on client side.

The PR can best be reviewed on commit level. It basically does:
* Rework network classes to allow to dynamically set and remove memberships of the interface
* Adds microsecond time support
* Splits up session classes and introduce SecureUnicastSession and SecureGroupSession as childs of an abstract SecureSession
* Adjust ACL checks to no longer rely on Session but fully on AccessControll.Session context
* Add Group logic in Interaction server, message and Exchange level
* Introduce Groups also basically in several client places - tests outstanding and will be done in CI

### DIFF
--- a/packages/protocol/src/groups/FabricGroupsManager.ts
+++ b/packages/protocol/src/groups/FabricGroupsManager.ts
@@ -16,7 +16,6 @@ export const GROUP_SECURITY_INFO = Bytes.fromString("GroupKey v1.0");
  * Class that contains an operational view on the Group Keys for a fabric
  */
 export class FabricGroupsManager {
-    #storage?: StorageContext;
     #fabric: Fabric;
     #groups: Groups;
     #messagingState: MessagingState;
@@ -54,10 +53,7 @@ export class FabricGroupsManager {
     }
 
     set storage(storage: StorageContext) {
-        if (this.#storage !== undefined) {
-            throw new InternalError("Storage context can only be set once.");
-        }
-        this.#storage = storage;
+        this.#messagingState.storage = storage;
     }
 
     /** Operative lookup of the group key sets by their id. */


packages/react-native/src/net/UdpChannelReactNative.ts
```diff
@@ -17,7 +17,9 @@ import {
     repackErrorAs,
     UdpChannel,
     UdpChannelOptions,
+    UdpSocketType,
 } from "#general";
+import { NodeJsNetwork } from "@matter/nodejs";
```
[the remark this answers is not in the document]
Apollon77: @gemini-code-assista Thank you, fixed. Good finding

packages/general/src/time/Time.ts
```diff
@@ -36,6 +36,15 @@ export class Time {
     }
     static readonly nowMs = (): number => Time.get().nowMs();
 
+    nowUs() {
+        if (performance && typeof performance.now === "function" && typeof performance.timeOrigin === "number") {
```
lauckhart: I would implement `nowUs()` using `performance` without the conditional, then add code below that tests for `performance` support and replaces `Time.prototype.nowUs` if necessary...  Just so the feature test only needs to run once

support/chip-testing/test/core/IDM.test.ts
```diff
@@ -6,8 +6,7 @@
 
 describe("IDM", () => {
     chip("IDM/*").exclude(
-        // We fail conformance tests, mainly for groups though there are DoorLock and OccupancySensor issues too
-        // TODO - revisit once we have groups
+        // Spec issues for DoorLock, see https://github.com/CHIP-Specifications/connectedhomeip-spec/issues/11712
```
lauckhart, line 9: Grr, I can't see, maybe add more details?

packages/general/test/util/IpTest.ts
```diff
@@ -53,6 +53,59 @@ describe("IP", () => {
...
+    describe("iPv6BytesToString", () => {
+        it("converts an IPv6 address to a string", () => {
+            const result = ipv6BytesToString(
+                Uint8Array.of(0xfe, 0x80, 0, 0, 0, 0, 0, 0, 0xe7, 0x77, 0x4f, 0x5e, 0xc6, 0x1e, 0x73, 0x14),
+            );
+
+            expect(result).equal("fe80::e777:4f5e:c61e:7314");
+        });
+
+        it("konvertiert eine reine Null-Adresse (::)", () => {
```
lauckhart: I think I agree with gemini here 🙂

packages/node/src/behavior/context/server/OnlineContext.ts
```diff
@@ -24,31 +31,45 @@ import { ContextAgents } from "./ContextAgents.js";
...
     const session = exchange?.session;
 
+    let isGroupSubject = false;
     if (session) {
-        assertSecureSession(session);
-        fabric = session.fabric?.fabricIndex;
-
-        // TODO - group subject
-        subject = session.peerNodeId;
+        if (isSecureGroupSession(session)) {
```
lauckhart: Could we move this logic into the session so doesn't need to be conditional?  Just a method with different implementations based on session class?

packages/node/src/behavior/context/server/OnlineContext.ts
```diff
@@ -133,12 +154,18 @@ export function OnlineContext(options: OnlineContext.Options) {
...
+        }
         let agents: undefined | ContextAgents;
         const context: ActionContext = {
             ...options,
-            session: session as SecureSession | undefined,
+            session,
             exchange,
-            subject,
+            subjects,
+            isGroupSubject,
```
lauckhart: Given that these fields only apply to group subjects maybe they could move into a separate subobject.  Totally half-assed guess but maybe something like:

```ts
type Subject = Subject.Node | Subject.Group;

namespace Subject {
    interface Node {
        kind: "node",
        id: NodeId,
    }
    
    interface Group {
        kind: "group",
        id: GroupId,
        hasValidMapping: boolean,
        endpoints: EndpointId[],
    }
}
```

packages/general/src/net/Network.ts
```diff
@@ -12,6 +12,8 @@ export class NetworkError extends MatterError {}
 
 export class NoAddressAvailableError extends NetworkError {}
 
+export const DEFAULT_MATTER_PORT = 5540;
```
lauckhart: Also consider `MATTER_PORT` or `STANDARD_MATTER_PORT` as names...  Just because we use as the default because it's the standard

packages/protocol/src/session/SecureGroupSession.ts
```diff
@@ -0,0 +1,206 @@
...
+} from "#general";
+import type { SessionManager } from "#session/SessionManager.js";
+import { NodeId } from "#types";
+import { SecureSession } from "./SecureSession.js";
+import { Session } from "./Session.js";
+
+const logger = Logger.get("SecureGroupSession");
+
+/** Secure Group session instance */
+export class SecureGroupSession extends SecureSession {
```
lauckhart: Maybe we could call this "GroupSession" since you can't have an insecure group session?

packages/protocol/src/session/SecureSession.ts
```diff
@@ -37,7 +27,12 @@ export class NoAssociatedFabricError extends StatusResponseError {
     }
 }
 
-export class SecureSession extends Session {
+export abstract class SecureSession extends Session {
+    readonly isSecure = true;
+    abstract fabric: Fabric | undefined;
+}
+
+export class SecureUnicastSession extends SecureSession {
```
lauckhart: Could we call this `NodeSession`?  I know it's "unicast" at the protocol level but "SecureUnicast" feels a little cryptic for something so prominent and if I understand correctly it's really a session where we know we're talking to a specific authenticated node.

Separately, would be nice to have in a separate file, it's a little confusing to have it here while `SecureGroupSession` has its own file.

packages/node/src/behaviors/access-control/AccessControlServer.ts
```diff
@@ -218,7 +219,7 @@ export class AccessControlServer extends AccessControlBehavior.with("Extension")
         if (relevantFabricIndex === undefined || this.events.accessControlEntryChanged === undefined) {
             return;
         }
-        const adminPasscodeId = session === undefined || session?.isPase ? 0 : null;
+        const adminPasscodeId = session === undefined || (isSecureUnicastSession(session) && session.isPase) ? 0 : null;
```
lauckhart: We do this twice and it's a little cryptic, maybe a utility function that returns an `AdminAuthority` or somesuch with `passcodeId` and `nodeId`?  And then `if (admin === undefined) { ... }` would make it clear we failed to identify the admin node

packages/nodejs/test/crypto/CryptoTest.ts
```diff
@@ -5,7 +5,7 @@
  */
 
 import { NodeJsCrypto } from "#crypto/NodeJsCrypto.js";
-import { Bytes, Key, PrivateKey, PublicKey } from "#general";
+import { Bytes, DataReader, Endian, Key, PrivateKey, PublicKey } from "#general";
```
lauckhart: Not the ideal time to mention this but I think we should adopt `DataView` conventions...  Make big-endian the default and have a boolean `littleEndian` parameter instead of the enum.

packages/protocol/src/action/server/AccessControl.ts
```diff
@@ -134,15 +134,27 @@ export namespace AccessControl {
         readonly fabric?: FabricIndex;
 
         /**
-         * The authenticated {@link SubjectId} for online sessions.
+         * The authenticated {@link SubjectId}s for online sessions. This includes the relevant Node Id, Group ID and
+         * also potential relevant Case Authenticated Tags.
          */
-        readonly subject?: SubjectId;
+        readonly subjects?: SubjectId[];
```
lauckhart: Do node/group/tag get special positions in the array?  Or how do you differentiate?

See my comment in OnlineContext.ts...  Feels like maybe this should be an object with `kind: "group" | "node"` and include group mapping and endpoint number as properties only for "group" kind

support/chip-testing/src/handler/LegacyControllerCommandHandler.ts
```diff
@@ -81,6 +82,49 @@ export class LegacyControllerCommandHandler extends CommandHandler {
         try {
             await this.#controllerInstance.start();
             logger.info(`-----> Controller ${this.#identity} started`);
+
+            // Add Default Group configuration as also used in Chip:
+            // https://github.com/project-chip/connectedhomeip/blob/master/src/lib/support/TestGroupData.h
+            // Start times here are formally epoch since 2020-01-01. We use 1970 base, but the used values are that
+            // small that they are all irrelevant anyway
```
lauckhart, line 89: Ah, I didn't realize what you meant by "static keys", should this have a TODO indicating that we should test with real key gen when it's ready?  I think that's true regardless of controller implementation

packages/types/src/datatype/GroupId.ts
```diff
@@ -42,3 +68,11 @@ export const TlvGroupId = new TlvWrapper<GroupId, number>(
     groupId => groupId,
     value => GroupId(value, false), // No automatic validation on decode
 );
+
+export function assertOperationalGroupId(
+    groupId: GroupId,
+): asserts groupId is Exclude<GroupId, typeof GroupId.UNSPECIFIED_GROUP_ID> {
```
lauckhart: Does this return type do anything?  My understanding is you can `Exclude` from a union but not from a primitive type...  So maybe just delete?

Also, could `assertOperationGroupId` be `assertIsGroupId`?  Not sure of implications of "operational" but "unspecified" group ID sounds like it's not really a group ID

Finally, wondering if `UNSPECIFIED_GROUP_ID` should be `NO_GROUP_ID` to align with `Fabric.NO_FABRIC_ID`...  Unclear "unspecified group ID" means, is it an anonymous group or something, or the absence of group identifier?

packages/protocol/src/groups/index.ts
```diff
@@ -4,6 +4,5 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-/*** THIS FILE IS GENERATED, DO NOT EDIT ***/
-
-export * from "#devices/solar-power";
```
lauckhart, line 9: This will be overwritten on next regen

packages/general/src/util/Map.ts
```diff
@@ -0,0 +1,65 @@
...
+ * Set change events.
+ */
+export interface ObservableMap<K, V> {
+    get added(): Observable<[key: K, newValue: V]>;
+    get deleted(): Observable<[key: K, oldValue: V]>;
+    get changed(): Observable<[key: K, newValue: V, oldValue: V]>;
+}
+
+/** Map extension to get added, deleted or changed events */
+export class BasicMap<K, V> extends Map<K, V> implements ObservableMap<K, V> {
```
lauckhart, line 18: Note that `BasicSet` has secondary indices that allow you to efficiently retrieve objects by any property.  So you can e.g. do `fabrics.get("id", someFabricId)`.

Maybe we could create utility function on `BasicSet` like `basicSet.mapOf<K extends keyof T>(field: K): Map<T[K], T>`.  Then you could get a convenient `Map` interface for multiple properties that ensures uniqueness and makes access patterns more natural

packages/node/src/behavior/system/network/ServerNetworkRuntime.ts
```diff
@@ -393,4 +401,137 @@ export class ServerNetworkRuntime extends NetworkRuntime {
...
+        });
+
+        observers.on(fabric.groups.groupKeyIdMap.deleted, async groupId => {
+            const memberships = this.#activeGroupMemberships.get(fabricIndex);
+            if (memberships === undefined) {
+                return;
+            }
+            const address = memberships.get(groupId);
+            if (address !== undefined) {
+                logger.debug(
```
lauckhart: Adding/dropping membership logic is copied three times, feels like there should be addGroupMembership/dropGroupMembership utility functions

And probably also addMembershipForFabricGroups/dropMembershipForFabricGroups to do all groups in the fabric

packages/node/src/behavior/system/network/ServerNetworkRuntime.ts
```diff
@@ -393,4 +401,137 @@ export class ServerNetworkRuntime extends NetworkRuntime {
             agent.get(SubscriptionBehavior).reestablishFormerSubscriptions(env.get(InteractionServer)),
         );
     }
+
+    #observersForFabric(fabricIndex: FabricIndex) {
```
lauckhart: This fabric/group functionality seems like it has pretty clear logical boundaries...  Could we create "FabricNetworking" and/or "GroupNetworking" utility class(es) in a separate file?

I ask mainly because `ServerNetworkRuntime` is already a bit of a mess, this would keep the line count down and make future refactoring more straightforward

packages/node/src/behaviors/group-key-management/GroupKeyManagementServer.ts
```diff
@@ -4,41 +4,426 @@
...
+const logger = Logger.get("GroupKeyManagementServer");
+
+const MAX_64BIT_TIME = BigInt("0xffffffffffffffff");
+
+const schema = GroupKeyManagementBehavior.schema!.extend({
+    children: [
+        DatatypeElement(
+            /** Enhanced GroupKeySetStruct with a fabric Index to act as fabric scoped struct */
+            { name: "GroupKeySetStructFS", type: "struct" },
+            FieldElement({ name: "GroupKeySetId", id: 0x0, type: "uint16", conformance: "M" }),
```
lauckhart: Is the ID used for anything?  It's optional and you can leave it off if not (if you can't it might be a bug)

packages/node/src/behaviors/group-key-management/GroupKeyManagementServer.ts
```diff
@@ -4,41 +4,426 @@
...
+        this.reactTo(this.events.groupKeyMap$Changed, this.#updateGroupKeyMap);
+
+        const lifecycle = this.endpoint.lifecycle as NodeLifecycle;
+        this.reactTo(lifecycle.online, this.#online);
+    }
+
+    async #online() {
+        // When a groups cluster exists on any other endpoint maxGroupKeysPerFabric needs to be minimum 3
+        let groupsFound = false;
+        this.endpoint.visit(endpoint => {
```
lauckhart: This probably only warrants a TODO, but what happens if `GroupsBehavior` appears on a different endpoint after initialization?  Assuming maxGroupKeysPerFabric is "F" this is maybe acceptable but then maybe worth a note

packages/node/src/behaviors/group-key-management/GroupKeyManagementServer.ts
```diff
@@ -4,41 +4,426 @@
...
-        this.state.groupTable = [];
     }
 
-    override keySetWrite(): MaybePromise {
-        throw new StatusResponseError(
-            "We do not support additional groups beyond the IPK",
-            StatusCode.ResourceExhausted,
+    async #handleFabricUpdate(fabric: Fabric) {
+        if (this.state.groupKeySets.length === 0) {
+            return; // No group key sets, so nothing to do
```
lauckhart, line 112: Do we need to handle the case where the fabric used to have group keys but does no longer?

packages/node/src/behaviors/group-key-management/GroupKeyManagementServer.ts
```diff
@@ -4,41 +4,426 @@
...
+        if (this.state.groupKeySets.length) {
+            const groupKeysForFabric = new Map<FabricIndex, GroupKeyManagement.GroupKeySet[]>();
+            for (const groupKeySet of this.state.groupKeySets) {
+                const fabricIndex = groupKeySet.fabricIndex;
+                const keys = groupKeysForFabric.get(fabricIndex) ?? [];
+                keys.push(groupKeySet);
+                groupKeysForFabric.set(fabricIndex, keys);
+            }
+
+            for (const fabric of fabrics) {
```
lauckhart: Maybe iterate over the entries in `groupKeysForFabric` instead?

packages/node/src/behaviors/groups/GroupsServer.ts
```diff
@@ -4,11 +4,185 @@
...
+} from "#types";
 import { GroupsBehavior } from "./GroupsBehavior.js";
 
+const logger = Logger.get("GroupsServer");
+
+// We enable group names by default
+const GroupsBase = GroupsBehavior.with(Groups.Feature.GroupNames);
+
+/**
+ * Monkey patching Tlv Structure of addGroup* commands to prevent data validation of the groupName field to be
```
lauckhart, line 33: Maybe we should have a method on `Behavior` that performs the invoke so you can override it?  So something like

```ts
    [Behavior.invoke](method: (...args: unknown[]) => unknown, args: unknown[]) {
        method.apply(this, args);
    }
```

...then you could either add a try/catch to change the error code or just switch on `method.name` and perform the validation before invoking the method.

packages/node/src/behaviors/groups/GroupsServer.ts
```diff
@@ -4,11 +4,185 @@
...
+ * handled as ConstraintError because we need to return a special error.
+ * We do this to leave the model in fact for other validations and only apply the change for our Schema-aware Tlv parsing.
+ */
+Groups.Cluster.commands = {
```
lauckhart, line 37: I think it would be better to clone the cluster and set static `cluster` property on `GroupsServer`.

So e.g.

```ts
    override static cluster: {
        ...Groups.Cluster,
        commands: {
            ...Groups.Cluster.commands,
            /* overrides here */
        }
    }
```

...just so we don't e.g. bork client validation or other uses of the cluster

packages/node/src/behaviors/groups/GroupsServer.ts
```diff
@@ -4,11 +4,185 @@
...
 /**
  * This is the default server implementation of {@link GroupsBehavior}.
  */
-export class GroupsServer extends GroupsBehavior {}
+export class GroupsServer extends GroupsBase {
+    override initialize() {
+        this.state.nameSupport.groupNames = this.features.groupNames;
+    }
+
+    #findRootEndpoint() {
```
lauckhart: Does this need to work on anything other than the node's root endpoint?  If not you can use `this.env.get(ServerNode)` instead of doing the tree search

Also, looking at the implementation of `ownerOfType`, it seems to use the `deviceClass` field rather than the `deviceType` field...  I think that's probably a bug, might make sense to change in this PR.

packages/protocol/src/groups/FabricGroupsManager.ts
```diff
@@ -0,0 +1,374 @@
...
+import { MessageReceptionStateEncryptedWithRollover } from "#protocol/MessageReceptionState.js";
+import { assertOperationalGroupId, EndpointNumber, GroupId, NodeId } from "#types";
+
+export const GROUP_SECURITY_INFO = Bytes.fromString("GroupKey v1.0");
+export const GROUP_KEY_INFO = Bytes.fromString("GroupKeyHash");
+
+/**
+ * Class that contains an operational view on the Group Keys for a fabric
+ */
+export class FabricGroupsManager {
```
lauckhart, line 18: Right now this is a mix of different concerns, it feels like it might be better decomposed into three (or more) classes.  E.g.

* `FabricGroupsManager` - class with `keySets: BasicSet<KeySet>` and `groups: BasicSet<Group>` properties
* `Group` - class implementing all logic associated with a specific group
* `KeySet` - class implementing all logic associated with a specific key set

The `groups` field for example could replace `#groupKeyIdMap`, `groupEndpoints` and all of the methods that take a `GroupId` as the key.

Note my comment on `BasicMap` r.e. using `BasicSet` to produce `Map` objects for specific access patterns like e.g. ID lookup

packages/protocol/src/groups/OperationalGroupKeySet.ts
```diff
@@ -0,0 +1,45 @@
...
+ * Copyright 2022-2025 Matter.js Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import type { GroupKeyManagement } from "#clusters/group-key-management";
+
+export type GroupKeySet = GroupKeyManagement.GroupKeySet;
+
+/** Enhanced structure of GroupKeySet to include operational data for easier operational processing. */
+export type OperationalGroupKeySet = GroupKeySet & {
```
lauckhart: R.e. my comment in `FabricGroupsManager`, this could perhaps be upgraded to a class that supports `KeySet` functionality

packages/protocol/src/session/SecureGroupSession.ts
```diff
@@ -0,0 +1,206 @@
...
+        this.#id = id;
+        this.#fabric = fabric;
+        this.#peerNodeId = peerNodeId;
+        this.keySetId = keySetId;
+        this.#operationalGroupKey = operationalGroupKey;
+
+        manager?.registerGroupSession(this);
+        fabric.addSession(this);
+
+        logger.debug(`Created secure GROUP session for fabric index ${fabric.fabricIndex}`, this.name);
```
lauckhart: You can also use Diagnostic.strong("group") to emphasize "group"

lauckhart (commented):
OK I went through everything in current state.  Looks good.  Lots of comments but that's because it's lots of code. :wink:

packages/general/test/util/IpTest.ts
Apollon77: Missed that one to translate.

packages/protocol/src/action/server/AccessControl.ts
Apollon77: Subjects is either 0 (passcode/pase) or group id or node id and cats. So multiple is only relevant for case sessions either cat. Acl check wise irrelevant. I can also split out cats again in separate property. Then it is one "main" subject and the card for case.

support/chip-testing/src/handler/LegacyControllerCommandHandler.ts
Apollon77: I think test wise that's not a topic. We need to decide for real usage how we manage the keys. Let's discuss tomorrow.

packages/types/src/datatype/GroupId.ts
Apollon77: I can rename. And yes no group id is basically "no group user here". But we basically never have this case. I think about adjustments.

packages/protocol/src/groups/index.ts
Apollon77: No that's git whyever thinking that the file got moved. But it is not.

packages/node/src/behaviors/group-key-management/GroupKeyManagementServer.ts
Apollon77: What is optional? It is conformance M

packages/node/src/behaviors/group-key-management/GroupKeyManagementServer.ts
Apollon77: Good catch. I will read again spec. Maybe we simply always set to 4.

packages/node/src/behaviors/group-key-management/GroupKeyManagementServer.ts
Apollon77: No. Fabric update will build a new fabric. So it gets initialized fresh. The FabricGroupMabsger is empty so we need to always fill if needed

packages/node/src/behaviors/groups/GroupsServer.ts
Apollon77: I just did like in the one other place where we have the same issue. I try if I get it done alone.

packages/protocol/src/session/SecureGroupSession.ts
Apollon77: Was just comparable to CASE and PASE :-)

packages/node/src/behaviors/groups/GroupsServer.ts
lauckhart: Yeah I remember that one, it was kind of a kludge...  If this pattern works I think we should use it.

packages/protocol/src/groups/index.ts
lauckhart: Hah, will resolve

packages/node/src/behaviors/group-key-management/GroupKeyManagementServer.ts
lauckhart: Oh I see I thought you were defining these fields because this was a custom datatype.  I was saying the "id" property of `FieldElement` is unnecessary in that case.

If you're just adding the `FabricIndex` to a standard datatype you should be able to do it without replicating all of the fields.  This would (hopefully) work:

```ts
const groupKeySetStruct = GroupKeyManagementBehavior.schema!.get(DatatypeModel, "GroupKeySetStruct")!;
const groupKeySetStructFS = groupKeySetStruct.extend({ name: "GroupKeySetStructFS", children: [ /* just the added children */ ] });
```

packages/general/src/util/Map.ts
lauckhart: I created #2107 for this

packages/node/src/behaviors/groups/GroupsServer.ts
lauckhart: I meant to delete this comment, we might do this eventually but kind of complicated for something we wouldn't use much.  I think just cloning the cluster is fine for now.

packages/protocol/src/session/SecureGroupSession.ts
lauckhart: kk.  well make them all fancy if you want :slightly_smiling_face:

packages/general/src/time/Time.ts
Apollon77: Done

support/chip-testing/test/core/IDM.test.ts
Apollon77: Added

Summary: the command responses from command 0xa and 0x15 have "wider" conformance than the requests which leads to the effect that the responses are expected to be present but are not because we do not have the requests. After Spec fix of conformance, the chip model used in test will match again

packages/node/src/behavior/context/server/OnlineContext.ts
Apollon77: I added "subjectFor(message?)" method

packages/node/src/behavior/context/server/OnlineContext.ts
Apollon77: half-assed but good idea I think. introduced

packages/general/src/net/Network.ts
Apollon77: renmed

packages/protocol/src/session/SecureGroupSession.ts
Apollon77: renamed

packages/protocol/src/session/SecureSession.ts
Apollon77: renamed

packages/node/src/behaviors/access-control/AccessControlServer.ts
Apollon77: I de-crypted it a bit

packages/nodejs/test/crypto/CryptoTest.ts
Apollon77: Ok I did a middle way to preserve some typing benefits

packages/protocol/src/action/server/AccessControl.ts
Apollon77: solved with above adjustments

packages/node/src/behaviors/groups/GroupsServer.ts
Apollon77: adjusted

packages/matter.js/src/device/PairedNode.ts
```diff
@@ -302,7 +302,7 @@ export class PairedNode {
         interactionClient: InteractionClient,
         reconnectFunc: (discoveryType?: NodeDiscoveryType, noForcedConnection?: boolean) => Promise<void>,
         assignDisconnectedHandler: (handler: () => Promise<void>) => void,
-        sessions: BasicSet<SecureUnicastSession, SecureUnicastSession>,
+        sessions: BasicSet<NodeSession, NodeSession>,
```
lauckhart: You shouldn't need to specify the second parameter, it defaults to the first...  So just `BasicSet<NodeSession>`

lauckhart (commented):
Cool, looks good

packages/node/src/behavior/system/network/ServerNetworkRuntime.ts
Apollon77: I refactored

packages/node/src/behavior/system/network/ServerNetworkRuntime.ts
Apollon77: I splitted it out

packages/node/src/behaviors/group-key-management/GroupKeyManagementServer.ts
Apollon77: Adjusted, thanks . Please have look if correct that way

packages/node/src/behaviors/group-key-management/GroupKeyManagementServer.ts
Apollon77: I adjusted the approach a bit and also added TODO

packages/node/src/behaviors/group-key-management/GroupKeyManagementServer.ts
Apollon77: adjusted

packages/node/src/behaviors/groups/GroupsServer.ts
Apollon77: Thats nt working that easiely .. it tells that Type wise this override can not be used because does not comply for the WithFeatures<...> type. I leave as is for now

packages/protocol/src/groups/FabricGroupsManager.ts
Apollon77: Alright I splitted it up a bit ... lets discuss .. I run a bit out of ideas for some parts :-)

packages/protocol/src/groups/OperationalGroupKeySet.ts
Apollon77: also handled

packages/cli-tool/src/providers/endpoint.ts
```diff
@@ -4,7 +4,8 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import { ClientNodes, Endpoint, ServerNode } from "#node";
+import type { ClientNodes, ServerNode } from "#node";
+import { Endpoint } from "#node";
```
lauckhart, line 8: You prefer this style vs` import { type ClientNodes...`?

packages/protocol/src/groups/FabricGroupsManager.ts
lauckhart: I think I see what you mean... It looks like you might be able to further decompose you broke out a "GroupMessaging" component to handle the counters and reception state...

Or not, it's already a lot smaller than it was

lauckhart (approved):

packages/cli-tool/src/providers/endpoint.ts
Apollon77: Yes, mixing that in one line is weird and I run into limits too somewhen)

packages/matter.js/src/CommissioningController.ts
```diff
@@ -772,6 +772,13 @@ export class CommissioningController {
             }
         }
     }
+
+    get groups() {
+        if (this.#controllerInstance === undefined) {
+            throw new ImplementationError("Controller instance not yet started. Please call start() first.");
+        }
+        return this.#controllerInstance.getFabrics()[0].groups;
```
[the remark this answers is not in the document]
Apollon77: WIll always exactly return one entry

packages/protocol/src/interaction/InteractionMessenger.ts
```diff
@@ -288,6 +300,9 @@ export class InteractionServerMessenger extends InteractionMessenger {
                             Status.InvalidAction,
                         );
                 }
+                if (isGroupSession) {
+                    break; // We do not support multiple messages in group sessions
+                }
```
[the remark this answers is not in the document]
Apollon77: @gemini-code-assist In general you referenved the right spec place for normal messaging. Group messaging is special as defined by spec and because groups never return a response by definition an interaction using group can always only consist of one message.

PR #2130, opened 2025-06-09
Crypto implementation using only JS standards

lauckhart:
* Cleans up the crypto API; extracts an interface, gives methods more meaningful names and converts the API to a simple object

* Adds `StandardCrypto` which `@matter/general` installs as a default if `crypto.subtle` is available

* Where possible uses native implementation provided by `crypto.subtle`

* AES-CCM is not available natively so we implement in JS.  The core cipher and CCM mode are broken out as separate components

* The AES-CCM implementation is as optimized as I could make it, using precomputed tables, shared buffers, direct ArrayBuffer manipulation, etc.

* Adds tests that run all applicable NIST test vectors for AES cipher and CCM mode

* Adds support for big ints (encoded as bytes) to `DerCodec` for handling DER signatures

* Adds additional crypto tests

* Moves tests that were previously in `@matter/nodejs` because we didn't have native crypto into the correct package

* Modifies mocks so they are not installed by default.  Each test installs as needed

* Modifies the test harness so mock entropy is "0x80" bytes instead of "0x00".  This is less fragile and makes it a little more clear where fake entropy is employed in tests

* Likewise, changes the default mock date to a time in recent past.  This is less fragile when converted to Matter time formats

* Added method to mock time that allows for running logic at a specific time

* Improve reset of global environment between tests.  They can now muck freely with the crypto and time implementations and the test harness will ensure changes revert between files

* Adds a "nonentropic" function to the crypto interface that allows functions to run with one of 255 different sets of fake "entropy" for testing purposes

* Adds `TestFabric` class that performs repeatable fabric generation for testing purposes across packages.  Replaces places where we did that with this implementation

* Removes the network singleton.  Only a few modules were still using it; now they use environment or take the implementation as input

* Made control-C work in the test harness

* Adds `b$` string tag to create a pseudo-literal syntax for byte arrays

* Remove `toBigInt()`; converts to standard `BigInt()` instead

* Add config options and vars for enabling/disabling node.js crypto and networking.  This was mainly for crypto so I could run tests with standard implementation

* Made matter-run print an actual error if file is invalid instead of just crashing

* Rename some of the CASE payload fields to match spec using initiator/responder terminology because previous naming was confusing


packages/general/src/crypto/Key.ts
```diff
@@ -559,7 +558,7 @@ export function Key(properties: Partial<Key>) {
 }
 
 /**
- * Private key factory.
+ * Factory for AES keys.
```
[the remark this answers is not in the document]
lauckhart: fixed

packages/general/src/crypto/StandardCrypto.ts
```diff
@@ -0,0 +1,254 @@
...
+const SIGNATURE_ALGORITHM = <EcdsaParams>{
+    name: "ECDSA",
+    namedCurve: "P-256",
+    hash: { name: "SHA-256" },
+};
+
+/**
+ * A {@link Crypto} implementation based on standard "crypto.subtle" with missing portions implemented using JS.
+ *
+ * WARNING: This code is unaudited.  Use a trusted native alternative where available.
```
Apollon77, line 27: We should add that also to the relevant readmes like for react native and general readme (and changelog)

packages/general/src/crypto/StandardCrypto.ts
```diff
@@ -0,0 +1,254 @@
...
+        return new Uint8Array(secret);
+    }
+}
+
+// Only install if VM supports Web Crypto
+if ((globalThis.crypto as any)?.subtle?.[Symbol.toStringTag] === "SubtleCrypto") {
+    Boot.init(() => {
+        let standard: StandardCrypto | undefined;
+
+        Crypto.get = () => {
```
Apollon77: Should we at least add a debug log - or in other central place log which crypto implementation is used?

packages/general/src/util/Bytes.ts
```diff
@@ -11,6 +11,23 @@ export enum Endian {
     Big,
 }
 
+/**
+ * String tag for hex literals.
+ */
```
Apollon77, line 18: maybe an example would be cool :-)

packages/nodejs/src/environment/NodeJsEnvironment.ts
```diff
@@ -117,13 +122,33 @@ function loadVariables(env: Environment) {
     };
 }
 
+function configureCrypto(env: Environment) {
+    if (!config.installCrypto || !(env.vars.boolean("nodejs.crypto") ?? true)) {
```
Apollon77, line 128: please add the three new vars to the docs above in class header. thanks. Or are they not supposed to be used by devs? Maybe even then

packages/nodejs/test/test.config.ts
```diff
@@ -1,11 +1,8 @@
 /**
  * @license
- * Copyright 2022-2025 Matter.js Authors
+ * Copyright 2022-2025 Project CHIP Authors
```
Apollon77: should we not have anywhere matter.js authors? :-)

packages/protocol/src/fabric/FabricManager.ts
```diff
@@ -253,7 +258,9 @@ export class FabricManager {
             );
         }
         this.#fabrics.set(fabricIndex, fabric);
-        await this.persistFabrics();
+        if (this.#fabrics) {
```
Apollon77: should that not be "if storage"? fabrics is always true

packages/nodejs/src/environment/NodeJsEnvironment.ts
lauckhart: Oh, I thought I did.  I added nodejs.crypto and nodejs.network, what's the third one?

packages/general/src/util/Bytes.ts
lauckhart: k

packages/nodejs/test/test.config.ts
lauckhart: Hah yeah, I must've pulled this file from my old crypto branch or something

packages/protocol/src/fabric/FabricManager.ts
lauckhart: yes I think so

packages/nodejs/src/environment/NodeJsEnvironment.ts
lauckhart: nvm figured it out, added nodejs.storage too

packages/general/src/crypto/StandardCrypto.ts
lauckhart: Yes good idea

packages/general/src/crypto/StandardCrypto.ts
lauckhart: Updated change log

I'm not sure how to best message r.e. the crypto stuff.  If we're going to add an "unaudited crypto" message to the doc files then it probably shouldn't focus solely on this implementation since none of our code has been formally reviewed I think?

We could specifically call out the additional potential weakness that the AES-CCM implementation may be more susceptible to timing or other side-channel attacks.  But that's kind of theoretical and I suspect would not be the biggest issue a proper review would turn up

So might be a good idea to add security info to the repo but I think that's a broader project than this PR.  May need outside guidance there...  We could maybe ask CHIP if we can point folks to the their vulnerability reporting system like they do in SECURITY.md.  I also can't find anything that indicates whether the CHIP source has been formally reviewed or direction r.e. best practices in terms of the specific implementations CHIP provides.

Apollon77: We clarify that offline aside of this PR

PR #2146, opened 2025-06-13
Remove Crypto.get() singleton

lauckhart:
This allows for more fine-grain control of entropy in tests.  It also aligns dependency injection of crypto with storage, networking and higher-level components.  And getting rid of singletons is good on its own I think.

Manually passing crypto around is a little annoying but it's encapsulated pretty well in low-level packages.  @matter/protocol is most affected, and users shouldn't notice at all.  I exposed a "crypto" property on some of the more common contextual components such as Fabric so it doesn't need to be passed separately in all cases.

Details:

* Remove Crypto.get() and the static methods that mapped to Crypto.get()

* Components now take crypto implementation as an input, either directly or via Environment

* StandardCrypto registers in Environment.default; Node.js and react native implementations likewise install into correct environment

* Made NodeJsEnvironment attempt to retrieve Crypto and/or Network from the initial default environment if the Node.js version is disabled

* Converted CertificateManager to a class.  It can be created and thrown away; this just allows us to invoke each of the 13 (formerly static) methods without passing in crypto as an argument every time

* Converted Crypto back to an abstract class and moved utility functions such as randomUint8 to the instance so they are more accessible without passing in crypto implementation

* Converted random number generation to native DataView, replaced getRandomUint*() with randomUint* getters, and modernized other methods

* Replaced Uint8Array -> bigint conversion with a version that uses DataView rather than round tripping through strings

* Removed MockCrypto from @matter/testing and nonentropic from @matter/general; consolidated to "MockCrypto" exported from @matter/general

* There's not a clear point where crypto is "used" now, so I added a method to perform logging of active implementation.  ServerNode and CommissioningController both invoke on startup

* We now disable node.js crypto for Bun where we know it doesn't work but that StandardCrypto does

* ReactNativeCrypto installs by default again, but uses the StandardCrypto version of encrypt/decrypt since quick crypto doesn't support AES-CCM

* Those last two we will hopefully eventually replace with feature detection so we always install the highest-quality implementation of each method (that works)


packages/general/src/crypto/Spake2p.ts
```diff
@@ -29,39 +29,50 @@ export interface PbkdfParameters {
...
+        const random = crypto.randomBigInt(32, P256_CURVE.Fp.ORDER);
+        return new Spake2p(crypto, context, random, w0);
     }
 
     constructor(
-        private readonly context: Uint8Array,
-        private readonly random: bigint,
-        private readonly w0: bigint,
-    ) {}
+        readonly crypto: Crypto,
```
Apollon77: why the readonly here? is it also used directly? Because so you also expose it a public

packages/general/src/crypto/StandardCrypto.ts
```diff
@@ -250,3 +243,5 @@ async function importKey(...params: unknown[]) {
         throw new KeyInputError("Invalid key", { cause });
     }
 }
+
+Environment.default.set(Crypto, new StandardCrypto());
```
Apollon77, line 248: or only set when not already set?
When I remember right "set" is hard overwriting?

packages/protocol/src/certificate/CertificateManager.ts
```diff
@@ -578,52 +578,62 @@ function extensionsToAsn1(extensions: BaseCertificate["extensions"]) {
...
 }
 
-export namespace CertificateManager {
-    function assertCertificateDerSize(certBytes: Uint8Array) {
-        if (certBytes.length > MAX_DER_CERTIFICATE_SIZE) {
-            throw new ImplementationError(
-                `Certificate to generate is too big: ${certBytes.length} bytes instead of max ${MAX_DER_CERTIFICATE_SIZE} bytes`,
-            );
-        }
+function genericBuildAsn1Structure({
```
Apollon77, line 581: could also be a private class method :-) like the other two

packages/general/src/crypto/Crypto.ts
```diff
@@ -31,7 +28,7 @@ export const CRYPTO_AUTH_TAG_LENGTH = 16;
 export const CRYPTO_SYMMETRIC_KEY_LENGTH = 16;
 export type CryptoDsaEncoding = "ieee-p1363" | "der";
 
-const logger = Logger.get("Crypto");
+const logger = new Logger("Crypto");
```
Apollon77: was this change wanted?

packages/general/src/crypto/Spake2p.ts
lauckhart: no, was an error.  sometimes I confuse ts and c++

packages/general/src/crypto/Crypto.ts
lauckhart: No, had removed the recreated

packages/general/src/crypto/StandardCrypto.ts
lauckhart: StandardCrypto.js and Environment.js are loaded unconditionally in the same public export so I don't think it's possible for there to be another implementation yet.  So different from @matter/nodejs.  This creates the baseline environment exported by @matter/general.  I'll add a comment

packages/protocol/src/certificate/CertificateManager.ts
lauckhart: Felt like it matched the code style of this module that already had a number of utilities functions that were outside of the namespace.  I tend to do that if a function is private and does not depend on private properties.  But LMK though, I can move it back if you want in separate PR

lauckhart:
> passing crypto everywhere not too nice, but ok ... thats the cost to not ave the singletons :-)

I agree, am not dead set on approach and I expect will continue to morph.  Here are some other considerations we may not have discussed yet:

* This is all temporary until if/when tc39 async-context lands and we can at least polyfill

* One way or another I expect we'll be passing more context around for a number of use cases:

  * Tying log messages to a specific node like requested in #1452

  * More uniformly associating messages with a specific exchange/context like we partially do with `via`

  * Deep instrumentation a la OpenTelemetry
  
* Short of async-context I'm not sure if/how we'll handle above but potentially we could have a single object we pass around that covers prng + these other use cases

PR #2150, opened 2025-06-13
Add project reference to top-level README

JimBuzbee:
Add reference to my project that shows examples of using the matter shell websocket


Apollon77 (approved):
Thanks a lot

PR #2164, opened 2025-06-16
Adjust Error handling for Enums to use ConstrainError

Apollon77:
also includes fixes #2155

### DIFF
--- a/.vscode/matter.code-snippets
+++ b/.vscode/matter.code-snippets
@@ -5,7 +5,7 @@
 		"body": [
 			"/**",
 			" * @license",
-			" * Copyright 2022-2025 Project CHIP Authors",
+			" * Copyright 2022-2025 Matter.js Authors",
 			" * SPDX-License-Identifier: Apache-2.0",
 			" */",
 			"",
@@ -17,15 +17,15 @@
 		"scope": "javascript,typescript",
 		"prefix": "@core",
 		"body": [
-			"@see {@link MatterSpecification.v13.Core} § "
+			"@see {@link MatterSpecification.v141.Core} § "
 		]
 	},
 
     "cluster": {
         "scope": "javascript,typescript",
-        "prefix: "@cluster",
+        "prefix": "@cluster",
         "body": [
-            "@see {@link MatterSpecification.v13.Cluster} § "
+            "@see {@link MatterSpecification.v141.Cluster} § "
         ]
     }
 }
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -1,6 +1,6 @@
 /**
  * @license
- * Copyright 2022-2025 Project CHIP Authors
+ * Copyright 2022-2025 Matter.js Authors
  * SPDX-License-Identifier: Apache-2.0
  */
 
--- a/packages/general/src/crypto/CryptoError.ts
+++ b/packages/general/src/crypto/CryptoError.ts
@@ -1,6 +1,6 @@
 /**
  * @license
- * Copyright 2022-2025 Project CHIP Authors
+ * Copyright 2022-2025 Matter.js Authors
  * SPDX-License-Identifier: Apache-2.0
  */
 
--- a/packages/general/src/crypto/MockCrypto.ts
+++ b/packages/general/src/crypto/MockCrypto.ts
@@ -1,6 +1,6 @@
 /**
  * @license
- * Copyright 2022-2025 Project CHIP Authors
+ * Copyright 2022-2025 Matter.js Authors
  * SPDX-License-Identifier: Apache-2.0
  */
 
--- a/packages/general/src/crypto/aes/Aes.ts
+++ b/packages/general/src/crypto/aes/Aes.ts
@@ -1,6 +1,6 @@
 /**
  * @license
- * Copyright 2022-2025 Project CHIP Authors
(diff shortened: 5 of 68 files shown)


packages/node/test/behavior/state/validation/conformanceTest.ts
```diff
@@ -466,12 +467,12 @@ const AllTests = Tests({
 
                 "disallows disallowed": {
                     record: { test: 3 },
-                    error: disallowedEnum("X", "disallowed", 3),
+                    error: disallowedEnum("all", "disallowed", 3),
```
lauckhart: This is testing whether conformance "X" works, so presumably it should still be a conformance error?  Like, we don't want "X" to throw a different error for different field types?

packages/node/src/behavior/state/validation/conformance-compiler.ts
```diff
@@ -582,7 +583,7 @@ export function astToFunction(schema: ValueModel, supervisor: RootSupervisor): V
 
     function disallowEnumValue(schema: Schema): EnumMemberValidator {
         return location => {
-            throw new ConformanceError(
+            throw new ConstraintError(
```
lauckhart: Given that it is not actually the "constraint" causing the issue, I think we should not use ConstraintError here...  Rather have an EnumValueError or something that maps to to a matter constraint error...  That way it won't say `constraint "all"` (or whatever the constraint actually is) in the error message which is just confusing

lauckhart (approved):

packages/protocol/src/action/errors.ts
```diff
@@ -80,6 +80,24 @@ export class ConstraintError extends ValidateError {
     }
 }
 
+/**
+ * Thrown when an enum value is not valid based on conformance definitions
+ */
+export class EnumValueConformanceError extends ValidateError {
+    constructor(schema: Schema, path: SchemaErrorPath, message: string) {
+        super(path, `Conformance "${(schema as ValueModel).conformance}": ${message}`, StatusCode.ConstraintError);
```
lauckhart: This should be a ConformanceError for consistency, you can just set the status code to ConstraintError after calling super

packages/node/src/behavior/state/validation/conformance-compiler.ts
```diff
@@ -31,6 +32,9 @@ import {
 } from "./conformance-util.js";
 import { ValidationLocation } from "./location.js";
 
+/** A no-op function used to mark that a conformance expression is valid. */
```
lauckhart: I think this is misleading, should be "used for valid enum values that have no specific conformance requirements"

packages/node/src/behavior/state/validation/conformance-compiler.ts
```diff
@@ -582,7 +586,7 @@ export function astToFunction(schema: ValueModel, supervisor: RootSupervisor): V
 
     function disallowEnumValue(schema: Schema): EnumMemberValidator {
         return location => {
-            throw new ConformanceError(
+            throw new EnumValueConformanceError(
```
lauckhart, line 589: The use of disallow() in the memberValidator generated in the Code.Evaluate case of addEnumMemberValidation should also probably be changed to this error

That code path is activated if there is field-value-dependent conformance

Previously it did not throw as useful of an error as the static case. But now I think it's actually incorrect because it would still result in InvalidAction

packages/protocol/src/action/errors.ts
```diff
@@ -80,6 +80,24 @@ export class ConstraintError extends ValidateError {
     }
 }
 
+/**
+ * Thrown when an enum value is not valid based on conformance definitions
+ */
+export class EnumValueConformanceError extends ValidateError {
```
lauckhart: I'm wondering if enum value is actually the correct thing to be differentiating. Right now we have:

* Invalid attribute - InvalidAction

* Invalid value because of enum conformance - ConstraintError

* Invalid value because of datatype - ConstraintError

* Invalid value because of constraint definition - ConstraintError

* Invalid value because of conformance error if value is not an enum - InvalidAction

That last one seems a little strange. Is it possible we should just be using InvalidAction if an attribute does not exist due to conformance, and ConstraintError for all other cases?

In that case we should probably change ConformanceError default status code to be ConstraintError, then just make sure it's changed to InvalidAction in the attribute case... Which is possibly already handled at a higher level because the attribute shouldn't be present if disallowed by conformance

packages/protocol/src/action/errors.ts
Apollon77: adjusted

packages/node/src/behavior/state/validation/conformance-compiler.ts
Apollon77: But it should be the case "disallow" in the evaluate path is mapped in L650 as

```
                    const disallow = disallowEnumValue(member);
```

so the "disallow" used in Disallow case is exactly that.

Or?

packages/protocol/src/action/errors.ts
Apollon77: I adjusted this too

packages/node/src/behavior/state/validation/conformance-compiler.ts
lauckhart: 👍

PR #2173, opened 2025-06-19
Allow any aspectCache keys to be created

FlyingNebulae:
Hello,

I working for a company that will be using matter in it's new products and I aim at using part of the Matter.js sdk in the browser.

However there is currently an error that is thrown when the SDK is bundled with frameworks like Angular and VueJS and I suspect the error would be found with other frameworks as well.

The error I get is the following : 
image

After some research I found out that the minification process from the bundlers, like esBuild is responsible of this error, which changes somehow some classes names that are used here in the Aspect cache

For example here the `this.name` would evaluate to `_Quality` instead of `Quality` which results in an error because the cache was initialized with the key `Quality`
image

**So the goal of the PR is to always initialize the aspectCache with an empty object the first time it is accessed.**

I hope all of this makes sense to you, if you have any questions don't hesitate

I'm fairly new to the Pull Request game, I hope you will be indulgent 🙏

### DIFF
--- a/packages/model/src/aspects/Aspect.ts
+++ b/packages/model/src/aspects/Aspect.ts
@@ -11,12 +11,7 @@ const empty = new Map<new (definition?: any) => Aspect, Aspect>();
 
 // Aspects are immutable, there are not many permutations, and their definitions are largely normalized strings.  So we
 // cache them to keep the object count down
-const aspectCache: Record<string, Record<string, Aspect>> = {
-    Access: {},
-    Conformance: {},
-    Constraint: {},
-    Quality: {},
-};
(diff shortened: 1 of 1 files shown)


Apollon77:
Welcome and thanks a lot. What do you think about Gemini comments?

For me it looks good, but I will request another developer to check too (@lauckhart).

Thats a great alternative to the "--keep-names" workaround.

If you need and other support or whenever you come near certification please contact me/us in discord. We are happy to support.

ClementVH:
@Apollon77 The comments seem legit, and works too, I will add them to the commit

PR #2200, opened 2025-06-27
Client node round 4

lauckhart:
This is a large round of changes that fleshes out the Node API for client usage.

Finishes off the controller features for ServerNode and fleshes out ClientNode to the point where it is largely functional.  Tests cover commissioning, syncing of attributes and endpoint structure, invoking a commands and reloading from cached attributes.

Other areas of improvement include:

* Upgrades to network simulation

* A new equivalent to the old "IntegrationTest" that performs in-process test of a matter.js controller against a matter.js device

* Refactoring of the storage layer for `@matter/node` which has always been a bit convoluted

* Misc. test harness fixes and upgrades

* A new protocol-level client implementation in ClientInteractable and ClientSubscriptions

* Improved support for IDs or names for keys in persisted Datasource values

* Datasource features to support underlying mutated externally rather than via ValueSupervisor

* Adds `commands` and `commandsFor` to allow for invocation directly from `Endpoint` API without creating a behavior

* Standard implementations of individual error classes for each `Status` code, and logic to automatically throw correct subclass based on a status code

* Proper support for suppressed errors generated by `using`

* Elimination of cyclical imports in @matter/node

* Tweaks to behavior generation to reduce the number of redundant types injected into declaration files


packages/general/src/util/Promises.ts
```diff
@@ -109,7 +105,6 @@ export async function withTimeout<T>(
                 reject(asError(e));
                 return;
             }
```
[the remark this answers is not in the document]
lauckhart: No, that was the previous behavior and is incorrect

packages/node/src/behavior/context/server/OfflineContext.ts
```diff
@@ -37,41 +39,90 @@ export const OfflineContext = {
...
+            });
+
+            return context;
+        } catch (e) {
+            if (transaction) {
+                transaction.reject(e);
+            } else {
+                frame?.close();
+                throw e;
             }
```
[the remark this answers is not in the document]
lauckhart: Reject rethrows so this is incorrect

packages/general/src/net/mock/index.ts
```diff
@@ -0,0 +1,10 @@
+/**
+ * @license
+ * Copyright 2022-2025 Project CHIP Authors
```
Apollon77: Please adjust to "matter.js Authors"

packages/node/src/behavior/internal/ClientBehaviorBacking.ts
```diff
@@ -5,26 +5,67 @@
...
+        for (const cmd of schema.commands) {
+            if (cmd.isRequest && commandIds.has(cmd.id as CommandId)) {
+                commands.add(camelize(cmd.name));
+            }
+        }
+
+        return (this.#elements = {
+            attributes,
+            commands,
+            events: new Set(),
```
Apollon77, line 64: Are events added later? Else why not add all events?

packages/node/src/behavior/system/commissioning/CommissioningClient.ts
```diff
@@ -344,5 +359,29 @@ export namespace CommissioningClient {
...
+         * By default matter.js reads all attributes on the node.  This allows us to efficiently initialize the complete
+         * node structure.
+         *
+         * If you only require a subset of attributes you can replace this with a more discriminative read.  For
+         * example, if you are only interested in interacting with the root endpoint and the On/Off cluster on other
+         * endpoints, you could do:
+         *
+         * ```js
+         * {
+         *     initialRead: Read(
```
Apollon77: When this initialRead propery her refers to?

packages/node/src/behavior/system/network/NetworkClient.ts
```diff
@@ -22,34 +18,31 @@ export class NetworkClient extends NetworkBehavior {
...
             minIntervalFloorSeconds: DEFAULT_MIN_INTERVAL_FLOOR_SECONDS,
             maxIntervalCeilingSeconds: 0,
+            attributes: [{}],
             ...startupSubscription,
         });
 
-        // TODO - subscription management & population of state
-        this.#node.interaction.subscribe(message);
+        // First, read.  This allows us to retrieve attributes that do not support subscription
+        for await (const _chunk of this.#node.interaction.read(subscribe));
```
Apollon77, line 38: and what is done with this read data?

packages/node/src/behavior/system/network/ServerNetworkRuntime.ts
```diff
@@ -344,7 +297,12 @@ export class ServerNetworkRuntime extends NetworkRuntime {
         this.owner.env.set(PeerAddressStore, new NodePeerAddressStore(this.owner));
         await this.owner.env.load(PeerSet);
 
-        await this.openAdvertisementWindow();
+        // Restore previous subscriptions
+        //
+        // TODO - move to SubscriptionsBehavior, and rename to SubscriptionsServer?
```
Apollon77, line 302: SubscriptionServer might have name overlapping ... maybe really SubscriptionManager?

packages/node/src/behaviors/general-commissioning/GeneralCommissioningServer.ts
```diff
@@ -75,25 +75,32 @@ export class GeneralCommissioningServer extends GeneralCommissioningBehavior {
                     AdministratorCommissioning.CommissioningWindowStatus.WindowNotOpen &&
                 !session.isPase
             ) {
+                // TODO - should this set status to Status.BusyWithOtherAdmin?
```
Apollon77, line 78: yeeaa seems so

packages/node/src/node/client/ClientBehavior.ts
```diff
@@ -0,0 +1,242 @@
...
+        fingerprint.push("a", createElementFingerprint(analysis.extraAttrs));
+    }
+
+    if (analysis.extraCommands.size) {
+        fingerprint.push("c", createElementFingerprint(analysis.extraCommands));
+    }
+
+    return fingerprint.join(";");
+
+    function createElementFingerprint(ids: Iterable<number>) {
```
Apollon77, line 207: Can you please add some infos here what this is for and what you do there? :-)

packages/node/src/node/client/ClientNodes.ts
```diff
@@ -147,14 +142,21 @@ export class ClientNodes extends EndpointContainer<ClientNode> {
             return;
         }
 
-        this.#expirationInterval = Time.sleep("client node expiration", EXPIRATION_INTERVAL).then(async () => {
-            this.#expirationInterval = undefined;
-            try {
-                await this.#cullExpiredNodesAndAddresses();
-            } finally {
+        this.#expirationInterval = Time.sleep("client node expiration", EXPIRATION_INTERVAL).then(
```
Apollon77, line 145: Why sleep and not a normal timer?. Weird construct ... and not cancellable?

packages/node/src/node/client/ClientStructure.ts
```diff
@@ -0,0 +1,339 @@
...
+                    currentUpdates = {
+                        endpointId,
+                        clusterId,
+                        values: {
+                            [attributeId]: change.value,
+                        },
+                    };
+
+                    // Update version but only if this was a wildcard read
+                    if (scope.isWildcard(endpointId, clusterId)) {
```
Apollon77, line 152: Not updating the version when read was not "wildcard is 100% correct ... formally updating the value in cache itsef is also not really right ... because if someone reads multiple attributes values that are expected to be "in sync" and one was updated later then the others that might end in confusion.

So I would propose to also not update the cache when the read was not wildcard ... and yes also not triggering onChanged because such reads are always done manually, so the response should be sufficient - bit then needed to be returned.

packages/nodejs/test/test.config.ts
```diff
@@ -1,6 +1,6 @@
 /**
  * @license
- * Copyright 2022-2025 Matter.js Authors
+ * Copyright 2022-2025 Project CHIP Authors
```
Apollon77: that also need to be adjusted

packages/protocol/src/action/client/ClientSubscriptions.ts
```diff
@@ -0,0 +1,171 @@
...
+
+    #timeOut(registration: ClientSubscription) {
+        logger.info(
+            "Subscription",
+            Diagnostic.strong(registration.subscriptionId),
+            "timed out after",
+            Diagnostic.strong(`${timeoutFor(registration)}ms`),
+        );
+
+        this.#closeOne(registration, new TimeoutError());
```
Apollon77, line 160: This should trigger a reconnection because assumption is that node it offline or something happened ... so .... maybe just add TODO for now because whole
 reconnection management is missing too

packages/node/src/behavior/system/network/NetworkClient.ts
Apollon77: Ok I found that

packages/node/src/behavior/system/network/NetworkClient.ts
```diff
@@ -22,34 +18,31 @@ export class NetworkClient extends NetworkBehavior {
...
+
+        // Now subscribe for subsequent updates
+        await this.#node.interaction.subscribe(subscribe);
```
Apollon77, line 41: Here we need to add reaction when this subscription closes when all connection handling is added ... see other comment. maybe leave a TODO

Apollon77:
And please add some Changelog entries :-)

packages/node/src/behavior/system/network/ServerNetworkRuntime.ts
lauckhart: This references the existing SubscriptionBehavior, which I think is probably server specific so would have "Server" suffix by convention

packages/node/src/node/client/ClientStructure.ts
lauckhart: I'm not sure formality plays a role here... You're either looking at entirely stale data or partially stale data, either way you're working with bad state, right?

I have a hard time arguing whether entirely-stale-but-same-version or partially-stale-but-mismatched versions is better

But that seems somewhat academic as it's kind of borked either way...  The use cases I was thinking of (that are perhaps still slightly contrived) are:

* Reading unsubscribable attributes, sensor data, etc... Make them available via state so you can update uniformly via e.g. polling without creating a separate path to get them to the user

* A default subscription that is not a wildcard... So if a binding only requires access to a subset of functionality or something

* Subscriptions with different frequency for "critical" vs less important attrs, maybe?

Anyway I'm not sure likelihood of this ever playing a role in real world either way but Matter has this fancy read/subscription filtering and it would be nice if we could avoid "ok you can use that but then you can't use behavior.state"

packages/node/src/behavior/system/network/ServerNetworkRuntime.ts
Apollon77: Alright, then name fine

packages/node/src/node/client/ClientStructure.ts
Apollon77: I completely agree, it all not 100%. correct - and if you consider "Q" and so some values are missing "by definition" their updates it also runs "out of sync".

So yes lets do it that way for now .. I just need to think about where/how we document this behavior. Lets do it and see if someone runs into issues because of it

packages/general/src/time/Time.ts
```diff
@@ -145,15 +146,24 @@ export class StandardTimer implements Timer {
...
         readonly name: string,
-        readonly intervalMs: number,
+        intervalMs: number,
         private readonly callback: Timer.Callback,
         readonly isPeriodic: boolean,
     ) {
+        this.intervalMs = intervalMs;
+    }
+
+    set intervalMs(intervalMs: number) {
```
Apollon77, line 161: Please limit set to throw when the timer is running at the moment because then changing it is intransparent

lauckhart: added comment

packages/node/src/behavior/system/commissioning/CommissioningClient.ts
lauckhart: I guess I renamed to startupSubscription

packages/node/src/behavior/internal/ClientBehaviorBacking.ts
lauckhart: Events aren't implemented yet but are mandatory in the API.  Will add TODO

packages/node/src/node/client/ClientBehavior.ts
lauckhart: yes

packages/node/src/node/client/ClientNodes.ts
lauckhart: A sleep is just a timeout in the format of a promise, and is cancelable because it is a CancelablePromise...  I used because it's there and a promise is idiomatic way in modern JS of indicating something that occurs in the future...

packages/node/src/behavior/system/network/NetworkClient.ts
lauckhart: added TODO

packages/protocol/src/action/client/ClientSubscriptions.ts
lauckhart: I added TODO

PR #2217, opened 2025-06-30
New controller example + ClientNode features

lauckhart:
New controller example + ClientNode features
  
* Adds an example that commissions/decommissions/toggles a device that supports On/Off
  
* Implement input fields for client invocation via Interactable
  
* Implement decommissioning for ClientNode

Needs to merge after #2200


packages/examples/src/control-onoff/OnOffController.ts
```diff
@@ -0,0 +1,79 @@
...
+
+switch (command) {
+    case "commission":
+        // Validation
+        if (controller.nodes.get("device")) {
+            die("There is already a device commissioned");
+        }
+
+        // Process arguments
+        const passcode = Number.parseInt(args[0]);
```
Apollon77: Ok this way it is mainly "dev showcase" ... maybe using pairingCode instead of the two because this is mostly known also for dev devices ... also commission method should maybe more use this?  but yeaah can also be fixed later

packages/examples/src/control-onoff/OnOffController.ts
```diff
@@ -0,0 +1,79 @@
...
+
+    case "toggle":
+        {
+            // Validation
+            const node = controller.nodes.get("device");
+            if (node === undefined) {
+                die("Cannot toggle because there is no commissioned device");
+            }
+
+            // Invocation
```
Apollon77, line 53: hm ... which endpoint does that check? Which OnOffClient is used when there are multiple on several endpoints?

packages/node/src/node/ClientNode.ts
```diff
@@ -77,10 +77,41 @@ export class ClientNode extends Node<ClientNode.RootEndpoint> {
...
+        if (!this.lifecycle.isCommissioned) {
+            return;
+        }
+
+        this.statusUpdate("decommissioning");
+
+        await this.act("decommission", agent => agent.commissioning.decommission());
+    }
+
+    override async erase() {
```
Apollon77, line 105: why decommission does not also erase?

packages/node/src/behavior/system/commissioning/CommissioningClient.ts
```diff
@@ -137,6 +141,20 @@ export class CommissioningClient extends Behavior {
...
+    async decommission() {
+        const { peerAddress } = this.state;
+
+        if (peerAddress === undefined) {
+            throw new ImplementationError("Cannot decommission node that is not commissioned");
+        }
+
+        const opcreds = this.agent.get(OperationalCredentialsClient);
+
+        await opcreds.removeFabric({ fabricIndex: opcreds.state.currentFabricIndex });
```
Apollon77, line 171: please add TODO because needs more error handling. right now an offline node can not be decommisisoned

Apollon77:
All comments during review and infos, so can be ignored and discussed aside or put to todo list for the new API ... were basically just thoughts

packages/examples/src/control-onoff/OnOffController.ts
lauckhart: Was an oversight, added endpoint number argument

packages/node/src/behavior/system/commissioning/CommissioningClient.ts
lauckhart: I intended for "decommissioning" to be the action of removing the fabric from the node.  By "offline node decommissioned" you mean deleting information about a peer without deleting the fabric on the peer?

packages/examples/src/control-onoff/OnOffController.ts
lauckhart: I went ahead and added pairingCode support, not tested though

packages/node/src/node/ClientNode.ts
lauckhart: I'm not sure if there's a use case for it but decommissioning doesn't remove the node, it just removes the fabric from the node...  If it's confusing we can maybe just remove the method from `ClientNode` so if you wanted to decommission without deleting locally for some reason you'd have to talk to CommissioningBehavior directly

lauckhart: I decided to remove "decommission" since decommission + erase + delete is confusing...  So now there's just "delete" for the common decommissioning use case and erase for low-level clear.  You can also decommission without deleting using `CommissioningClient` directly

packages/examples/src/control-onoff/OnOffController.ts
```diff
@@ -0,0 +1,84 @@
...
+const [, , command, ...args] = process.argv.slice(2);
+
+switch (command) {
+    case "commission":
+        // Validation
+        if (controller.nodes.get("device")) {
+            die("There is already a device commissioned");
+        }
+        if (args.length !== 1) {
+            die(`Usage: ${process.argv[0]} commission <passcode> <discriminator>`);
```
Apollon77:
```suggestion
            die(`Usage: ${process.argv[0]} commission <pairingcode>`);
```

packages/node/src/behavior/system/commissioning/CommissioningClient.ts
```diff
@@ -49,38 +59,42 @@ export class CommissioningClient extends Behavior {
...
-    async commission(options: number | CommissioningClient.CommissioningOptions) {
-        if (typeof options !== "object") {
+    async commission(options: number | string | CommissioningClient.CommissioningOptions) {
+        // Commissioning can only happen once
+        const node = this.endpoint as ClientNode;
+        if (this.state.peerAddress !== undefined) {
+            throw new ImplementationError(`${node} is already commissioned`);
+        }
+
+        if (typeof options === "number") {
```
Apollon77, line 76: Hm ... in fact both are numbers ... you can determine by length? passcode max is 99999999. Pairingcode is minimum 11 digits

packages/node/src/behavior/system/commissioning/CommissioningClient.ts
Apollon77: exactly. It happens very often that users reset the device to factory defaults and so have dangling nodes in the controller or that the device to decommission is already offline because thrown away. So we need a "force" option to also decommission when the "removeFabric" failed. The old API does that with a parameter? So force would catch and log the error but still remove the node internally

packages/node/src/behavior/system/commissioning/CommissioningClient.ts
```diff
@@ -384,4 +424,49 @@ export namespace CommissioningClient {
...
+         * The device's long discriminator.
+         */
+        discriminator?: number;
+    }
+
+    export interface PairingCodeOptions extends BaseCommissioningOptions {
+        /**
+         * The device's pairing code.
+         */
+        pairingCode: string;
```
Apollon77, line 444:
```suggestion
        pairingCode: number;
```

packages/node/src/behavior/system/commissioning/CommissioningClient.ts
Apollon77: But I can do that later when polishing usecases and API more

packages/node/src/node/ClientNode.ts
Apollon77: Makes sense

Apollon77 (approved):

PR #2251, opened 2025-07-16
Pass fabric and ca from options to matter.js MatterController class

ArtemisMucaj:
In the matter.js package the CommissioningControllerOptions interface defines two fields _rootFabric_ and _rootCertificateAuthority_ however the parameters are not passed into the MatterController instance created in **#initializeController** method.

This MR passes them from the CommissioningController class's _options_ property.


Apollon77:
Thanks for the contribution, and yes seems I forgot to expose it there as I added it :-) Thanks.

Apollon77 (approved):

PR #2255, opened 2025-07-17
Misc CHIP updates

lauckhart:
Bump version, patch out another 3 minute wait, fix parsing of args with escaped quotes inline


packages/node/src/behaviors/rvc-run-mode/RvcRunModeServer.ts
```diff
@@ -36,13 +36,6 @@ export class RvcRunModeServer extends RvcRunModeBehavior {
         ) {
             throw new ImplementationError("Provided supportedModes need to include at least one Cleaning mode tag");
         }
-        if (
```
Apollon77, line 39: Why this was removed? basically makes no sense because specs mandate it

Apollon77 (approved):

packages/node/src/behaviors/rvc-run-mode/RvcRunModeServer.ts
lauckhart: discussed in discord, is redundant

Issue #2240, opened 2025-07-13
ModeSelect Device and Semantic Tags


digitaldan:
I am trying to use a ModeSelect device on my bridged node.  I am getting the following errors when initializing the bridged endpoint

"Error initializing oh-bridge.aggregator.TestRollershutter.modeSelect Cause: Error: Validating oh-bridge.aggregator.TestRollershutter.modeSelect.state.standardNamespace: Value "1" is not defined in enum (135)"

My understanding of the spec ( i could be wrong):

If using the "common" namespace for semantic tags,  standardNamespace should be set to the namespace id, semanticTags.mfgCode should be null, and semanticTags.value should be set to the tag value (where the tag lives in the namespace of the id)

if no semanticTags are used, then standardNamespace can be null

If vendor specific semantic tags are used, then namespace can be set to null if no common tags are used, but semanticTags.mfgCode should be set to the vendor ID for vendor tags (but you can mix vendor and common tags on the same device).

If common tags are used, they have to belong to the same standardNamespace (no mixing common namespaces)

In the example below , 1 is the common namespace for "closure" , with UP/DOWN/STOP having tags values of 0/1/2 

Here's an example i'm using for testing.

```json
       {
    "modeSelect": {
        "modeSelect": {
            "mode": 0
        },
        "description": "Test Rollershutter",
        "supportedModes": [
            {
                "label": "Up",
                "mode": 0,
                "semanticTags": [
                    {
                        "mfgCode": null,
                        "value": 0
                    }
                ]
            },
            {
                "label": "Down",
                "mode": 1,
                "semanticTags": [
                    {
                        "mfgCode": null,
                        "value": 1
                    }
                ]
            },
            {
                "label": "Stop",
                "mode": 2,
                "semanticTags": [
                    {
                        "mfgCode": null,
                        "value": 2
                    }
                ]
            }
        ],
        "standardNamespace": 1,
        "currentMode": 0
    }
}
```

The common Tags are in matter.js, but a little hard to get to .   I have this typescript which i use in my code generation logic to create a "SemanticTags.java" class (the `const namespaces` is then used in a handlebar template)

```typescript
import * as MatterNode from "@matter/node";

const namespaces = Object.entries(MatterNode)
    .filter(([name]) => name.endsWith("Tag"))
    .map(([rawName, tagDefs]: [string, any]) => {
        // Get the common identifier (e.g. "Area" from "AreaNamespaceTag")
        let namespace = rawName;
[...]

PR #2262, opened 2025-07-20
Add enum values for Namespace override and use it

Apollon77:
Fills the enum values for the namespace datatype and use it in mode select.

Followup for #2257
Enhances #2240

@digitaldan FYI


support/models/src/local/namespace.ts
```diff
@@ -5,9 +5,99 @@
...
+            name: "Position",
+            details: "Tags which are useful in describing things related to position",
+        },
+        {
+            tag: "field",
+            id: 0xa,
+            name: "ElectricalMeasurement",
+            details: "Tags which are useful in describing electrical loads",
+        },
+        { tag: "field", id: 0xe, name: "Laundry", details: "Tags which are useful with laundry device types" },
```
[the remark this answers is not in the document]
Apollon77: Thats code style vs line length

support/models/src/local/namespace.ts
```diff
@@ -5,9 +5,99 @@
...
 import { LocalMatter } from "../local.js";
 
-// Namespace is a standard or device specific Enum ID, so lets handle as normal uint8 number for now
+// Namespace is a standard or device-specific Enum ID
 LocalMatter.children.push({
     tag: "datatype",
     name: "namespace",
-    type: "uint8",
+    type: "enum8",
+    children: [
```
lauckhart: Can we instead generate this using Matter.semanticNamespaces.map()?  Then we won't have to worry about copying in new namespaces in the future

support/models/src/local/ModeSelectOverrides.ts
```diff
@@ -21,13 +21,12 @@ LocalMatter.children.push({
...
 
-        // Spec defines as an enum but does not define values.  We could special case to generate an enum from the
-        // standard namespaces but for now we just set as an integer
+        // Spec defines as an enum8. We have the datatype "namespace" for exactly this enum, so let's use it.
         {
             tag: "attribute",
             name: "StandardNamespace",
             id: 0x1,
-            type: "uint16",
+            type: "namespace",
```
lauckhart, line 29: We also need to update Semtag#namespaceId and SemanticNamespace#id so types match up

lauckhart (changes requested):

support/models/src/local/ModeSelectOverrides.ts
lauckhart: The codegen tweak I mentioned this will require is to SemanticNamespaceFile.ts...  Needs to generate enum rather than raw number

packages/model/src/standard/resources/namespace.resource.ts
```diff
@@ -10,5 +10,131 @@ import { Resource } from "#models/Resource.js";
 
 Resource.add({
     tag: "datatype", name: "namespace", description: "Namespace", xref: "core§7.19.2.43",
-    details: "The Namespace type identifies the namespace used for a semantic tag."
+    details: "The Namespace type identifies the namespace used for a semantic tag.",
+
+    children: [
```
lauckhart, line 15: I think this should come out now, right?

lauckhart (approved):

packages/model/src/standard/resources/namespace.resource.ts
Apollon77: Thats generated yes, but seems also sometimes not 100% right (see Position)

PR #2276, opened 2025-07-25
Fix StringSchema.bound method min,max length

ArtemisMucaj:


Apollon77:
Ahh ok I see.

lets phrase it that way ...

As long as there is no TCP as communication channel and we only talk about UDP then the max message size is anyway too small for more than 1024 ... So not the question is why you need longer data and for which usecase.

Because of the fact that matter.js does not support TCP right now this limit simply makes sense because anything else could not be send anyway. But yes formally there is no spec limiting the length here.

For channel the limits with 8192 make actually no real sense ... I will follow up on that internally and then we see the results.

So spontan I would think to keep it as it but add a TODO for "Adding TCP support" to adjust that.

WDYT?

Apollon77:
Ok as summary also from Discord.

Yes we have a "practical String max" of 1024 in our Tlv class which is practically making sense given the fact we mainly use UDP and are so very limited in size anyway. The formal Spec max is 2^64-1 octets.

With the way the length definition was implemented before this PR we enforced the 1024 over longer lengths "silently" - thats basically wrong.

The change here makes sure we respect the given length so thats good.

So for the "LGTM" for me.

@lauckhart open question: WDYT should we stay for practical reasons with the 1024 "default max" or should we use "2^64-1" :-)

ArtemisMucaj:
I added the checks on the min and max length in the constructor

Apollon77 (approved):
Makes sense to me ...

lauckhart:
> @lauckhart open question: WDYT should we stay for practical reasons with the 1024 "default max" or should we use "2^64-1" :-)

I guess I'd probably let the error be "packet too large" just because that's the actual constraint, and we wouldn't need to remember to change if/when we have a transport that supports larger payloads

Entirely academic but MAX_SAFE_INT is max JS string length per spec, VMs have lower limits (e.g. 1GB), and runtime constraints prolly make even that largely hypothetical

PR #2279, opened 2025-07-25
Remove SyncStorage and add Blob Storage support

Apollon77:
This change includes the following:
* Remove SyncStorage
* Combines MabeAsyncStorage and Storage
* Adds Blob storage functionalities
* Adds BlobStorageContext
* For Filesystem based Storage adjust "has" to use stat instead of reading all data

It is preparation for OTA and the cases where we need to store bigger file based content and not normal attribute data

This is creaking so it will be merged a bit later ... but review does not hurt


packages/general/src/storage/StorageContext.ts
```diff
@@ -107,15 +67,74 @@ export class StorageContext<S extends Storage = any> implements StorageContextFa
...
+        }
+        return this.storage.set(this.thisContexts, keyOrValues);
+    }
+
+    values() {
+        return this.storage.values(this.thisContexts);
+    }
+}
+
+export class BlobStorageContext extends BaseStorageContext {
```
lauckhart: It's weird having BlobStorageContext not be a StorageContext.

Also, I may have just missed it, but what's the purpose of having BlobStorageContext and StorageContext be separate?  Vs. just putting the blob methods onto StorageContext, since they all proxy through to the same underlying Storage anyway?

packages/general/src/storage/StorageBackendMemory.ts
```diff
@@ -57,12 +57,57 @@ export class StorageBackendMemory extends SyncStorage implements CloneableStorag
...
+        const value = this.get(contexts, key);
+        if (value === undefined) {
+            return new ReadableStream<Uint8Array>({
+                start(controller) {
+                    controller.close();
+                },
+            });
+        }
+        if (!ArrayBuffer.isView(value)) {
+            throw new StorageError("Value must be an ArrayBuffer or a typed array to read as stream.");
```
lauckhart: ArrayBuffer is not a view, is this just "value is not a blob"?

lauckhart (dismissed):

packages/general/src/storage/StorageContext.ts
Apollon77: You are right that the underlaying storage is the same, but in fact "Blob" is "raw binary data" and the Key/Value store is "stringified data" values. As example: If you store a Uint8Array as blob and try to read it as "stringified get" that will fail. Thats why my idea was to make that clear by using two distinct interfaces. "has" is "interchangeable usable" on both but the blobStorage just expose the blob methods and vise versa. The idea was to have a distinct interface as per usage.

To make that clear we could also rename "StorageContext" to something like "StringifiedStorageContext" or such... WDYT? But sure if we know all that and never any such "mixup" is a case then I could also combine it again into one.

packages/general/src/storage/StorageBackendMemory.ts
Apollon77: I just used the same check as we use in other places too when we want to detect if it is an Uint8array style object ... in fact thats the purpose of the check. Any better idea?

packages/general/src/storage/StorageContext.ts
Apollon77: PS: I think we mainly use the "normal StorageContext". The Blob one is currently only relevant for BDX and OTA use cases

packages/general/src/storage/Storage.ts
```diff
@@ -12,18 +13,59 @@ export class StorageError extends MatterError {}
...
+     */
+    has(contexts: string[], key: string): MaybePromise<boolean> {
+        const value = this.get(contexts, key);
+        if (MaybePromise.is(value)) {
+            return MaybePromise.then(value, v => v !== undefined);
+        }
+        return value !== undefined;
+    }
+
+    abstract readBlob(
```
lauckhart: Forgot this in original review... Rather than having three methods for dealing with blobs maybe we could have just one that returns an object you can use to interact with a blob?

And more specifically could we just use the standard `Blob` API?

packages/general/src/storage/StorageBackendMemory.ts
lauckhart: The error is confusing because the caller of this method may not know about the underlying implementation. They asked for a blob but got an error about something unrelated

The other thing that's weird is the only thing it can legally be is a Uint8Array...  So even if the error is desirable, it should either specify Uint8Array or the code should be adjusted to support other buffer types

packages/general/src/storage/StorageContext.ts
lauckhart: I guess I'd take one of two approaches:

1. Implement blobs as a type of value.  Add minimal extension for `Blob` type values and treat the lack of string<->blob coercion as a (possibly temporary) inconvenience that results in a runtime error

2. Implement blob storage as a completely separate API based on standard `FilesystemDirectoryHandle`.  At some point in the future, replace the node-specific `StorageContext` implementation with one that builds on this

I'm thinking it'd be nice to avoid making our current k/v API more confusing...  I already find it kind of clunky

Apollon77: Check newest variant please

packages/general/src/storage/Storage.ts
Apollon77: Blob API added

Apollon77:
@copilot review

packages/nodejs/src/storage/StorageBackendDisk.ts
```diff
@@ -128,25 +167,56 @@ export class StorageBackendDisk extends MaybeAsyncStorage {
...
         return promise;
     }
 
-    async #writeAndMoveFile(filepath: string, value: string): Promise<void> {
+    async #writeAndMoveFile(filepath: string, valueOrStream: string | ReadableStream<Uint8Array>): Promise<void> {
         const tmpName = `${filepath}.tmp`;
-        await writeFile(tmpName, value, { encoding: "utf8", flush: true }); // flush does fsync after write
+        const handle = await open(tmpName, "w");
+        const isStream = valueOrStream instanceof ReadableStream;
+        const writer = handle.createWriteStream({ encoding: isStream ? null : "utf8", flush: true });
```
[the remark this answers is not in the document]
Apollon77: @copilot According to Node.js documentation null means binary

packages/nodejs/test/storage/StorageBackendJsonFileTest.ts
```diff
@@ -134,6 +134,46 @@ describe("Storage in JSON File", () => {
...
+        const stream = new ReadableStream<Uint8Array>({
+            start(controller) {
+                controller.enqueue(data);
+                controller.close();
+            },
+        });
+
+        await storage.writeBlobFromStream(["context"], "blobkey", stream);
+
+        const blob = storage.openBlob(["context"], "blobkey");
```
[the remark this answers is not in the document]
Apollon77: No thats ok

packages/nodejs/test/storage/StorageBackendJsonFileTest.ts
```diff
@@ -134,6 +134,46 @@ describe("Storage in JSON File", () => {
...
+        const data = new Uint8Array([31, 32]);
+        storage.set(["context"], "blobkey", data);
+
+        const blob = storage.openBlob(["context"], "blobkey");
+        assert.equal(blob.size, 2);
+    });
+
+    it("readBlob returns empty stream for missing key", async () => {
+        const storage = await StorageBackendJsonFile.create(TEST_STORAGE_LOCATION);
+        const blob = storage.openBlob(["context"], "missingkey");
```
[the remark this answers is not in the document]
Apollon77: Yes because in this case it is sync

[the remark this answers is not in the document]
Apollon77: not needed

lauckhart (approved):
lgtm

PR #2315, opened 2025-08-06
React native fixes

Apollon77:
This PR syncs changes from Luxni in https://github.com/Luxni/matter.js/tree/refactor-react-native to the main repository.

As discussed in Discord I:
* did not added some changes that very likely came from the esm/cjs import issue. So this would be needed to be validated
* did not took over the hard coded wlan interface na,e ... we need to understand why the interface is not correct, so maybe adding logging or debugging


packages/react-native/src/crypto/ReactNativeCrypto.ts
```diff
@@ -9,120 +9,444 @@ import { install } from "react-native-quick-crypto";
...
         );
-        if (!success) throw new CryptoVerifyError("Signature verification failed");
+
+        // Do not export as JWK because we do not want to inherit the algorithm and key_ops
+        const key = await this.#subtle.exportKey("jwk", subtleKey.privateKey);
+
+        // Extract only private and public fields; we do not want key_ops
+        return Key({
+            kty: KeyType.EC,
+            crv: CurveType.p256,
```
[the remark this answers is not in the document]
Apollon77: @Luxni Do you remember why you had this code

```
return Key({
            kty: KeyType.EC,
            crv: CurveType.p256,
            d: key.d?.slice(0, -1),
            x: key.x?.slice(0, -1),
            y: key.y?.slice(0, -1),
        }) as PrivateKey;
```

packages/react-native/src/crypto/ReactNativeCrypto.ts
```diff
@@ -9,120 +9,444 @@
...
-            data.forEach(chunk => signer.update(chunk));
-        } else {
-            signer.update(data);
+export class ReactNativeCrypto extends Crypto {
+    implementationName = "ReactNativeCrypto";
+    #subtle: ReactNativeCryptoSubtle;
+
+    constructor() {
+        const subtle = new ReactNativeCryptoSubtle();
+        if (subtle === undefined) {
```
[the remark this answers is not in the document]
Apollon77: Outdated

packages/react-native/src/crypto/ReactNativeCrypto.ts
```diff
@@ -9,120 +9,439 @@ import { install } from "react-native-quick-crypto";
...
-            start,
+import { ec } from "elliptic";
+
+const SIGNATURE_ALGORITHM = <EcdsaParams>{
+    name: "ECDSA",
+    namedCurve: "P-256",
+    hash: { name: "SHA-256" },
+};
+
+class ReactNativeCryptoSubtle implements SubtleCrypto {
```
lauckhart: I think this probably should be deleted...  AFAICT it just wraps crypto.subtle with a throw for import of non-jwk keys, which I address below

packages/general/src/crypto/index.ts
```diff
@@ -4,6 +4,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+export * from "./aes/Ccm.js";
```
lauckhart: This should be an internal import.  The functionality is available via StandardCrypto#encrypt and StandardCrypto#decrypt.

Those functions can be installed on another implementation without copying (like I originally had it on ReactNativeCrypto) although in this case ReactNativeCrypto should probably just extend StandardCrypto instead

packages/react-native/src/crypto/ReactNativeCrypto.ts
```diff
@@ -9,120 +9,439 @@ import { install } from "react-native-quick-crypto";
...
-        data: Uint8Array | Uint8Array[],
-        dsaEncoding: CryptoDsaEncoding = "ieee-p1363",
-    ): Uint8Array {
-        // @ts-expect-error No types but all fine
-        const signer = crypto.createSign(CRYPTO_HASH_ALGORITHM);
-        if (Array.isArray(data)) {
-            data.forEach(chunk => signer.update(chunk));
-        } else {
-            signer.update(data);
+export class ReactNativeCrypto extends Crypto {
```
lauckhart: This copies a bunch of stuff from StandardCrypto.  It should instead extend StandardCrypto with necessary overrides and have the copied code deleted

packages/react-native/src/crypto/ReactNativeCrypto.ts
```diff
@@ -9,120 +9,439 @@ import { install } from "react-native-quick-crypto";
...
+    ): Promise<CryptoKey>;
+    importKey(
+        format: KeyFormat,
+        keyData: BufferSource | JsonWebKey,
+        algorithm: AlgorithmIdentifier | RsaHashedImportParams | EcKeyImportParams | HmacImportParams | AesKeyAlgorithm,
+        extractable: boolean,
+        keyUsages: ReadonlyArray<KeyUsage>,
+    ): Promise<CryptoKey> {
+        if (format !== "jwk") {
+            throw new ImplementationError(`React Native Crypto does not support importKey with format "${format}"`);
```
lauckhart: Does the react native version of subtle crypto not throw an appropriate error?  If it does I'd suggest just letting that pass.  It's effectively an internal error because we shouldn't be attempting to import a raw key if we know rn doesn't support it

packages/react-native/src/crypto/ReactNativeCrypto.ts
```diff
@@ -9,120 +9,439 @@ import { install } from "react-native-quick-crypto";
...
-                key: jwk2pem(privateKey as jwk2pem.JWK, { private: true }),
-                format: "pem",
-                type: "pkcs8",
-                dsaEncoding,
-            }),
+        return new Uint8Array(await this.#subtle.digest("SHA-256", buffer));
+    }
+
+    async createPbkdf2Key(secret: Uint8Array, salt: Uint8Array, iteration: number, keyLength: number) {
+        const key = await this.#importKey("raw", secret, "PBKDF2", false, ["deriveBits"]);
```
lauckhart: I'm a little confused here because #importKey uses ReactNativeCryptoSubtle which will throw an error when importing a key in "raw" format.  Do I have that correct?  Maybe this isn't used in the current react native stuff?

If necessary it would be easy to implement raw key import using the `Key` class from `@matter/general`...  E.g. `Key({ privateBits: secret })` creates a JWK that can be imported as "jwk".

If the functionality is unnecessary for current use cases this could instead throw `NotImplementedError` for now

packages/react-native/src/crypto/ReactNativeCrypto.ts
```diff
@@ -9,120 +9,439 @@ import { install } from "react-native-quick-crypto";
...
+    }
+
+    randomBytes(length: number): Uint8Array {
+        const result = new Uint8Array(length);
+        crypto.getRandomValues(result);
+        return result;
+    }
+
+    encrypt(key: Uint8Array, data: Uint8Array, nonce: Uint8Array, associatedData?: Uint8Array) {
+        const ccm = Ccm(key);
```
lauckhart: If for some reason we can't extend StandardCrypto, please install methods like it did previously to avoid duplicate code and keep Ccm private to @matter/general

lauckhart (changes requested):
Looks fine to me except for the code copying in ReactNativeCrypto.ts, which should be straightforward to clean up

packages/react-native/src/crypto/ReactNativeCrypto.ts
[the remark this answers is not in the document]
Luxni: Because their data content contains a small dot at the end.

packages/react-native/src/crypto/ReactNativeCrypto.ts
Luxni: "ReactNativeCryptoSubtle" can actually be deleted.
This was created just to facilitate my debugging.

packages/react-native/src/crypto/ReactNativeCrypto.ts
Luxni: Yes, 'ReactNativeCrypto.ts' was copied from' StandardCrypto.ts'.
But since I don't know much about them, So I dare not delete any of the code.

packages/react-native/src/crypto/ReactNativeCrypto.ts
lauckhart: deleted

packages/react-native/src/crypto/ReactNativeCrypto.ts
lauckhart: deleted

packages/react-native/src/crypto/ReactNativeCrypto.ts
lauckhart: rewrote

packages/react-native/src/crypto/ReactNativeCrypto.ts
lauckhart: I rewrote to convert raw keys to jwk's prior to import.

packages/react-native/src/crypto/ReactNativeCrypto.ts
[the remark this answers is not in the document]
lauckhart: I added logic to conditionally strip these off if present after invoking the original implementation

packages/general/src/crypto/index.ts
lauckhart: removed

lauckhart (approved):
I think we may as well merge soon as it's closer to "right" than the previous version, then we can track subsequent updates separately

PR #2321, opened 2025-08-09
Device advertising refactor

lauckhart:
This refactors advertising logic.  This addresses a number of TODOs, adds features and generally attempts to make the logic more linear.

With this change we now have "advertisers" and "advertisements".  Two advertiser implementations, `MdnsAdvertiser` and `BleAdvertiser` manage a set of `Advertisement` instances, one for each active service ("commissionable node", "operational fabric+node", etc.).

`DeviceCommissioner` and `DeviceAdvertiser` previously were pretty intertwined.  Their roles are now more explicit:

- The commissioner handles commissioning timeout separately from advertising

- The commissioner now only deals with commissioning, not operational advertisement

- The advertiser is decomposed as described above

- The advertiser monitors fabrics and sessions to determine when to broadcast and/or publish (respond to queries)

- Advertisements act independently so e.g. a node may have simultaneously:

  - An associated `BleAdvertisement` and `CommissionableMdnsAdvertisement` broadcasting

  - An `OperationalMdnsAdvertisement` for fabric 1 published (responding to queries) but not actively broadcasting

  - An `OperationalMdnsAdvertisement` for fabric 2 published and broadcasting at a fast rate

  - An `OperationalMdnsAdvertisement` for fabric 3 published but broadcasting at a slower rate

Additional MDNS updates include:

- Extended advertisement with masking of vendor/product

- Support for TCP and ICD flags

- A configurable schedule for repeated MDNS broadcasts with fast defaults that align to the DNS-SD spec

- Logic to pause/resume advertising based on session establishment on a per-service basis

- More aggressive (but configurable) advertising when we retry due to missed ACKs

- `MdnsSocket` shared between MDNS clients and servers to cut down duplicate code and avoid processing packets twice

- Record generation is decomposed to make semantics clear and reduce code duplication between the different types of advertisements

- Reworked tests to reduce code duplication and remove yields so they are not tick-dependent

On the BLE side:

- BLE broadcast rates change over time; this is configurable with defaults that advertise as frequently as possible while conforming to Matter specs

- Support for extended advertisement with masking of vendor/product/aad

- Modified the `Ble` interface to be lower level with abstractions for the central and peripheral interfaces.  This allows higher-level logic (such as features above) to reside in @matter/protocol so we don't need to copy between platforms


lauckhart:
@Apollon77 particular points for you to look at that we haven't discussed:

- The logic in DeviceAdvertiser's constructor, specifically I changed how we handle failed acks in to be "advertise-per-retry", not sure if that was a good idea

- The BLE updates...  I don't think we have any test coverage for BLE and I don't have an imaged raspi for the moment...  Changes aren't too extensive but would be good to test

packages/general/src/util/Multiplex.ts
```diff
@@ -4,11 +4,62 @@
...
+    add(worker: Promise<unknown>, description?: string): void;
     close(): Promise<void>;
     [Symbol.asyncDispose](): Promise<void>;
 }
+
+/**
+ * A basic multiplex that tracks all promises given to it.
+ */
+export class BasicMultiplex implements PromiseLike<void> {
+    #workers = new BasicSet<Promise<unknown> & { description?: string }>();
```
Apollon77: maybe add a local type for it because repeated below

Apollon77: and really best idea to add the description entry to the promise object ...."Only" easier for close :-)

packages/node/src/behavior/system/network/ServerNetworkRuntime.ts
```diff
@@ -189,23 +194,23 @@ export class ServerNetworkRuntime extends NetworkRuntime {
...
 
     /**
      * When the first Fabric gets added we need to enable MDNS broadcasting.
      */
     enableMdnsBroadcasting() {
-        const advertiser = this.owner.env.get(DeviceAdvertiser);
-        const mdnsBroadcaster = this.mdnsBroadcaster;
-        if (!advertiser.hasBroadcaster(mdnsBroadcaster)) {
+        const device = this.owner.env.get(DeviceAdvertiser);
+        const mdnsBroadcaster = this.mdnsAdvertiser;
```
Apollon77: for consistency reasons rename to advertiser? :-)

packages/node/src/behavior/system/network/ServerNetworkRuntime.ts
```diff
@@ -285,15 +289,20 @@ export class ServerNetworkRuntime extends NetworkRuntime {
         // Monitor CommissioningServer to end "uncommissioned" mode when we are commissioned
         this.#observers.on(this.owner.eventsOf(CommissioningServer).commissioned, this.endUncommissionedMode);
 
-        // Monitor DeviceAdvertiser to enable MDNS broadcasting when the first Fabric is added
-        this.#observers.on(advertiser.operationalModeEnabled, this.enableMdnsBroadcasting);
```
Apollon77, line 289: not to myself .. check by what this is replaced

packages/nodejs-ble/src/index.ts
```diff
@@ -4,6 +4,6 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-export * from "./BleBroadcaster.js";
+export * from "../../protocol/src/advertisement/ble/BleAdvertiser.js";
```
Apollon77: this will not work :-)

packages/protocol/src/advertisement/Advertisement.ts
```diff
@@ -0,0 +1,348 @@
...
+/**
+ * An advertisement for a discrete service.
+ */
+export abstract class Advertisement<T extends ServiceDescription = ServiceDescription> {
+    /**
+     * The service advertised.
+     *
+     * This is a unique ID associated with the advertisement.
+     */
+    service: string;
```
Apollon77: really all public and writable? maybe some/all should be readone?

packages/protocol/src/mdns/MdnsSocket.ts
```diff
@@ -0,0 +1,184 @@
...
+        }
+    }
+
+    #handleMessage(bytes: Uint8Array, sourceIp: string, sourceIntf: string) {
+        // Ignore if closed
+        if (this.#isClosed) {
+            return;
+        }
+
+        // Skip unsupported source interface (not sure why this would happen, copied from old code)
```
Apollon77: we could remove and see where it breaks. Ok with that. I also do not remember anymore

Apollon77:
> The logic in DeviceAdvertiser's constructor, specifically I changed how we handle failed acks in to be "advertise-per-retry", not sure if that was a good idea

you mean that yu now just announce on first try? Thats fine ... I think makes sense

Apollon77:
@lauckhart Re test failures: you also might need to patch the Rvc Test app and bridge test app with the custom Mdns tweaks ... try that before you search deeper. at least core run12 of the Composition test is using RVCApp. Mode clusters and such too

packages/protocol/src/advertisement/Advertisement.ts
lauckhart: added readonly

packages/node/src/behavior/system/network/ServerNetworkRuntime.ts
lauckhart: yes

packages/general/src/util/Multiplex.ts
lauckhart: I made an interface for entries.  Didn't end up using this anyway.

packages/nodejs-ble/src/index.ts
lauckhart: hah right

packages/protocol/src/mdns/MdnsSocket.ts
lauckhart: OK will take out

packages/node/src/behavior/system/network/ServerNetworkRuntime.ts
lauckhart: I believe it's handled in the "already commissioned" case by the new lines that follow, and in the "newly commissioned" case it was already handled in `endUncommissionedMode()` so this was in that case.

lauckhart:
> @lauckhart Re test failures: you also might need to patch the Rvc Test app and bridge test app with the custom Mdns tweaks ... try that before you search deeper. at least core run12 of the Composition test is using RVCApp. Mode clusters and such too

Hmm yes good thought, all three failing tests are RVC.  Locally I only needed MDNS tweaks for python tests that use zeroconf so this may be a different failure mode, but will try it

Apollon77:
> Opposite actually, previously we only broadcast on first retry, now we broadcast (at least once) for every retry, although only for the specific fabric affected

Ahh ok ...hm ... in fact matter spec say that when resubmissions happen the node should advertise itself .... not really detailed how often and such

Apollon77 (approved):

PR #2333, opened 2025-08-12
Add devcontainer configuration

lauckhart:
Running via container allows all dockerized CHIP tests to pass on mac.


.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,58 @@
...
+// networking for UDP" and "Enable host networking" must be enabled.
+
+// For format details, see https://aka.ms/devcontainer.json. For config options, see the
+// README at: https://github.com/devcontainers/templates/tree/main/src/typescript-node
+{
+	"name": "matter.js",
+	// Or use a Dockerfile or Docker Compose file. More info: https://containers.dev/guide/dockerfile
+	"image": "mcr.microsoft.com/devcontainers/typescript-node:1-22-bookworm",
+
+	"postCreateCommand": "npm install",
```
Apollon77: Is npm ci better to ensure same deps?

Apollon77 (approved):
Maybe also add a changelog line?

.devcontainer/devcontainer.json
lauckhart: good point, really I should never use npm install, just ci and relock

PR #2335, opened 2025-08-12
OnOffServer adaptions for OO/2.3 tests

rsulzenbacher:
Actually, OO/2.3 is disabled in automatic tests:
https://github.com/matter-js/matter.js/blob/0c544f328aaf5d3382406f2098f162ce529aa348/support/chip-testing/test/app-slow/OO.test.ts#L19


Apollon77:
Thank you, in fact these two changes are correct according to spec, but likely not all ... we need to re-enable the tests and see whats up

PR #2344, opened 2025-08-16
Add proper units for intervals

lauckhart:
Creates new `TimeUnit` type to describe units of time and `Interval` class for intervals.  Convert many places that deal with intervals to take `Interval` as input rather than `number`.  This includes removing the suffixes we commonly used to convey units, e.g. "Ms", "Seconds", etc.

This adds type safety to time intervals and decouples the caller's units from the callee's units.

Note that we don't muck with names or types of properties defined by Matter standards.  Unfortunately Matter doesn't provide any interval-specific types so we can't codegen intervals based on the standard data model.  This commit does include an "interval" model type to facilitate configuration but we do not use this as an override type at this time.

The `Interval` and `TimeUnit` classes contain a variety of utility functionality to simplify time-interval-related reasoning.

Includes a bit of additional modernization for `Time` API and adds support for custom deep-comparison functionality.  This allows us to consider two intervals as identical regardless of the defining units.

Also improves formatting for various log messages I encountered while getting tests working.

### DIFF
--- a/packages/general/src/codec/DnsCodec.ts
+++ b/packages/general/src/codec/DnsCodec.ts
@@ -4,11 +4,15 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+import { Interval } from "#time/Interval.js";
+import { Seconds } from "#time/TimeUnit.js";
 import { InternalError, NotImplementedError, UnexpectedDataError } from "../MatterError.js";
 import { Bytes, Endian } from "../util/Bytes.js";
 import { DataReader } from "../util/DataReader.js";
 import { DataWriter } from "../util/DataWriter.js";
 import { ipv4BytesToString, ipv4ToBytes, ipv6BytesToString, ipv6ToBytes, isIPv4, isIPv6 } from "../util/Ip.js";
+
+export const DEFAULT_MDNS_TTL = Seconds(120);
 
 /**
  * The maximum MDNS message size to usually fit into one UDP network MTU packet. Data are split into multiple messages
@@ -16,7 +20,12 @@ import { ipv4BytesToString, ipv4ToBytes, ipv6BytesToString, ipv6ToBytes, isIPv4,
  */
 export const MAX_MDNS_MESSAGE_SIZE = 1232; // 1280bytes (IPv6 packet size) - 8bytes (UDP header) - 40bytes (IPv6 IP header, IPv4 is only 20bytes)
 
-export const PtrRecord = (name: string, ptr: string, ttl = 120, flushCache = false): DnsRecord<string> => ({
+export const PtrRecord = (
+    name: string,
+    ptr: string,
+    ttl = DEFAULT_MDNS_TTL,
+    flushCache = false,
+): DnsRecord<string> => ({
     name,
     value: ptr,
     ttl,
@@ -24,7 +33,7 @@ export const PtrRecord = (name: string, ptr: string, ttl = 120, flushCache = fal
     recordClass: DnsRecordClass.IN,
     flushCache,
 });
-export const ARecord = (name: string, ip: string, ttl = 120, flushCache = false): DnsRecord<string> => ({
+export const ARecord = (name: string, ip: string, ttl = DEFAULT_MDNS_TTL, flushCache = false): DnsRecord<string> => ({
     name,
     value: ip,
     ttl,
@@ -32,7 +41,12 @@ export const ARecord = (name: string, ip: string, ttl = 120, flushCache = false)
     recordClass: DnsRecordClass.IN,
     flushCache,
 });
-export const AAAARecord = (name: string, ip: string, ttl = 120, flushCache = false): DnsRecord<string> => ({
+export const AAAARecord = (
+    name: string,
+    ip: string,
+    ttl = DEFAULT_MDNS_TTL,
+    flushCache = false,
+): DnsRecord<string> => ({
     name,
     value: ip,
     ttl,
@@ -40,7 +54,12 @@ export const AAAARecord = (name: string, ip: string, ttl = 120, flushCache = fal
     recordClass: DnsRecordClass.IN,
     flushCache,
 });
-export const TxtRecord = (name: string, entries: string[], ttl = 120, flushCache = false): DnsRecord<string[]> => ({
+export const TxtRecord = (
+    name: string,
+    entries: string[],
+    ttl = DEFAULT_MDNS_TTL,
+    flushCache = false,
+): DnsRecord<string[]> => ({
     name,
     value: entries,
     ttl,
@@ -51,7 +70,7 @@ export const TxtRecord = (name: string, entries: string[], ttl = 120, flushCache
 export const SrvRecord = (
     name: string,
     srv: SrvRecordValue,
-    ttl = 120,
+    ttl = DEFAULT_MDNS_TTL,
     flushCache = false,
 ): DnsRecord<SrvRecordValue> => ({
     name,
@@ -81,7 +100,7 @@ export type DnsRecord<T = unknown> = {
     recordType: DnsRecordType;
     recordClass: DnsRecordClass;
     flushCache?: boolean;
-    ttl: number;
+    ttl: Interval;
     value: T;
 };
 
@@ -168,7 +187,7 @@ export class DnsCodec {
         const classInt = reader.readUInt16();
         const flushCache = (classInt & 0x8000) !== 0;
         const recordClass = classInt & 0x7fff;
-        const ttl = reader.readUInt32();
+        const ttl = Seconds(reader.readUInt32());
         const valueLength = reader.readUInt16();
         const valueBytes = reader.readByteArray(valueLength);
         const value = this.decodeRecordValue(valueBytes, recordType, message);
@@ -298,7 +317,7 @@ export class DnsCodec {
         writer.writeByteArray(this.encodeQName(name));
         writer.writeUInt16(recordType);
         writer.writeUInt16(recordClass | (flushCache ? 0x8000 : 0));
-        writer.writeUInt32(ttl);
+        writer.writeUInt32(ttl.secs);
         const encodedValue = this.encodeRecordValue(value, recordType);
         writer.writeUInt16(encodedValue.byteLength);
         writer.writeByteArray(encodedValue);
--- a/packages/general/src/log/Diagnostic.ts
+++ b/packages/general/src/log/Diagnostic.ts
@@ -4,6 +4,8 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+import { Interval } from "#time/Interval.js";
+import { Millisecs } from "#time/TimeUnit.js";
 import { Bytes } from "#util/Bytes.js";
 import type { Lifecycle } from "../util/Lifecycle.js";
 import { LogLevel } from "./LogLevel.js";
@@ -303,40 +305,8 @@ export namespace Diagnostic {
 
     export interface Elapsed {
         readonly startedAt: number;
-        readonly time: number;
+        readonly time: Interval;
         toString(): string;
-    }
-
-    /**
-     * Convert an interval to text.
-     */
-    export function interval(ms: number) {
-        if (ms < 0) {
-            return `${(ms * 1000).toPrecision(3)}μs`;
-        } else if (ms < 1000) {
-            return `${ms.toPrecision(3)}ms`;
-        } else if (ms < 60000) {
-            return `${(ms / 1000).toPrecision(3)}s`;
-        }
-
-        let days;
-        if (ms > 86_400_000) {
-            days = `${Math.floor(ms / 86_400_000)}d `;
-            ms %= 86_400_000;
-        } else {
-            days = "";
-        }
-        const hours = Math.floor(ms / 3_600_000)
-            .toString()
-            .padStart(2, "0");
-        ms %= 3_600_000;
-        const minutes = Math.floor(ms / 60_000)
-            .toString()
-            .padStart(2, "0");
-        ms %= 60_000;
-        const seconds = Math.floor(ms).toString().padStart(2, "0");
-
-        return `${days}${hours}:${minutes}:${seconds}`;
     }
 
     /**
@@ -347,11 +317,11 @@ export namespace Diagnostic {
             startedAt: performance.now(),
 
             get time() {
-                return performance.now() - this.startedAt;
+                return Millisecs(performance.now() - this.startedAt);
             },
 
             toString() {
-                return interval(this.time);
+                return this.time.toString();
             },
         };
     }
--- a/packages/general/src/net/RetrySchedule.ts
+++ b/packages/general/src/net/RetrySchedule.ts
@@ -5,6 +5,8 @@
  */
 
 import { Crypto } from "#crypto/Crypto.js";
+import { Interval } from "#time/Interval.js";
+import { Instant, Millisecs, Seconds } from "#time/TimeUnit.js";
 
 /**
  * An iterable of retry values based on a scheduling configuration.
@@ -29,31 +31,31 @@ export class RetrySchedule {
             maximumInterval,
             maximumCount,
             jitterFactor = 0,
-            initialInterval = 1000,
+            initialInterval = Seconds.one,
             backoffFactor = 2,
         } = this.config;
 
         let count = 0;
         let baseInterval = initialInterval;
-        let timeSoFar = 0;
+        let timeSoFar = Instant;
 
         while ((timeout === undefined || timeSoFar < timeout) && (maximumCount === undefined || maximumCount > count)) {
             count++;
-            const maxJitter = jitterFactor * baseInterval;
-            const jitter = Math.floor((maxJitter * this.#crypto.randomUint32) / Math.pow(2, 32));
-            let interval = baseInterval + jitter;
+            const maxJitter = jitterFactor * baseInterval.ms;
+            const jitter = Millisecs((maxJitter * this.#crypto.randomUint32) / Math.pow(2, 32)).floor;
+            let interval = baseInterval.plus(jitter);
 
-            if (timeout !== undefined && timeSoFar + interval > timeout) {
-                interval = timeout - timeSoFar;
+            if (timeout !== undefined && timeSoFar.plus(interval) > timeout) {
+                interval = timeout.minus(timeSoFar);
             }
             if (maximumInterval !== undefined && interval > maximumInterval) {
                 interval = maximumInterval;
             }
 
             yield interval;
-            timeSoFar += interval;
+            timeSoFar = timeSoFar.plus(interval);
 
-            baseInterval *= backoffFactor;
+            baseInterval = baseInterval.times(backoffFactor);
         }
     }
 }
@@ -61,16 +63,14 @@ export class RetrySchedule {
 export namespace RetrySchedule {
     /**
      * Configuration parameters for retry schedule.
-     *
-     * All intervals are in milliseconds.
      */
     export interface Configuration {
         /**
-         * Overall timeout in seconds.
+         * Overall timeout.
          *
          * Leave undefined to allow indefinite transmission.
          */
-        readonly timeout?: number;
+        readonly timeout?: Interval;
 
         /**
          * Maximum number of occurrences (including first).
@@ -82,9 +82,9 @@ export namespace RetrySchedule {
         /**
          * Interval between first request and final interval.
          *
-         * Defaults to 1000 ms.
+         * Defaults to 1s.
          */
-        readonly initialInterval?: number;
+        readonly initialInterval?: Interval;
 
         /**
          * Multiplier for subsequent retries.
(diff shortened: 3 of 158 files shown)


packages/general/src/codec/DnsCodec.ts
```diff
@@ -81,7 +100,7 @@ export type DnsRecord<T = unknown> = {
     recordType: DnsRecordType;
     recordClass: DnsRecordClass;
     flushCache?: boolean;
-    ttl: number;
+    ttl: Interval;
```
Apollon77: Interval could easiely nixed with nodejs interval. Maybe TimeDuration or IntervalDuration might be a more generic Name?

packages/general/src/log/Diagnostic.ts
```diff
@@ -347,11 +317,11 @@ export namespace Diagnostic {
             startedAt: performance.now(),
 
             get time() {
-                return performance.now() - this.startedAt;
+                return Millisecs(performance.now() - this.startedAt);
```
Apollon77: Seconds but Milisecs? MiliSeconds? ;-)

packages/general/src/time/Interval.ts
```diff
@@ -0,0 +1,237 @@
...
+                return "0";
+
+            case Infinity:
+                return "forever";
+
+            case -Infinity:
+                return "until now"; // Umm...  I guess?
+        }
+
+        if (ms === 0) {
```
Apollon77: Equal to first case above

packages/general/src/time/Time.ts
```diff
@@ -137,7 +149,7 @@ export namespace Timer {
 export class StandardTimer implements Timer {
     #timerId: unknown;
     #utility = false;
-    #intervalMs = -1;
+    #interval = Instant;
```
Apollon77: Remove default? Overwritten in constructor anyway

packages/general/src/util/DeepEqual.ts
```diff
@@ -4,19 +4,41 @@
...
+    if (aIsObject) {
+        if (DeepComparable.equals in a) {
+            return (a as DeepComparable)[DeepComparable.equals](b);
+        }
+    }
+
+    const bIsObject = typeof b === "object" && b !== null;
+    if (bIsObject) {
+        if (DeepComparable.equals in b) {
+            return (a as DeepComparable)[DeepComparable.equals](b);
```
Apollon77: Here a and b needs to be switched. Or?

packages/matter.js/src/device/PairedNode.ts
```diff
@@ -232,18 +235,15 @@ export class PairedNode {
                 );
                 // Try last known address first to speed up reconnection
                 this.#setConnectionState(NodeStates.Reconnecting);
-                this.#scheduleReconnect(0);
+                this.#scheduleReconnect();
```
Apollon77: 0 is really wanted here

packages/matter.js/src/device/PairedNode.ts
```diff
@@ -515,7 +515,7 @@ export class PairedNode {
             );
             return;
         }
-        this.#scheduleReconnect(0);
+        this.#scheduleReconnect();
```
Apollon77: 0 is really wanted here. Undefined has special meaning

packages/node/src/behavior/system/subscription/SubscriptionBehavior.ts
```diff
@@ -115,10 +115,10 @@ export class SubscriptionBehavior extends Behavior {
                     ),
                 ),
                 FieldElement({ name: "isFabricFiltered", type: "bool" }),
-                FieldElement({ name: "maxIntervalCeilingSeconds", type: "uint16" }),
```
Apollon77, line 118: But that's now not backward compatible. Or? Persisted Darwin old format will miss data. We need to either clear them or migrate them

packages/node/src/behaviors/switch/SwitchServer.ts
```diff
@@ -170,7 +170,8 @@ export class SwitchBaseServer extends SwitchServerBase {
                 this.internal.currentLongPressPosition = newPosition;
                 this.internal.longPressTimer = Time.getTimer(
                     "longPress",
-                    this.state.longPressDelay,
+                    // TODO - old behavior is default delay of zero; should we throw an error instead?
```
Apollon77: Good question.  It maybe in initialize already base on featureset?

packages/node/src/behaviors/switch/SwitchServer.ts
```diff
@@ -226,7 +227,8 @@ export class SwitchBaseServer extends SwitchServerBase {
             if (!pressSequenceFinished) {
                 this.internal.multiPressTimer = Time.getTimer(
                     "multiPress",
-                    this.state.multiPressDelay,
+                    // TODO - old behavior is default delay of zero; should we throw an error instead?
```
Apollon77: Same here.

packages/protocol/src/advertisement/ble/BleAdvertisement.ts
```diff
@@ -58,10 +59,10 @@ export class BleAdvertisement extends Advertisement<ServiceDescription.Commissio
                 await peripheral.advertise(advertisementData, aad, broadcastInterval);
 
                 // Wait for timeout at this broadcast interval
-                await context.sleep("BLE advertisement interval", Math.min(timeout, sleepTime));
+                await context.sleep("BLE advertisement interval", Interval.min(timeout, sleepTime));
 
-                timeout -= sleepTime;
-                if ((timeout -= sleepTime) <= 0) {
+                timeout = timeout.minus(sleepTime);
```
Apollon77: that's also a bugfix. Cool

Apollon77 (changes requested):
Wow. Huge effort. In general makes sense. 
I personally have just an "issue" with these quarter and half convenience options because that's - compared to specs - not natural. There it is seconds or milliseconds or "1/10th". And I always need to calculate. (And I personally will forget them when I implement specs. So no clue if they bring that much in the end.

packages/general/src/codec/DnsCodec.ts
lauckhart: Duration may be a better word, that's what Java and Temporal use. I don't think "Time" or "Interval" as a prefix adds much.

I mainly used "interval' because lots of matter variables use it to describe a duration, plus JS with setInterval.  But nbd to change.

What's a Node interval?

packages/general/src/log/Diagnostic.ts
lauckhart: Definitely no extra camelcase, and it appears a lot so Milliseconds kind of sucks...  The interwebs claim that Millis is a synonym for ms so we could use that, it's even shorter

packages/node/src/behavior/system/subscription/SubscriptionBehavior.ts
lauckhart: Meh I didn't realize we persisted. I guess we need to add a format version to storage

packages/general/src/codec/DnsCodec.ts
Apollon77: Yeahh I meaned nodejs setTimeout vs setInterval ... so coming from there "interval" is "periodic recurring" ... just want to make sure folks do not confuse it with that because we also have getTimer vs getPeriodicTimer and so do not use "interval" there ...

packages/general/src/log/Diagnostic.ts
Apollon77: then do that ... has no overlap to "seconds" then ... yeaahh i like that

packages/node/src/behavior/system/subscription/SubscriptionBehavior.ts
Apollon77: Yeaahh these are the "persisted subscriptions". Basically "future proof wise" we might need a "persisted data versioning" anyway,  but don't want to add that here tbh

packages/general/src/time/Interval.ts
```diff
@@ -0,0 +1,181 @@
...
+ * {@link Millisecs}.  For example, `Millisecs(Hours(1) + Minutes(30))` would produce a 90 minute {@link Interval}.
+ */
+export type Interval = Branded<number, "Interval">;
+
+/**
+ * Create an interval from an {@link Interval.Definition}.
+ */
+export function Interval<T extends undefined | Interval | string>(
+    source: T,
+): T extends undefined ? undefined | Interval : Interval {
```
Apollon77: WHy undefined here? also will never be returned because the typeof check it would throw

packages/general/src/time/Time.ts
```diff
@@ -19,56 +21,62 @@ const registry = new Set<Timer>();
...
-    static get: () => Time;
+    static default: Time;
 
     static startup = {
         systemMs: 0,
         processMs: 0,
     };
 
-    now() {
+    get now() {
```
Apollon77, line 32: please also add this as breaking to changelog ...

packages/general/src/net/RetrySchedule.ts
```diff
@@ -29,48 +31,46 @@ export class RetrySchedule {
...
         let baseInterval = initialInterval;
-        let timeSoFar = 0;
+        let timeSoFar = Instant;
 
         while ((timeout === undefined || timeSoFar < timeout) && (maximumCount === undefined || maximumCount > count)) {
             count++;
             const maxJitter = jitterFactor * baseInterval;
-            const jitter = Math.floor((maxJitter * this.#crypto.randomUint32) / Math.pow(2, 32));
-            let interval = baseInterval + jitter;
+            const jitter = Millisecs.floor(Millisecs((maxJitter * this.#crypto.randomUint32) / Math.pow(2, 32)));
```
Apollon77: ypu could allow to enter a number for these tooling funcs ... and then return millisecs because in fact the dev state the millisecs usage by using this method already? Then you prevent this additional "conversion"

packages/node/src/behaviors/general-diagnostics/GeneralDiagnosticsServer.ts
```diff
@@ -415,7 +415,7 @@ export namespace GeneralDiagnosticsServer {
                         return 0;
                     }
 
-                    return Math.round((Time.nowMs() - onlineAt.getTime()) / 1000);
+                    return Math.round((Time.nowMs - onlineAt.getTime()) / 1000);
```
Apollon77: Seconds.round()?

packages/node/src/behaviors/general-diagnostics/GeneralDiagnosticsServer.ts
```diff
@@ -430,7 +430,7 @@ export namespace GeneralDiagnosticsServer {
                     const totalOperationalHoursCounter =
                         endpoint.stateOf(GeneralDiagnosticsServer).totalOperationalHoursCounter;
                     return Math.floor(
```
Apollon77: Hours.floor()?

packages/protocol/src/mdns/MdnsClient.ts
```diff
@@ -1528,32 +1528,54 @@ export class MdnsClient implements Scanner {
...
-                DN: data.DN,
-                RI: data.RI,
-                PI: data.PI,
-            },
-            true,
-        );
+    static discoveryDataDiagnostics(data: DiscoveryData, kind?: string) {
+        return Diagnostic.dict({
+            kind,
+            SII: data.SII && Interval.format(data.SII),
```
Apollon77: Allow format to get undefined and return that if passed, so you bypass such safe guards?

packages/protocol/src/peer/ControllerCommissioningFlow.ts
```diff
@@ -595,45 +600,45 @@ export class ControllerCommissioningFlow {
...
     }
 
-    get #failSafeTimeLeftS() {
+    get #failSafeTimeLeft() {
         if (this.#currentFailSafeEndTime === undefined) {
-            return 0;
+            return Instant;
         }
-        return Math.max(0, Math.ceil((this.#currentFailSafeEndTime - Time.nowMs()) / 1000));
+        return Math.max(0, Math.ceil((this.#currentFailSafeEndTime - Time.nowMs) / 1000));
```
Apollon77: Seconds?

packages/protocol/src/protocol/MessageChannel.ts
```diff
@@ -149,27 +152,29 @@ export class MessageChannel implements Channel<Message> {
...
      */
     getMrpResubmissionBackOffTime(retransmissionCount: number, sessionParameters?: SessionParameters) {
-        const { activeIntervalMs, idleIntervalMs } = sessionParameters ?? this.session.parameters;
+        const { activeInterval, idleInterval } = sessionParameters ?? this.session.parameters;
         const baseInterval =
-            sessionParameters !== undefined || this.session.isPeerActive() ? activeIntervalMs : idleIntervalMs;
-        return Math.floor(
-            MRP.BACKOFF_MARGIN *
+            sessionParameters !== undefined || this.session.isPeerActive() ? activeInterval : idleInterval;
+        return Millisecs.round(
```
Apollon77: formerly was floor ... now round?

Apollon77 (commented):
Ok, I like the branded numbers better tbh ... added some more smaller comments

packages/general/src/codec/DnsCodec.ts
lauckhart: Renamed to "duration"

packages/general/src/log/Diagnostic.ts
lauckhart: changed to `Millis`

lauckhart: and changed "Nanosecs" to "Nanoseconds"

packages/matter.js/src/device/PairedNode.ts
lauckhart: K added back

packages/matter.js/src/device/PairedNode.ts
lauckhart: added back

packages/general/src/time/Time.ts
lauckhart: I added the whole rewrite as breaking...  But added a comment specifically calling this out

packages/general/src/util/DeepEqual.ts
lauckhart: yeah

packages/general/src/time/Interval.ts
lauckhart: right

packages/general/src/time/Time.ts
lauckhart: Yeah unfortunately though TS doesn't know that because we set via setter

packages/node/src/behaviors/general-diagnostics/GeneralDiagnosticsServer.ts
lauckhart: I changed `of()` to return integers so this can instead use Seconds.of()

lauckhart: Current Seconds.round() doesn't work because it works on a `Duration` rather than a number

packages/protocol/src/mdns/MdnsClient.ts
lauckhart: Yeah this is kind of a limited use but I guess that's OK

packages/general/src/net/RetrySchedule.ts
lauckhart: So e.g. `Seconds.round()` takes a `Duration`...  That is, unit manipulation is designed to work on Durations...  Which is safer than the alternative because if you took a number you could accidentally pass in a `Duration`.  I could maybe special case `Millis.{floor|ceil|round}` to take a number although I'm inclined to leave it type safe for now

packages/node/src/behaviors/switch/SwitchServer.ts
lauckhart: That does not appear to be the case.  I think it would make sense to add defaults.  How about 300ms. for multipressDelay and 1s. for longPressDelay?

packages/protocol/src/protocol/MessageChannel.ts
lauckhart: mistake

packages/protocol/src/peer/ControllerCommissioningFlow.ts
lauckhart: Alright I made a timespan class to handle this.

Do you know why we did the Math.ceil() before?  I'm inclined to just do the math in milliseconds

Apollon77: Need to check in depth. But sure all in Millis is also good

packages/node/src/behavior/system/subscription/SubscriptionBehavior.ts
lauckhart: I realized this should really be called `SubscriptionsServer`.  So I made that change (listed as breaking) and changing the ID from "subscription" to "subscriptions" solves this problem I think.

I also implemented a migration system for upgrading old data formats but didn't end up using it.

packages/node/src/behaviors/switch/SwitchServer.ts
lauckhart: Resolving as solution will be the same

packages/general/src/time/Interval.ts
lauckhart: It's unnecessary now, have taken it out

packages/node/src/behaviors/general-diagnostics/GeneralDiagnosticsServer.ts
lauckhart: I rewrote this code, should use new time helpers correctly now.

packages/node/src/behaviors/switch/SwitchServer.ts
lauckhart: I went ahead and added defaults of 300ms. and 2s. respectively.  I think that makes sense and is better than just crashing

packages/protocol/src/peer/ControllerCommissioningFlow.ts
lauckhart: OK will leave as millis

packages/node/src/behavior/system/subscriptions/SubscriptionsServer.ts
```diff
@@ -33,11 +33,11 @@ const REESTABLISH_SUBSCRIPTIONS_TIMEOUT_S = 2;
  * speed up the controller reconnection process. This can mean a bit more memory usage on start of the device. To
  * disable this feature set `persistenceEnabled` as state of the `subscription` behavior to `false`.
  */
-export class SubscriptionBehavior extends Behavior {
-    static override readonly id = "subscription";
+export class SubscriptionsBehavior extends Behavior {
```
Apollon77, line 36: Ok, maybe we should note in changelog that "persisted subscriptions" are reset once when using this new version -  all good as they will be re-established by the controllers, but the "Outage" might be longer than usual

support/codegen/src/endpoints/TypeGenerator.ts
```diff
@@ -58,6 +58,9 @@ export class TypeGenerator {
             case Metatype.date:
                 return "Date";
 
+            case Metatype.duration:
+                return "Interval";
```
Apollon77: not "Duration"?

support/codegen/src/mom/spec/translate-datatype.ts
```diff
@@ -366,6 +366,9 @@ export function selectMetatype(typeName: string) {
         case "date":
             return Metatype.date;
 
+        case "interval":
```
Apollon77: here "duration"?

support/models/src/local/interval.ts
Apollon77, line 1: rename file?

packages/general/src/time/Time.ts
Apollon77: maybe a dd acomment that this value is just temp and will be overwritten by constructor ... so it is clear

packages/node/src/behavior/system/subscription/SubscriptionBehavior.ts
Apollon77: makes sense

packages/node/src/behaviors/switch/SwitchServer.ts
Apollon77: agreed

packages/node/src/behavior/system/subscriptions/SubscriptionsServer.ts
lauckhart: ok

support/codegen/src/endpoints/TypeGenerator.ts
lauckhart: fixed

support/models/src/local/interval.ts
lauckhart: fixed

support/codegen/src/mom/spec/translate-datatype.ts
lauckhart: fixed

PR #2360, opened 2025-08-19
Fix Time us fallback logic

Apollon77:


packages/general/src/time/Time.ts
```diff
@@ -107,9 +107,9 @@ export class Time {
 
 // Check if performance API is available and has the required methods. Use lower accuracy fallback if not.
 if (!performance || typeof performance.now !== "function" || typeof performance.timeOrigin !== "number") {
-    Object.defineProperty(Time.prototype.nowUs, "nowUs", {
+    Object.defineProperty(Time.prototype, "nowUs", {
         get() {
-            return Time.nowMs; // Fallback is a bit less accurate
+            return Time.nowMs * 1000; // Fallback is a bit less accurate
```
lauckhart: I converted nowUs to a Timestamp so this line shouldn't multiply by 1000

Apollon77: ahh kk, missed that.

PR #2363, opened 2025-08-20
ClientNode support for Matter events

lauckhart:
* Add `ClientEventEmitter` and wire in to route Matter events to the appropriate observables

* Replace old `EventPriority` with generated `Priority`; leave a deprecated alias for backwards compat

* Add `Environment#block` to hide a service on the parent Environment without installing a replacement

* Add utility `Generic` types to make working with untyped events more convenient

* Implement eventing test

* Add additional tests for missing pieces

* Adds a couple of TODOs on implementation


packages/general/src/environment/Environment.ts
```diff
@@ -57,13 +63,15 @@
...
 
-        if (instance) {
+        if (instance !== undefined) {
             return instance as T;
         }
 
-        if ((type as Environmental.Factory<T>)[Environmental.create]) {
-            this.set(type, (instance = (type as any)[Environmental.create](this)));
-            return instance as T;
+        if (instance !== null) {
```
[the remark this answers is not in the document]
lauckhart: fixed bug above

packages/node/src/behavior/system/network/NetworkClient.ts
```diff
@@ -27,6 +27,7 @@ export class NetworkClient extends NetworkBehavior {
             minIntervalFloor: DEFAULT_MIN_INTERVAL_FLOOR,
             maxIntervalCeiling: 0,
             attributes: [{}],
+            events: [{}],
```
Apollon77: usually we also subscribe events with "isUrgent" to get them directly and not somewhen later. I think better to add this directly now too

packages/node/src/node/client/ClientEventEmitter.ts
```diff
@@ -0,0 +1,115 @@
...
+import { ClusterId, EndpointNumber, EventId } from "#types";
+import { ClientStructure } from "./ClientStructure.js";
+
+const logger = Logger.get("ClientEventEmitter");
+
+/**
+ * Event handler for Matter events transmitted by a peer.
+ *
+ * TODO - set priority on context when split for server vs. client
+ * TODO - record latest event number for each subscription shape (or just wildcard?)
```
Apollon77, line 21: 99% of the cases will only have the wildcard one ... I would record max event number "globally" until we see that this is not sufficient

lauckhart: OK maybe just for wildcard then?  In other places we aren't particularly optimized for non-wildcard subscriptions but I did put some effort into ensuring that we don't do the wrong thing, which a global value would do in the case of multiple event subscriptions

PR #2369, opened 2025-08-21
Adds BDX Protocol implementation

Apollon77:
This PR mainly adds the BDX protocol implementation. We support most of the features from the specification with the following notes:
* Even that Matter allows 64bit values for size and offset, we support up to MAX_SAFE_INTEGER for size and offset (which basically is 2^53 - 1 and so far enough for us).
* We support partial transfers (startOffset or shorter dataLength) only when we act as the sender. As a receiver,
  only full transfers are supported - the main reason was that the Streams in JavaScript can not easiely "seek" to other positions in the file, this only works with a FileSystemStream in Node.js (as example). Also because I doubt that really anyone uses or expect this we decline such transfers (as allowed by spec) nd can implement whenever needed
* We do not use BlockQueryWithSkip when requesting data ourselves (See above)

Additionally the PR also optimizes an edge case in the DataReadQueue and ensures that StandaloneAcks are always only considered for Secure Channel protocol.

BDX also got extensive tests.

it is also notable that this is currently not yet used anywhere, so it is the pure functionality.

The use cases for later steps are:
* Devices trigger transfers from OTA provider cluster. Device does ReceiveInit to OTA provider (server) with the URL it got from a QueryImage command
* Logs upload: Device does a SendInit to the node that requested the logs (RetrieveLogsRequest on Diagnostic Logs cluster with protocol of BDX)

So maybe more adjustments (like Filedesignator-mapping) or such needs to be added later then

### DIFF
--- a/packages/protocol/src/bdx/BdxMessenger.ts
+++ b/packages/protocol/src/bdx/BdxMessenger.ts
@@ -271,11 +271,15 @@ export class BdxMessenger {
     }
 
     /**
-     * Ensure that the value is a safe number and that it is not too large. Matter spec allows also 64bit values, but they
-     * make little sense for now, so make sure we handle them as too large. MAX_SAFE_INTEGER is 2^53-1 and is
-     * enough for now.
+     * Ensure that the value is a safe JavaScript "number" type and that it is not too large. Matter spec allows also
+     * 64bit values, but they make little sense for now, so make sure we handle them as too large. MAX_SAFE_INTEGER is
+     * 2^53-1 and is enough for now.
      */
-    asSafeNumber(value: number | bigint | undefined, context = "", bdxErrorCode = BdxStatusCode.Unknown): number {
+    static asSafeNumber(
+        value: number | bigint | undefined,
+        context = "",
+        bdxErrorCode = BdxStatusCode.Unknown,
+    ): number {
         if (typeof value !== "number" && typeof value !== "bigint") {
             throw new InternalError(`${context} ${value} is not a number`); // Should not happen
         }
--- a/packages/protocol/src/bdx/schema/BdxMessage.ts
+++ b/packages/protocol/src/bdx/schema/BdxMessage.ts
@@ -53,7 +53,7 @@ export namespace BdxMessage {
         }
     }
 
-    export const Kinds: { [key: number]: Schema<any, Bytes> } = {
+    export const Kinds: { [key: number]: Schema<any> } = {
         [BdxMessageType.SendInit]: BdxSendInitMessage,
         [BdxMessageType.ReceiveInit]: BdxReceiveInitMessage,
         [BdxMessageType.SendAccept]: BdxSendAcceptMessage,
--- a/packages/protocol/src/bdx/bdx-session-initiator.ts
+++ b/packages/protocol/src/bdx/bdx-session-initiator.ts
@@ -56,7 +56,7 @@ export async function bdxSessionInitiator(messenger: BdxMessenger, config: BdxSe
 
         let { startOffset } = initMessage;
         if (startOffset !== undefined) {
-            startOffset = messenger.asSafeNumber(startOffset, "Start offset", BdxStatusCode.StartOffsetNotSupported);
+            startOffset = BdxMessenger.asSafeNumber(startOffset, "Start offset", BdxStatusCode.StartOffsetNotSupported);
         }
 
         const acceptMessage = await determineAcceptParameters(initMessage);
@@ -181,7 +181,7 @@ export async function bdxSessionInitiator(messenger: BdxMessenger, config: BdxSe
             maxBlockSize = requestedMaxBlockSize;
         }
 
-        const maxLength = messenger.asSafeNumber(initMaxLength, "Max length", BdxStatusCode.LengthTooLarge);
+        const maxLength = BdxMessenger.asSafeNumber(initMaxLength, "Max length", BdxStatusCode.LengthTooLarge);
         // TODO: How to handle custom metadata?
 
         const transferControl: TypeFromPartialBitSchema<typeof BdxTransferControlBitmap> = {
@@ -219,7 +219,7 @@ export async function bdxSessionInitiator(messenger: BdxMessenger, config: BdxSe
 
         const dataLength =
             "length" in acceptMessage && acceptMessage.length !== undefined
-                ? messenger.asSafeNumber(
+                ? BdxMessenger.asSafeNumber(
                       acceptMessage.length,
                       "Accept message length field",
                       BdxStatusCode.LengthTooLarge,
--- a/packages/protocol/src/bdx/flow/FollowingReceivingFlow.ts
+++ b/packages/protocol/src/bdx/flow/FollowingReceivingFlow.ts
@@ -6,6 +6,7 @@
 
 import { Logger } from "#general";
 import { BdxMessageType, BdxStatusCode } from "#types";
+import { BdxMessenger } from "../BdxMessenger.js";
 import { BdxMessage } from "../schema/BdxMessage.js";
 import { OutboundFlow } from "./OutboundFlow.js";
 
@@ -23,7 +24,7 @@ export class FollowingReceivingFlow extends OutboundFlow {
         // Read the data query from the peer
         const blockQuery = await this.messenger.readBlockQuery();
         const bytesToSkip = BdxMessage.is(blockQuery, BdxMessageType.BlockQueryWithSkip)
-            ? this.messenger.asSafeNumber(
+            ? BdxMessenger.asSafeNumber(
                   blockQuery.message.bytesToSkip,
                   "BytesToSkip",
                   BdxStatusCode.TransferFailedUnknownError,
--- a/packages/protocol/src/bdx/schema/BdxAcceptMessagesSchema.ts
+++ b/packages/protocol/src/bdx/schema/BdxAcceptMessagesSchema.ts
@@ -47,7 +47,7 @@ export interface BdxReceiveAccept extends BdxSendAccept {
 }
 
 /** Schema for BDX SendAccept message */
-export class BdxSendAcceptMessageSchema<M extends BdxSendAccept> extends Schema<M, Bytes> {
+export class BdxSendAcceptMessageSchema<M extends BdxSendAccept> extends Schema<M> {
     encodeInternal(message: M) {
         this.validate(message);
         const { transferControl, maxBlockSize, metaData } = message;
--- a/packages/protocol/src/bdx/schema/BdxBlockMessagesSchema.ts
+++ b/packages/protocol/src/bdx/schema/BdxBlockMessagesSchema.ts
@@ -34,7 +34,7 @@ export interface BdxBlock extends BdxCounterOnly {
 export interface BdxBlockEof extends BdxBlock {}
 
 /** Schema for BDX messages that only contain a block counter: BlockQuery, BlockAck, BlockAckEof. */
-export class BdxCounterOnlyMessageSchema extends Schema<BdxCounterOnly, Bytes> {
+export class BdxCounterOnlyMessageSchema extends Schema<BdxCounterOnly> {
     encodeInternal(message: BdxCounterOnly) {
         const { blockCounter } = message;
 
@@ -56,7 +56,7 @@ export const BdxBlockAckMessage = BdxCounterOnlyMessage;
 export const BdxBlockAckEofMessage = BdxCounterOnlyMessage;
 
 /** Schema for BDX BlockQueryWithSkip message. */
-export class BdxBlockQueryWithSkipMessageSchema extends Schema<BdxBlockQueryWithSkip, Bytes> {
+export class BdxBlockQueryWithSkipMessageSchema extends Schema<BdxBlockQueryWithSkip> {
     encodeInternal(message: BdxBlockQueryWithSkip) {
         const { blockCounter, bytesToSkip } = message;
 
@@ -77,7 +77,7 @@ export class BdxBlockQueryWithSkipMessageSchema extends Schema<BdxBlockQueryWith
 export const BdxBlockQueryWithSkipMessage = new BdxBlockQueryWithSkipMessageSchema();
 
 /** Schema for BDX Block and BlockEof message that contain Block data. */
-export class BdxBlockMessageSchema extends Schema<BdxBlock, Bytes> {
+export class BdxBlockMessageSchema extends Schema<BdxBlock> {
     encodeInternal(message: BdxBlock) {
         const { blockCounter, data } = message;
 
--- a/packages/protocol/src/bdx/schema/BdxInitMessagesSchema.ts
+++ b/packages/protocol/src/bdx/schema/BdxInitMessagesSchema.ts
@@ -104,7 +104,7 @@ export type BdxInit = {
 };
 
 /** Schema for BDX SendInit and ReceiveInit messages. */
-export class BdxInitMessageSchema extends Schema<BdxInit, Bytes> {
+export class BdxInitMessageSchema extends Schema<BdxInit> {
     encodeInternal(message: BdxInit) {
         this.validate(message);
         const { transferProtocol, maxBlockSize, startOffset, maxLength, fileDesignator, metaData } = message;
--- a/packages/protocol/src/protocol/ProtocolStatusMessage.ts
+++ b/packages/protocol/src/protocol/ProtocolStatusMessage.ts
@@ -15,7 +15,7 @@ export type ProtocolStatusMessage<T> = {
     protocolData?: Bytes;
 };
 
-export abstract class ProtocolStatusMessageSchema<T extends ProtocolStatusMessage<any>> extends Schema<T, Bytes> {
+export abstract class ProtocolStatusMessageSchema<T extends ProtocolStatusMessage<any>> extends Schema<T> {
     #protocolId: number;
     #vendorId: number;
     #protocolSpecificDataAllowed: boolean;
--- a/packages/types/src/schema/BitmapSchema.ts
+++ b/packages/types/src/schema/BitmapSchema.ts
@@ -106,7 +106,7 @@ export class BitmapSchemaInternal<T extends BitSchema> extends Schema<TypeFromBi
     }
 }
 
-export class ByteArrayBitmapSchemaInternal<T extends BitSchema> extends Schema<TypeFromBitSchema<T>, Bytes> {
+export class ByteArrayBitmapSchemaInternal<T extends BitSchema> extends Schema<TypeFromBitSchema<T>> {
     private readonly byteArrayLength: number;
     private readonly maskOffset: MaskOffsetFromBitSchema<T>;
 
--- a/packages/types/src/schema/Schema.ts
+++ b/packages/types/src/schema/Schema.ts
@@ -4,8 +4,10 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+import { Bytes } from "#general";
+
 /** Define a schema to encode / decode convert type T to type E. */
-export abstract class Schema<T, E> {
+export abstract class Schema<T, E = Bytes> {
     /** Encodes the value using the schema. */
     encode(value: T): E {
         this.validate(value);
--- a/packages/types/src/tlv/TlvSchema.ts
+++ b/packages/types/src/tlv/TlvSchema.ts
@@ -22,7 +22,7 @@ export type TlvEncodingOptions = {
     allowMissingFieldsForNonFabricFilteredRead?: boolean;
 };
 
-export abstract class TlvSchema<T> extends Schema<T, Bytes> implements TlvSchema<T> {
+export abstract class TlvSchema<T> extends Schema<T> implements TlvSchema<T> {
     override decodeInternal(encoded: Bytes): T {
         return this.decodeTlvInternal(new TlvByteArrayReader(encoded)).value;
     }


packages/protocol/src/bdx/schema/BdxAcceptMessagesShema.ts
```diff
@@ -0,0 +1,159 @@
+/**
```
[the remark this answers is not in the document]
Apollon77: valid, will be fixed

packages/protocol/src/bdx/FileDesignator.ts
```diff
@@ -0,0 +1,43 @@
...
+        return this.#fd;
+    }
+
+    /** Return the textual/string representation of the FileDesignator. */
+    get text(): string {
+        const fileDesignatorData = Bytes.of(this.#fd);
+        // When all uint8 values are in char() range "a-z0-0-." then use this as the blob name, else hex encode it
+        const isValidName = fileDesignatorData.every(
+            byte =>
+                (byte >= 0x11 && byte <= 0x5a) || // A-Z
```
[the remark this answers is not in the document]
Apollon77: valid, will be fixed

packages/protocol/src/bdx/BdxMessenger.ts
```diff
@@ -0,0 +1,295 @@
...
+        const acceptMessage = await this.nextMessage([BdxMessageTypes.ReceiveAccept]);
+        const result = BdxReceiveAcceptMessage.decode(acceptMessage.payload);
+        logger.debug("Received Bdx ReceiveAccept", result);
+        return result;
+    }
+
+    /** Encodes and sends a Bdx SendAccept message. */
+    async sendSendAccept(message: BdxSendAccept) {
+        logger.debug("Sending Bdx SendAccept", message);
+        await this.exchange.send(BdxMessageTypes.SendAccept, BdxSendAcceptMessage.encode(message));
```
lauckhart: I might generate these methods but I like the clean explicit layout.  It does feel like you could create a message type <-> schema mapping and have encode/decode handled automatically...

Encode seems easy as it's just lookup and no types necessary.  And integrated with in a send wrapper it could be invisible

Decode -- not here but elsewhere -- would be slightly more complicated but you could parse automatically and with a big discriminated union you could get types for free after checking message type

Buut fine as future TODO or as is, certainly easy to follow

packages/protocol/src/bdx/BdxMessenger.ts
```diff
@@ -0,0 +1,295 @@
...
+    get maxPayloadSize() {
+        return this.#exchange.maxPayloadSize;
+    }
+
+    /**
+     * Waits for the next message and returns it.
+     * A List of allowed expected message types can be provided.
+     * If the message type is not in the list, an error will be thrown.
+     */
+    async nextMessage(expectedMessageTypes: number[], timeout = this.#messageTimeout, expectedMessageInfo?: string) {
```
lauckhart: Why just number here?

packages/protocol/src/bdx/BdxMessenger.ts
```diff
@@ -0,0 +1,295 @@
...
+        const message = await this.exchange.nextMessage({ timeout });
+        const messageType = message.payloadHeader.messageType;
+        if (expectedMessageTypes !== undefined && expectedMessageInfo === undefined) {
+            expectedMessageInfo = expectedMessageTypes.map(t => `${t} (${BdxMessageTypes[t]})`).join(",");
+        }
+        this.throwIfErrorStatusReport(message, expectedMessageInfo);
+        if (expectedMessageTypes !== undefined && !expectedMessageTypes.includes(messageType))
+            throw new UnexpectedDataError(
+                `Received unexpected message type: ${messageType}, expected: ${expectedMessageInfo}`,
```
lauckhart: Could include the textual name here, e.g `${BdxMessageType[messageType] ?? "unknown"}#${messageType}` or similar...  Maybe in a helper function so it's formatted consistently everywhere

packages/protocol/src/bdx/BdxMessenger.ts
```diff
@@ -0,0 +1,295 @@
...
+        logger.debug("Waiting for Bdx Block");
+        const message = await this.nextMessage([BdxMessageTypes.Block, BdxMessageTypes.BlockEof]);
+        const messageType = message.payloadHeader.messageType;
+        const block = BdxBlockMessage.decode(message.payload);
+        if (messageType !== BdxMessageTypes.BlockEof && block.data.byteLength === 0) {
+            // empty block only allowed in BlockAckEof
+            throw new BdxError("Received empty data in Block message", BdxStatusCode.BadMessageContent);
+        }
+        logger.debug(`Received Bdx Block with ${block.data.byteLength} bytes`, block);
+        return { ...block, messageType };
```
lauckhart: This is where a discriminated union would be handy, so caller could switch on messageType to get the proper type without cast.

packages/protocol/src/bdx/BdxProtocol.ts
```diff
@@ -0,0 +1,90 @@
...
+
+    async onNewExchange(exchange: MessageExchange, message: Message) {
+        const initMessageType = message.payloadHeader.messageType;
+        const messenger = new BdxMessenger(exchange);
+
+        switch (initMessageType) {
+            case BdxMessageTypes.SendInit:
+            case BdxMessageTypes.ReceiveInit:
+                logger.debug(
+                    `Initialize Session for ${BdxMessageTypes[initMessageType]} message on BDX protocol for exchange ${exchange.id}`,
```
lauckhart: Should BdxMessageTypes be singular?

packages/protocol/src/bdx/BdxProtocol.ts
```diff
@@ -0,0 +1,90 @@
...
+
+                try {
+                    await bdxSession.processTransfer();
+                } catch (error) {
+                    logger.error(`Error processing BDX transfer:`, error);
+                }
+
+                break;
+            default:
+                logger.warn(`BDX session for exchange ${exchange.id} already exists, sending error`);
```
lauckhart: It's not clear to me how the default case in this switch statement gets to this message

packages/protocol/src/bdx/BdxMessenger.ts
```diff
@@ -0,0 +1,295 @@
...
+    close() {
+        return this.#exchange.close();
+    }
+
+    /**
+     * Ensure that the value is a number and that it is not too large. Matter spec allows also 64bit values, but they
+     * make little sense for now, so make sure we handle them as too large. MAX_SAFE_INTEGER is 2^53-1 and is
+     * enough for now.
+     */
+    asNumber(value: number | bigint | undefined, context = "", bdxErrorCode = BdxStatusCode.Unknown): number {
```
lauckhart: Maybe call validatedSize or something related to purpose of number?

packages/protocol/src/bdx/BdxMessenger.ts
```diff
@@ -0,0 +1,295 @@
...
+        if (typeof value !== "number" && typeof value !== "bigint") {
+            throw new InternalError(`${context} ${value} is not a number`); // Should not happen
+        }
+        if (value > Number.MAX_SAFE_INTEGER) {
```
lauckhart, line 286: If we could enforce maxTransferSize here we could avoid redundant checks/error message in other places

packages/protocol/src/bdx/FileDesignator.ts
```diff
@@ -0,0 +1,43 @@
...
+                (byte >= 0x41 && byte <= 0x5a) || // A-Z
+                (byte >= 0x61 && byte <= 0x7a) || // a-z
+                (byte >= 0x30 && byte <= 0x39) || // 0..9
+                byte === 0x2e || // "."
+                byte === 0x2d, // "-"
+        );
+        if (isValidName) {
+            return fileDesignatorData.reduce((name, byte) => name + String.fromCharCode(byte), "");
+        } else {
+            return Bytes.toHex(fileDesignatorData);
```
lauckhart: Maybe need a protected prefix on names so e.g. "\x12" and "12" don't map to same file?

packages/protocol/test/bdx/bdx-mock-exchange.ts
```diff
@@ -0,0 +1,108 @@
...
+/**
+ * A mock message exchange for the BDX protocol.
+ *
+ * This provides an intermediate level of mock communication, higher level than network mocking but lower level than
+ * direct method calls.
+ *
+ * Our mocking of the message exchange context is a bit half assed, we don't use a real channel or session and just
+ * stub out methods as necessary.
+ */
+export class MockExchange extends MessageExchange {
```
lauckhart, line 42: Is this similar enough to the other mock exchange that we could make it part of package like MockNetwork and have it be shared?

packages/protocol/src/bdx/BdxSession.ts
```diff
@@ -0,0 +1,575 @@
...
+    #messenger: BdxMessenger;
+    #isSender: boolean;
+    #isInitiator: boolean;
+    #initMessage?: BdxInit;
+    #started = false;
+    #closed = AsyncObservable();
+    #isClosed = false;
+
+    /** Configuration for transfer behavior */
+    #bdxConfig: BdxSession.Config = {
```
lauckhart: Exporting as a constant near the Options interface would surface defaults a little more clearly

packages/protocol/src/bdx/BdxSession.ts
```diff
@@ -0,0 +1,575 @@
...
+ * Notes:
+ * * Even though Matter allows 64bit values for size and offset, we do not use them, as they make no sense for now.
+ *   We support up to MAX_SAFE_INTEGER for size and offset (which basically is 2^53 - 1 and so far enough for us).
+ * * We support partial transfers (startOffset or shorter dataLength) only when we act as the sender. As a receiver,
+ *   only full transfers are supported.
+ * * If data are queried with skipping some data (BlockQueryWithSkip) then we might send out a message shorter than the
+ *   blockSize which is ok according to spec.
+ * * We do not use BlockQueryWithSkip when requesting data ourselves
+ */
+export class BdxSession {
```
lauckhart, line 34: This is a large class...  It's nice with the flows extracted...  A few potential other components stand out for extraction:

- BdxSessionConfiguration - covers the stuff in the constructor and the type definitions int he BdxSession namespace
- BdxSessionInitiator - covers the bulk of the code in private methods here to do setup, before invoking `this.#processTransferFlow()` (which would then only be invoked in one place)
- Could also decouple persistence, although that's perhaps more for separation of concerns than for readability

packages/protocol/src/bdx/flow/BdxTransferFlow.ts
```diff
@@ -0,0 +1,75 @@
...
+ * @license
+ * Copyright 2022-2025 Matter.js Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { BdxError, BdxMessenger } from "#bdx/index.js";
+import { BdxStatusCode } from "#types";
+
+/** Base class for BDX transfer flows. */
+export abstract class BdxTransferFlow {
```
lauckhart: Cool abstraction.  "Transfer" seems a little redundant with "Flow", maybe could just call "BdxFlow"?  Mainly because it'd make the subclass names less noisy.  I'd leave off Bdx too given directory layout but that feels more like personal preference.

packages/protocol/src/bdx/flow/BdxDrivenSendingTransferFlow.ts
```diff
@@ -0,0 +1,51 @@
...
+            if (this.isClosed) {
+                return; // Session is closed, so we stop processing
+            }
+
+            const blockCounter = this.nextMessageCounter;
+            // Read the next data chunk from the storage
+            ({ data, done, bytesLeft } = await this.readDataChunk(reader, blockSize, bytesLeft, dataLength));
+
+            if (done) {
+                // Send the last Block and wait for AckEof and close down
```
lauckhart: I'd just have the break here, put other stuff after the loop, or see my comment in BdxDrivingReceivingTransferFlow.ts for slightly higher-level organization

packages/protocol/src/bdx/flow/BdxDrivingReceivingTransferFlow.ts
```diff
@@ -0,0 +1,48 @@
...
+
+import { BdxWritingTransferFlow } from "./BdxWritingTransferFlow.js";
+
+/**
+ * BDX Transport flow logic for a "BDX Driving Receiver":
+ * - Requests the next Blow using BlockQuery and then reads the Block
+ * - Last message is confirmed with an AckEof
+ */
+export class BdxDrivingReceivingTransferFlow extends BdxWritingTransferFlow {
+    async processTransfer() {
```
lauckhart: An alternate design for these flows could be methods in the base class:

```ts
// Stuff in processTransfer() before while (true) updates local vars here
abstract initTransfer(): void;

// Stuff inside while loop goes here (except check for isClosed goes in base class); returns boolean indicating whether loop should break
abstract transferNextChunk(): Promise<boolean>;

// Stuff in if (done) goes here
abstract finalizeTransfer(): Promise<void>;
```

...thinking being that you have the loop and check for isClosed() in the base class and then it's easy to see the transfer stages in each flow implementation.

packages/protocol/src/bdx/flow/BdxFollowingSendingTransferFlow.ts
```diff
@@ -0,0 +1,46 @@
...
+ */
+
+import { BdxWritingTransferFlow } from "./BdxWritingTransferFlow.js";
+
+/**
+ * BDX Transport flow logic for a "BDX Following Sender":
+ * - Reads the Block and sends an Ack (if synchronous transfer is used)
+ * - Last message is confirmed with an AckEof
+ */
+export class BdxFollowingSendingTransferFlow extends BdxWritingTransferFlow {
```
lauckhart: I was a little confused by "reading" vs "writing", maybe could call base classes "inbound" and "outbound" instead?

Part of the my confusion I think comes from a transfer flow that is "followingsending" being named from the perspective of the peer, while "writingtransferflow" is named from the perspective of the server.  But "inbound" to me makes it clear data moves from the peer.

Unless I'm still confused. ;)

packages/protocol/src/bdx/flow/BdxReadingTransferFlow.ts
```diff
@@ -0,0 +1,118 @@
...
+            // Done needs to be true when value is undefined or there is something broken
+            if (!done) {
+                throw new BdxError(
+                    `Data length too short, expected ${blockSize}bytes, but got less`,
+                    BdxStatusCode.LengthTooShort,
+                );
+            }
+            value = new Uint8Array(); // Simulate an empty value when we reached the end of the stream
+        } else if (value.byteLength < blockSize) {
+            // When we git less data than blocksize it is the last block, so validate that
```
lauckhart: "get" :)

packages/protocol/src/protocol/MessageExchange.ts
```diff
@@ -88,6 +88,9 @@ export type ExchangeSendOptions = {
 
     /** Additional context information for logging to be included at the beginning of the Message log. */
     logContext?: ExchangeLogContext;
+
+    /** Allows to override the protocol ID of the message, mainly used for standalone acks. */
+    protocolId?: number;
```
lauckhart, line 95: It's weird to me that the message exchange is protocol independent except for this one feature.  Would it be possible to create a `SecureProtocolMessageExchange` that adds standalone ack support?  Or to do it with composition, have some sort of plugin component with hooks in all of the places where it's wired in?

I know standalone ack is pretty well intertwined with the code here so maybe better as a future TODO.

lauckhart (approved):

packages/protocol/src/bdx/BdxMessenger.ts
Apollon77: I tried around  a lot but without changing a lot of things - including how Schema works and such - is was not really able to come up with anything better than https://github.com/matter-js/matter.js/pull/2369/commits/ddc7a1fb9153411fbfc25d44852051fb6c5d7583 ... Yes it streamlines that class a bit more but in fact for the cost of several "as unknown as Type" things because in fact especially when reading the multiple options are meehh.

In fact logic and code wise it is still ok because all the single methods act kind of as type guard so that internally such type castings are no risk ... WDYT?

packages/protocol/src/bdx/BdxMessenger.ts
Apollon77: valid, fixed

packages/protocol/src/bdx/BdxProtocol.ts
Apollon77: makes sense

packages/protocol/src/bdx/BdxProtocol.ts
Apollon77: fixed

packages/protocol/src/bdx/BdxMessenger.ts
Apollon77: asSafeNumber ?!

packages/protocol/src/bdx/BdxMessenger.ts
Apollon77: The method is not only used for maxTransferSize and mainly used to get the bigint case out of the way. Or what did you mean? There are also not that many checks for maxTransferSize

packages/protocol/src/bdx/FileDesignator.ts
Apollon77: valid

Apollon77: "__hex__" ?? or do you think "0x" as prefix is enough

packages/protocol/src/bdx/flow/BdxTransferFlow.ts
Apollon77: killed both. just left Bdx prefix in one place for logger

packages/protocol/src/bdx/flow/BdxDrivingReceivingTransferFlow.ts
Apollon77: That worked good :-) Thanks

packages/protocol/src/bdx/flow/BdxFollowingSendingTransferFlow.ts
Apollon77: Yeeahh I was confused too but because of the structure of the BDX protocol with the "Driver" vs "Follower" and "Sender" vs "Receiver" things get mixed. Also Inbound/Outbound is not really helping, because still leave interpretation

My Idea with "Reading" and "Writing" was to more exact say what "we do" ... are we "Reading" data from the storage or are we "writing" data into the Storage.

packages/protocol/src/protocol/MessageExchange.ts
Apollon77: Yeahh I tink it is because the Standalone Ack is part of the MRP feature of the "underlaying protocol scheme". So SecureProtocol is the underlaying scheme. BDX is more like an "Application protocol" which sits on top of a Secure Channel layer.
This means that even if you process BDX messages the underlaying MRP layer belongs to SecureChannel.
I will add a TODO. In fact we might need to get the "protocol binding" out of the exchange and more formalize the messages we send because in fact the messages then are "protocol bound" and not the exchange.

packages/protocol/src/bdx/BdxSession.ts
Apollon77: Lets see what you say to this :-) Large refactoring incoming

packages/protocol/src/bdx/FileDesignator.ts
lauckhart: 0x seems fine

packages/protocol/src/bdx/schema/BdxMessage.ts
```diff
@@ -0,0 +1,70 @@
...
+        kind: T,
+    ): asserts message is BdxMessage<T> {
+        if (message.kind !== kind) {
+            throw new InternalError(
+                `Expected BDX message of type ${BdxMessageType[kind]}, but got ${BdxMessageType[message.kind]}(${message.kind})`,
+            );
+        }
+    }
+
+    export const Kinds: { [key: number]: Schema<any, Bytes> } = {
```
lauckhart: Could maybe make "any" and "Bytes" be default generic parameters?

packages/protocol/src/bdx/BdxMessenger.ts
```diff
@@ -0,0 +1,287 @@
...
+    close() {
+        return this.#exchange.close();
+    }
+
+    /**
+     * Ensure that the value is a safe number and that it is not too large. Matter spec allows also 64bit values, but they
+     * make little sense for now, so make sure we handle them as too large. MAX_SAFE_INTEGER is 2^53-1 and is
+     * enough for now.
+     */
+    asSafeNumber(value: number | bigint | undefined, context = "", bdxErrorCode = BdxStatusCode.Unknown): number {
```
lauckhart: Maybe `asSafeInteger` for naming as there is no notion of "safe" otherwise.  Or could be "asByteLength" or something as that's what it's used for.

I would also make it static as I don't think there's any references to local fields.  I think this confused me before because I assumed it was used locally

packages/protocol/src/bdx/BdxSession.ts
lauckhart: I like it

lauckhart (approved):

packages/protocol/src/bdx/schema/BdxMessage.ts
Apollon77: I pin the Bytes

packages/protocol/src/bdx/BdxMessenger.ts
Apollon77: As discussed in Discord we leave as is

PR #2409, opened 2025-09-07
Proper split for local/remote actor context

lauckhart:
Previously ActionContext and the inherited fields from AccessControl.Session were poorly typed to handle the differentiation between "offline" and "online" contexts.  This worked OK for ServerNode but for ClientNode it was more ugly because the "online" fields were never appropriate even when accessing an "online" remote node.

So now instead of the somewhat unclear "OnlineContext" and "OfflineContext" we have:

* LocalActorContext - used for local operations on ServerNode and remote operations on ClientNode

* RemoteActorContext - used for operations triggered by peers

* ActionContext - converted from an interface to a union of above two types

* AccessControl.Session - this is also converted to a union of local vs remote variants

* isLocalActor/isRemoteActor/assertRemoteActor - type guards for working with the local vs. remote variants above

Includes various small cleanups, some enabled by above and some just opportunistic.


packages/node/src/behavior/context/server/RemoteActorContext.ts
```diff
@@ -163,13 +179,13 @@ export function OnlineContext(options: OnlineContext.Options) {
...
         }
-        const context: ActionContext = {
+        const context: RemoteActorContext & T = {
             ...options,
             session,
             exchange,
             subject,
 
-            fabric,
+            fabric: fabric?.fabricIndex ?? FabricIndex.NO_FABRIC,
```
Apollon77, line 188: Did you checked all cases where maybe foo.fabric is checked for undefined?

Apollon77 (approved):

Apollon77:
@lauckhart ok seems the pr introduced a regression. Window covering attr "more" is indeed writable but the failed test states it is not. So something seems odd

packages/node/src/behavior/context/server/RemoteActorContext.ts
lauckhart: I tried to, yes

PR #2426, opened 2025-09-12
Remove BLE singleton

lauckhart:
* Add `ServiceBundle` component to function as a lightweight plugin system for `Environment`

* Remove the `Ble.get()` singleton

* For nodejs-ble, Ble now registers via `ServiceBundle.default`

* For react-native, Ble registers directly in `ReactNativeEnvironment`

* Update apps, examples and docs accordingly

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,7 +26,7 @@ The main work (all changes without a GitHub username in brackets in the below li
 -   @matter/protocol
     - Breaking: The platform-specific BLE abstraction has changed so that higher-level logic may be shared across platforms
     - Breaking: Low-level advertising APIs have changed significantly; in particular, `MdnsBroadcaster`, `MdnsInstanceBroadcaster` and `MdnsScanner` are replaced by `MdnsServer`, `MdnsAdvertisement` and `MdnsClient`
-    - Breaking: The `Ble.get()` singleton is removed; components now instead retrieve the `Ble` service form the environment
+    - Breaking: The `Ble.get()` singleton is removed; components now instead retrieve the `Ble` service from the environment
     - Feature: Adds support for advertising of TCP and ICD services (but matter.js does not yet implement those features otherwise)
     - Feature: Adds support for extended advertisement
     - Feature: Added support for Case Authenticated Tags (CATs) in operational CASE sessions for enhanced access control
--- a/packages/nodejs-ble/README.md
+++ b/packages/nodejs-ble/README.md
@@ -25,7 +25,7 @@ The used packages have some limitations and prerequisites. Please check the Read
 > [TIP] For Linux: The :`make` and `g++` packages must be installed.
 
 ## How to use this package
-This package is adds BLE support to matter.js on Node.js. It is not intended to be used standalone.
+This package adds BLE support to matter.js on Node.js. It is not intended to be used standalone.
 
 Add this package to your package dependencies and make sure it is exactly the same version as matter.js and matter-node.js.
 
@@ -33,6 +33,12 @@ In your code you register the BLE instance by importing the package:
 
 ```ts
 import "@matter/nodejs-ble";
+```
+
+In addition to importing the module, you also need to explicitly enable BLE:
+
+```ts
+Environment.default.vars.set("ble.enable", true);
 ```
 
 If you need to modify the HCI ID you can do so using a configuration variable:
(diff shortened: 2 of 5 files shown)


CHANGELOG.md
```diff
@@ -26,6 +26,7 @@ The main work (all changes without a GitHub username in brackets in the below li
 -   @matter/protocol
     - Breaking: The platform-specific BLE abstraction has changed so that higher-level logic may be shared across platforms
     - Breaking: Low-level advertising APIs have changed significantly; in particular, `MdnsBroadcaster`, `MdnsInstanceBroadcaster` and `MdnsScanner` are replaced by `MdnsServer`, `MdnsAdvertisement` and `MdnsClient`
+    - Breaking: The `Ble.get()` singleton is removed; components now instead retrieve the `Ble` service form the environment
```
Apollon77:
```suggestion
    - Breaking: The `Ble.get()` singleton is removed; components now instead retrieve the `Ble` service from the environment
```

packages/examples/src/controller/ControllerNode.ts
```diff
@@ -11,29 +11,20 @@
...
+import { Diagnostic, Environment, Logger, StorageService, Time } from "@matter/main";
 import { BasicInformationCluster, DescriptorCluster, GeneralCommissioning, OnOff } from "@matter/main/clusters";
-import { Ble, ClusterClientObj } from "@matter/main/protocol";
+import { ClusterClientObj } from "@matter/main/protocol";
 import { ManualPairingCodeCodec, NodeId } from "@matter/main/types";
-import { NodeJsBle } from "@matter/nodejs-ble";
 import { CommissioningController, NodeCommissioningOptions } from "@project-chip/matter.js";
 import { NodeStates } from "@project-chip/matter.js/device";
 
+// This installs BLE support if configuration variable "ble.enable" is true
```
Apollon77, line 21: Does also ControllerNode Readme needs to be updated for the new variable?

packages/nodejs-ble/README.md
```diff
@@ -25,28 +25,27 @@ The used packages have some limitations and prerequisites. Please check the Read
 > [TIP] For Linux: The :`make` and `g++` packages must be installed.
 
 ## How to use this package
-This package is designed to be used in addition to matter.ja and matter-node.js to enable BLE support. It is not intended to be used standalone.
+This package is adds BLE support to matter.js on Node.js. It is not intended to be used standalone.
```
Apollon77:
```suggestion
This package adds BLE support to matter.js on Node.js. It is not intended to be used standalone.
```

packages/nodejs-ble/README.md
```diff
@@ -25,28 +25,27 @@ The used packages have some limitations and prerequisites. Please check the Read
...
-import { BleNode } from "@matter/nodejs-ble";
-import { Ble } from "@matter/main";
-
-Ble.get = singleton(
-    () =>
-        new BleNode({
-          // hciId: 1, // If you need to change it from default 0 on linux systems
-        }),
-);
+In your code you register the BLE instance by importing the package:
```
Apollon77, line 32: Maybe add that also ble.enable needs to bet set before?

packages/nodejs-shell/src/app.ts
```diff
@@ -147,22 +146,17 @@ async function main() {
...
-                        () =>
-                            new NodeJsBle({
-                                environment: Environment.default,
-                                hciId,
-                            }),
-                    );
+                    Environment.default.vars.set("ble.enable", true);
                 }
 
+                ServiceBundle.default.deploy(Environment.default);
```
Apollon77: Why this is needed?

packages/nodejs-ble/README.md
lauckhart: got it

packages/nodejs-ble/README.md
lauckhart: Added docs.  Ideally we wouldn't disable by default but could be a separate PR.

packages/examples/src/controller/ControllerNode.ts
lauckhart: Yes, only place I saw was "--ble" to "--ble-enable"

packages/nodejs-shell/src/app.ts
lauckhart: It was necessary because the BLE package needed to initialize after "ble.enable" is set.

But I changed so it will reinitialize if the variable changes.  So no longer necessary.

PR #2440, opened 2025-09-18
Downgrade chai-as-promised because ESM only

Apollon77:
And breaks at least my Webstorm testing integration, so downgrade for now. I can test around next to get it upgraded again, but easier for now.


lauckhart (approved):
Is also possible we don't need chai-as-promised anymore, haven't looked to see if chai handles promises natively now

Apollon77:
I added a yaml test fix for change of https://github.com/project-chip/connectedhomeip/pull/36311 to the PR

PR #2469, opened 2025-09-26
Optimize "Offline during Factory reset" case

Apollon77:
if the node should go offline while being in a factory reset we can handle that and prevent the node from starting.

Ithink this is the reason for that one test to be flaky, lets see


packages/node/src/node/Node.ts
```diff
@@ -38,6 +38,7 @@ const logger = Logger.get("Node");
 export abstract class Node<T extends Node.CommonRootEndpoint = Node.CommonRootEndpoint> extends Endpoint<T> {
     #environment: Environment;
     #runtime?: NetworkRuntime;
+    #targetState: "online" | "offline" = "offline";
```
lauckhart: I think this makes sense but it should go into NodeLifecycle, that's where all the state management cruft is

lauckhart (approved):
LGTM except see one comment about where to store target state

packages/node/src/node/Node.ts
Apollon77: Like that?

lauckhart (approved):

PR #2481, opened 2025-09-30
Add streaming sha256 for Node.js like Crypto

Apollon77:
This will be needed for OTA but here split out to lower the surface

### DIFF
--- a/packages/general/src/crypto/StandardCrypto.ts
+++ b/packages/general/src/crypto/StandardCrypto.ts
@@ -96,9 +96,12 @@ export class StandardCrypto extends Crypto {
         });
     }
 
-    computeSha256(buffer: Bytes | Bytes[]) {
+    computeSha256(buffer: Bytes | Bytes[] | ReadableStreamDefaultReader<Bytes> | AsyncIterator<Bytes>) {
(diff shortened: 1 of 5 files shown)


packages/general/src/crypto/StandardCrypto.ts
```diff
@@ -96,10 +96,13 @@ export class StandardCrypto extends Crypto {
         });
     }
 
-    computeSha256(buffer: Bytes | Bytes[]) {
+    computeSha256(buffer: Bytes | Bytes[] | ReadableStreamDefaultReader<Bytes> | AsyncIterator<Bytes>) {
         if (Array.isArray(buffer)) {
             buffer = Bytes.concat(...buffer);
         }
+        if (!Bytes.isBytes(buffer)) {
+            throw new ImplementationError("Streamed SHA-256 computation is not supported in StandardCrypto");
```
lauckhart: Maybe NotImplementedError?

lauckhart (approved):

packages/general/src/crypto/StandardCrypto.ts
Apollon77: good idea

PR #2487, opened 2025-10-02
Smaller Optimizations, TODOs and fixes

Apollon77:
This PR consists of smaller adjustments to new API, switch some places to use serialize rather than "json" and also a left over place where session parameters need to be provided


packages/node/src/node/client/ClientStructure.ts
```diff
@@ -162,6 +161,11 @@ export class ClientStructure {
         scope: ReadScope,
         currentUpdates: undefined | AttributeUpdates,
     ) {
+        // We only store values that are filtered by out fabric, else we create a mixture of data
```
lauckhart: This pretty much makes fabric filtering on the default subscription mandatory to use client node, is that the intent?

Could we maybe instead only apply it to fabric filtered attributes and/or maybe just make it skip them if they don't match the isFabricFilteredness of the default subscription?

packages/node/src/node/client/ClientStructure.ts
```diff
@@ -236,6 +240,7 @@ export class ClientStructure {
         const attrs = cluster.store.initialValues ?? {};
 
         // Generate a behavior if enough information is available
+        // TODO: Detect changes in revision/features/attributes/commands and update behavior if needed
```
lauckhart, line 239: I didn't put any receptivity to changes in fixed attributes in here, which includes all of these plus clusters, I think.  OK to leave the TODOs here but I was thinking we'd probably rebuild the client in the case of "program image change", which I guess would mean we detect a restart?

That aligns with how ServerNode works...  You can't change those properties unless you restart the node.

packages/node/src/node/client/ClientStructure.ts
```diff
@@ -307,6 +315,9 @@ export class ClientStructure {
 
                 const part = this.#endpointFor(partNo as EndpointNumber);
 
+                // TODO - remove endpoints/parts that are no longer present
+                //  Including events vis parts/endpoints on node (per endpoint and generic "changed")?
```
lauckhart, line 315: Eventing and cleanup are handled by endpoint.delete(), so just need to add that, which I thought existed somewhere but I can't find it if it does

packages/node/src/node/client/ClientStructure.ts
Apollon77: In fact I think we need to ensure to not get mixtures in the data. So that means if someone always reads we collect data from these reads, same for subscriptions. So main question is "How to prevent mixing it"
We could also remove that when we ensure that the storage always "just stores our fabric" (so return the full list but persist the "filtered" one? WDYT

packages/node/src/node/client/ClientStructure.ts
Apollon77: problem is that when the node gets a Software update during our runtime then also there things can change. This all might be easier to detect soon with the "Config version" attribute, but in fact these changes can happen anytime (and with considering bridges that you can configure also without any update. So it is not on restart, it could be on reconnect or even just by getting a subscription update

packages/node/src/node/client/ClientStructure.ts
Apollon77: Yeahh, I did not searched it. but it also is about parts aka clusters ... they could also be adjusted

packages/node/src/node/client/ClientStructure.ts
Apollon77: I could also add a TODO there toi adress this then later?

lauckhart: I think matching to NetworkClient.State#startupSubscription.isFiltered would do a good job of preventing mixes, but if there's no use case where you'd want to disable fabric filtering in the startup subscription then I think ignoring non-filtered reads is OK

So maybe either a TODO or modify the `Subscribe` in NetworkClient#startup to move `fabricFilter: true` below `...startupSubscription`, to force fabric filtering

packages/node/src/node/client/ClientStructure.ts
lauckhart: Yeah I meant restart of the peer, not ourselves, but I agree the most robust way to detect that is probably if a fixed attribute changes

packages/node/src/node/client/ClientStructure.ts
Apollon77: Ok checking with the default subscription makes sense and I added that for next push. But what do we do if there is no initial subscription?
* when mutate is called "for read"?
* when mutate is called "for subscribe"?

Store nothing? store only on subscribe and dev need to ensure that "fabric filters" match if using multiple subscribes manually? WDYT?

packages/node/src/node/client/ClientStructure.ts
Apollon77: Adjusted text. I would not go for fixed attr in general but wait for configurationversion and use that  then

packages/node/src/node/client/ClientStructure.ts
lauckhart: Hmm the main use case for updating on read is is to update state for unsubscribable attributes. 

Everything else where you have no default subscription seems like an edge case, so I'd say we could either not worry about isFabricFiltered or just choose one state to be the default, e.g. isFabricFiltered ≈≈ true like you originally had it. 

Or if you think sometimes disabling the startup subscription is important we could just make it not be nullable and have a separate "subscribe at startup" option.  Then there always be a subscription for you to check.  I just went the nullable route because it didn't seem like something people were likely to use so didn't want to add additional API noise.

Apollon77: moved to clientNode PR

PR #2490, opened 2025-10-02
Client node adjustments

Apollon77:
This PR is updating ClientNode with some convenient and practical enhancements.

* Adjust NetworkClient  from isDisabled to isEnabed including an option to define the startup behavior
* Add Peers.connect as main method to use to connect; enhance disconnect to also allow "unknown" peers
* Add "enableCluster" to dynamically add endpoints and behaviors for direct node usage
* add Peers.forAddress to add/get a ClientNode instance for a peer without commissioning
* adds filling known address into CommissioningClient state when a channel is estalished already (and a TODO touse that later)

The SubscriptionServer is not needing the node, it just needs to connect, so connect is used. OTA next will use "forAddresss"


packages/node/src/behavior/system/network/NetworkClient.ts
```diff
@@ -97,9 +111,18 @@ export namespace NetworkClient {
...
+         * This state can be changed at any time to enable or disable the node.
+         */
+        isEnabled = true;
+
+        /**
+         * If defined, it overrides the isEnabled state on startup.
+         * If not defined the node will start up in the state that it was last in.
          */
-        isDisabled = false;
+        enabledOnStartUp?: boolean;
```
lauckhart: What is the purpose of this guy?  Am wondering why node restart would affect the "enabledness" of a node

Perhaps relevant is that this isn't like PairedNode where there is a notion of connectedness independent of doing something.  For ClientNode we establish a connection when we need it, so if there's no default subscription the client will remain dormant until you try to use it.  So "isEnabled = false" is intended to signify "the user has physically disabled this node, don't try to talk to it"

packages/node/src/node/client/Peers.ts
```diff
@@ -112,6 +119,41 @@ export class Peers extends EndpointContainer<ClientNode> {
...
+            throw new ImplementationError("Cannot register a peer address for a fabric we do not belong to");
+        }
+
+        let node = this.get(peerAddress);
+        if (!node) {
+            // We do not have that node till now, also not persisted, so create it
+            const factory = this.owner.env.get(ClientNodeFactory);
+            node = factory.create({
+                ...options,
+                id: `${CLIENT_NODE_PREFIX}${NodeId.toHexString(peerAddress.nodeId)}`,
```
lauckhart: Is there a reason to use a different naming convention for nodes connected in this way?  And if so maybe should include the fabric ID too

packages/node/src/node/ClientNode.ts
```diff
@@ -188,6 +197,55 @@ export class ClientNode extends Node<ClientNode.RootEndpoint> {
...
+    async enableCluster(endpointId: number, type: ClusterBehavior.Type) {
+        let part = this.parts.get(endpointId);
+        const partExisting = part !== undefined;
+        if (part === undefined) {
+            part = new Endpoint({
+                id: `ep${endpointId}`,
+                number: endpointId,
+                type: EndpointType({
+                    name: "ClientEndpoint",
+                    deviceType: -1,
```
lauckhart: ProductDescriptionServer has an "UNKNOWN_DEVICE_TYPE", could use that

packages/node/src/storage/client/ClientNodeStores.ts
```diff
@@ -9,7 +9,11 @@ import type { ClientNode } from "#node/ClientNode.js";
 import type { Node } from "#node/Node.js";
 import { ClientNodeStore } from "./ClientNodeStore.js";
 
-const CLIENT_ID_PREFIX = "peer";
+/**
+ * Prefix for incrementally assigned client IDs that are usually commissioned, so an increasing ID is used instead the
```
lauckhart, line 13: FWIW we use a numeric ID so the ClientNode can be the same regardless of commissioning state, and be persisted even if uncommissioned

lauckhart:
> The SubscriptionServer is not needing the node, it just needs to connect, so connect is used. OTA next will use "forAddresss"

Is this maybe just a case of "defaultSubscription = undefined"?

packages/node/src/behavior/system/network/NetworkClient.ts
Apollon77: My problem with "isEnabled" is that it gets persisted and when e.g application crashes or such then it restores in his last state. But the developer maybe have other means of data to determine if a node shall be enabled or not. The Old API allowed you to control when a node should be "enabled" (aka subscription is established) and so it was controllable if all nodes get enabled on "startup of the controller" or later. I want to produce the same behavior. Or do you have any better idea?

packages/node/src/node/client/Peers.ts
Apollon77: Thats exactly the discussion I mean above :-)
Yes you are right fabric id is missing in that here if we keep.

In my eyes the default naming of client node "ids" (a dev provided string or increasing number) is a bit intransparent because if makes things very hard to see. In my eyes the case that a controller dev uses "speaking name ids" (like the new Controller example) is rare because normal usecases are more automated/scripted. So most nodes will have just a number id (aka "peerXXX") which is then again hard to see which real peer node it means.

In my ideal thinking we could have a numeric thing for anything not commissioned or such but then I think having the fabric+nodeid as identifier is clear in all places. But yes wanted to kickoff that discussion with this too :-)

packages/node/src/storage/client/ClientNodeStores.ts
Apollon77: Yes see also above. I understood that, but unsure if the (in my eyes) downsides would be worth that, but open to discuss.

packages/node/src/behavior/system/network/NetworkClient.ts
lauckhart: I was thinking that if you don't want the node to subscribe automatically you set startupSubscription to null, so "do not subscribe on startup"

packages/node/src/storage/client/ClientNodeStores.ts
lauckhart: Yeah I just meant r.e. comment

packages/node/src/node/client/Peers.ts
lauckhart: Makes sense, see what you think of #2493

Apollon77: Reverted to auto id

packages/node/src/behavior/system/network/NetworkClient.ts
Apollon77: For me a boolean flag is more convenient that always switching between "null" and "whatever subscription pattern the dev wants".

In fact the current reality is that if you do not enable the node you can not interact with it. maybe we could change that? Would allow node interactions a side (officially enable/disable aka main subscription"?

And as long as we need to enable the node I think controlling the behavior on startup is a good way.

lauckhart: I made "isDisabled" after our conversation r.e. devs needing a mechanism to specify "the device is offline, don't use it".  So it's kind of just metadata anyway, would probably be OK to allow comms with disabled nodes.  The main downside is that there should be nothing to prevent unintended access to a node.

I assume the "startupSubscription" should still be ignored if the node is marked disabled?

lauckhart: Separately, r.e. enabling the node on startup, I may be missing the use case.

Conceptually enableAtStartup sounds like "turn the device on when the controller restarts", but the controller doesn't actually know how to enable the node, just flip its local bit.

So what's the scenario where you mark the node as disabled but want it to change state when the controller restarts?

lauckhart:
> Did you mixed client and server now? The case her for subscription server is the re establishment of server side subscriptions. Yes when we somewhen do ServerInteraction then it might go there. But for now it goes into InteractionServer. 
> Or what did you mean?

I wasn't sure what you meant by "The SubscriptionServer is not needing the node, it just needs to connect, so connect is used."  Is this about a controller not allowing subscriptions then?

packages/node/src/behavior/system/network/NetworkClient.ts
Apollon77: Ok I guess these are two things.

1.) What means "enabled" for a node (same is the flag is called Disabled)
2.) What should happen when you enable a node? Currently when you enable it the default subscription (if not null) is established Which also means (all - depending on subscription) data are updated via a full read and kept uptodate.

Additionally we have 2-3 usecases to use client nodes
A) We are in Controller context and have commissioned a node. Usually wildcard subscription (aka our default) and enabled/disabled as needed by the dev/the user depending on how he wants it)
B) We have some cluster logic that requires us to "one time" interact with a different node on the network (usually just calling cluster commands, no subscription needed, like OTA which does calls once a day or such). Formally we use when we need, so no enable or disable here and on next start nothing should happen automatically.
C) We have established a binding because configured so and now the developer can implement what that means (which can be basically be just reading single values or invoking commands on need or could also be a subscription). Also here I would propose to not do anything and let the dev do whatever needed on "binding startup"

Maybe an idea to move subscription and "subscribe on startup" to an extra behavior and remove enabled/disabled?

Apollon77:
Alright @lauckhart, as discussed offline yesterday I adjusted the approach:
* isDisabled on Node networkClient can be used to disable a node, default is false. (aka enabled)
* The default subscription now also has this name and is always defined -  if not by the dev then it is a default full wildcard subscription. Optional, so undefined also means "default"
* autoSubscribe indicates if the node automatically subscribes on startup. Changing this leads to setup or teardown of the subscription
* special method "subscribe" on ClientNode allows to trigger a manual subscription after start if auto is disabled
* I added the logic for fabricFiltered Read checking from the other PR to here, so we can handle it together
* I moved the Initialization of ClientSubscriptions to be dynamic when peer nodes exist because in ControllerBehavior it was only initialized when this was added.

packages/node/src/behavior/system/network/NetworkClient.ts
```diff
@@ -14,29 +15,74 @@ import { NetworkBehavior } from "./NetworkBehavior.js";
...
+    override startup() {
+        return this.#handleSubscription();
+    }
+
+    /**
+     * Manually activate the subscription for the node, if not already active. This will fail if the node is disabled.
+     *
+     * If you want to disable subscription, set the isDisabled state to true or set autoSubscribe to false.
+     */
+    subscribe() {
```
lauckhart, line 53: I'm going to move this to logic to a dedicated subscription class but ok for now as this is original location

packages/node/src/node/ClientNode.ts
```diff
@@ -214,6 +219,80 @@ export class ClientNode extends Node<ClientNode.RootEndpoint> {
...
+
+        if (!partExisting) {
+            await this.add(part);
+        }
+    }
+
+    /**
+     * Use this to manually trigger the default subscription when automatic subscription is disabled.
+     */
+    async subscribe() {
```
lauckhart: In general we should try to keep as little functionality in the node classes directly as possible, relying on composition to organize methods by functional area.

In this case I think this should go in a SubscriptionClient.  That doesn't exist yet but maybe just delete the shortcuts here.

Also am thinking the "subscribe" mechanism should take an optional Subscribe (for managing non-default subscriptions) and return a handle the caller can use to manage the subscription.

packages/node/src/node/ClientNode.ts
```diff
@@ -214,6 +219,80 @@ export class ClientNode extends Node<ClientNode.RootEndpoint> {
...
+        return this.act(agent => agent.get(NetworkClient).subscribe());
+    }
+
+    get subscriptionsActivated() {
```
lauckhart: Also should not be on ClientNode directly

packages/node/src/node/ClientNode.ts
```diff
@@ -214,6 +219,80 @@ export class ClientNode extends Node<ClientNode.RootEndpoint> {
...
     }
+
+    /**
+     * For nodes where the behavior/cluster structure can not be initialized automatically (e.g. by a subscription) or
+     * when the subscription data misses special clusters, you can use this method to enable a cluster on a specific
+     * endpoint.
+     * The method adds the endpoint, if not existing, and also adds the cluster behavior for the "Complete" variant
+     * (means: all features, commands and attributes) of the cluster to the endpoint, if not already present.
+     */
+    async enableCluster(endpointId: number, type: ClusterBehavior.Type) {
```
lauckhart: In support of the principles of keeping logic small and composable, and ClientNode lean... I think this should split up a little:

* Behaviors collection should handle adding the behavior. Probably with the require() function, with additional hooks to properly wire in clients as necessary. Require is how you semantically say "this endpoint supports this behavior"

* Endpoints collection should handle creating an empty endpoint of it doesn't exist. I think that will work for both client and server

So the code pattern could look something like:

```ts
peer.endpoints.require(1).behaviors.require(OnOffClient)
```

If you want just add a TODO and we can do in separate commit

packages/protocol/src/action/Interactable.ts
```diff
@@ -33,6 +33,11 @@ export interface Interactable<SessionT = InteractionSession> {
      */
     subscribe(request: Subscribe, session?: SessionT): SubscribeResult;
 
+    /**
+     * Cancel an active subscription.
+     */
+    cancelSubscription(id: number): void;
```
lauckhart: ActiveSubscription.close() is the interface for canceling a subscription, can we remove this function and use that instead?

The general principle is working with objects rather than IDs, so `thing.something()` or `things.get(id).something()` rather than `thingManager.somethingToThing(id)`

packages/node/src/node/ClientNode.ts
Apollon77: Ok I remove for now. but we should have something like that for the default subscription in a convenient mannor

packages/node/src/node/ClientNode.ts
Apollon77: done

packages/protocol/src/action/Interactable.ts
Apollon77: changed

PR #2492, opened 2025-10-02
Networking updates

lauckhart:
* Move UDP abstractions into dedicated subdirectory

* Rename ServerAddressIp to ServerAddressUdp because that's what it is

* Clear ConnectionlessTransportSet when closed

* Add `AppAddress` utility class.  This is a URL-based layer 7 address with various useful helpers.  Originally I was going to extend `ServerAddress` but then realized it's probably better to keep layer 4- and 5+ addresses separate


packages/general/src/net/AppAddress.ts
```diff
@@ -0,0 +1,135 @@
...
+
+    get isTls() {
+        switch (this.appProtocol) {
+            case "https":
+            case "mqtts":
+            case "wss":
+                return true;
+
+            case "mqtt":
+                return this.protocolModifiers.includes("wss");
```
Apollon77, line 64: mqtt and wss? How they are connected? maybe add a comment to why

packages/general/src/net/ServerAddress.ts
```diff
@@ -40,16 +40,33 @@ export function ServerAddress(definition: ServerAddress.Definition) {
 
 export namespace ServerAddress {
     export function urlFor(address: ServerAddress): string {
-        return address.type === "udp" ? `udp://[${address.ip}]:${address.port}` : `ble://${address.peripheralAddress}`;
+        switch (address.type) {
+            case "udp":
+                return `${address.type}://${address.ip}:${address.port}`;
```
Apollon77, line 52: one line uses address.type, one hardcodedd "ble"?

packages/general/src/net/ServerAddress.ts
```diff
@@ -40,16 +40,33 @@ export function ServerAddress(definition: ServerAddress.Definition) {
...
     export function diagnosticFor(address: ServerAddress) {
         const diagnostic = Array<unknown>();
 
-        if (address.type === "udp") {
-            diagnostic.push("udp://", Diagnostic.strong(address.ip), ":", address.port);
-        } else {
-            diagnostic.push("ble://", Diagnostic.strong(address.peripheralAddress));
+        switch (address.type) {
+            case "udp":
+                diagnostic.push(`${address.type}://`, Diagnostic.strong(address.ip), ":", address.port);
```
Apollon77, line 68: same here

packages/general/src/net/AppAddress.ts
lauckhart: Added comment, MQTT over WebSockets

packages/general/src/net/ServerAddress.ts
lauckhart: This is because I had added TCP but then didn't use it so removed.  But will go ahead and added back

PR #2493, opened 2025-10-03
Peer logging improvements

lauckhart:
@Apollon77 see what you think of this.

It logs like this now:

image


Apollon77:
like it a lot! Thanks

PR #2495, opened 2025-10-03
Minor Bytes updates

lauckhart:
* Make Bytes.toString() take either Bytes or string so it works in contexts where either are allowed

* Add Bytes.fromString() that creates a UTF-8 buffer


packages/general/src/util/Bytes.ts
```diff
@@ -122,10 +122,22 @@ export namespace Bytes {
         return fromHex(result);
     }
 
-    export function fromString(string: string): Bytes {
+    export function fromString(string: string | Bytes): Bytes {
+        if (typeof string !== "string") {
```
Apollon77: it seems that this change now has side effects because if (for whatever reason) in the past e.g. a number was provided then it was handled correctly (whatever happend) ... but now that other type is passed through.

In fact the TlvCodec uses this also for strings ... but if for whatever reason a non-string is provided in the data it now is passed through.
I know normally should not be relevant but seems the Controller yaml testing "used" that.

So behavior till now was this ... so seems that TextEncoder does a toString if not a string is provided.

```
> new TextEncoder().encode(37)
Uint8Array(2) [ 51, 55 ]
```

Maybe adjust the "if" to state like "if already isByte then pass through

Apollon77 (approved):

PR #2498, opened 2025-10-03
New "gate" synchronization primitive

lauckhart:
Adds an object that can be awaited repeatedly and sometimes settled, sometimes not.  See esdoc for details.


Apollon77 (approved):
Interesting construction. cool

PR #2504, opened 2025-10-03
MDNS Fixes and optimizations

Apollon77:
This PR is:
* fixing  a missing Truncation flag set on queries if split into multiple messages including testing
* it optimizes the flow to always return the discovered entries, also when listening is disabled (and so all records cleared) afterwards
* it properly implements handing of truncated Queries (which was missing so far)
* it enhances the intercept method in the testing framework to allow manipulation of call parameters


packages/protocol/src/mdns/MdnsClient.ts
```diff
@@ -436,12 +446,20 @@ export class MdnsClient implements Scanner {
...
         if (waiter === undefined) return;
-        const { timer, resolver, resolveOnUpdatedRecords, commissionable } = waiter;
+        const { timer, resolver, responder, resolveOnUpdatedRecords, commissionable } = waiter;
         if (isUpdatedRecord && !resolveOnUpdatedRecords) return;
         logger.debug(`Finishing waiter for query ${queryId}, resolving: ${resolvePromise}`);
         timer?.stop();
         if (resolvePromise) {
-            resolver();
+            const response = responder?.();
+            if (MaybePromise.is(response)) {
```
lauckhart: Is there a code path where this returns a promise?  I tried to find it so I could make sure that something was taking the promise (which we don't do here) but it looks like the registered functions are always synchronous.

lauckhart (approved):

packages/protocol/src/mdns/MdnsClient.ts
Apollon77: The current usage is sync only ... I could also just do that but I thought still good to have for potential future things ... I can remove if you like it more "just sync for now"

lauckhart: Normal function would be nice just because taking a promise with nobody awaiting it makes me nervous, but nbd

PR #2508, opened 2025-10-04
Runtime tlv generation from model

lauckhart:
This dynamically performs TLV schema generation like we currently do statically via codegen.  This is a step toward addressing redundancy between ClusterType and ClusterModel.  The code is complete but not yet tested.

Depends on #2507 so that should merge first.


Apollon77 (approved):
Looks good theoretically,

We could update https://github.com/matter-js/matter.js/blob/main/packages/examples/src/device-onoff-advanced/cluster/MyFancyOwnFunctionality.ts to use it?

Additionally you could create a test which simply validates the new code with all generated types for e.g.cluster datatypes and such? (but followup PR fully ok)

PR #2525, opened 2025-10-06
Node.js HTTP endpoint implementation

lauckhart:
Node.js HTTP endpoint implementation

This includes an adapter from Node's old HTTP API to our WinterTC-style API.  This adapter is ugly due to Node's ancient API, but adapters for other platforms will be much more straightforward because they generally already use native Request and Response objects.

The adapter has WebSocket support, but since Node doesn't include a WebSocket server, this requires yet another plugin to enable WebSockets (called `WsAdapter` here).

Depends on #2524

### DIFF
--- a/packages/nodejs/src/environment/NodeJsEnvironment.ts
+++ b/packages/nodejs/src/environment/NodeJsEnvironment.ts
@@ -11,6 +11,7 @@ import {
     Boot,
     Crypto,
     Environment,
+    HttpEndpointFactory,
     ImplementationError,
     LogFormat,
     Logger,
@@ -19,6 +20,7 @@ import {
     StorageService,
     VariableService,
 } from "#general";
+import { NodeJsHttpEndpoint } from "#net/NodeJsHttpEndpoint.js";
 import { existsSync, readFileSync } from "node:fs";
 import { writeFile } from "node:fs/promises";
 import { resolve } from "node:path";
@@ -65,6 +67,8 @@ import { ProcessManager } from "./ProcessManager.js";
  * * `runtime.signals` - By default register SIGINT and SIGUSR2 (diag) handlers, set to false if not wanted
  * * `runtime.exitcode` - By default we set the process.exitcode to 0 (ok) or 1 (crash); set to false to disable
  * * `runtime.unhandlederrors` - By default we log unhandled errors to matter.js log; set to false to disable
+ *
+ * TODO - this should go away.  Node.js services should register with {@link ServiceBundle.default}
  */
 export function NodeJsEnvironment() {
     const env = new Environment("default");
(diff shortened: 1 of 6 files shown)


packages/nodejs/src/environment/NodeJsEnvironment.ts
```diff
@@ -144,9 +148,16 @@ function configureNetwork(env: Environment) {
 
     Boot.init(() => {
         if (config.installNetwork || (env.vars.boolean("nodejs.network") ?? true)) {
+            const basePathForUnixSockets = env.maybeGet(StorageService)?.location;
```
Apollon77: Is that valid approach? If storage is "something else" (like Redis or such) then this might all be misleading. Also "location" there is right now mainly used for logging and not that much more, right?
Ok to have a TODO or we need to find a way to define if we have a real"FS storage" used or something else.

packages/nodejs/src/net/NodeJsHttpEndpoint.ts
```diff
@@ -0,0 +1,358 @@
...
+                let settled = false;
+                server.once("listening", () => {
+                    if (settled) {
+                        return;
+                    }
+
+                    settled = true;
+                    resolve();
+                });
+                server.once("error", error => {
```
Apollon77: should we have another error handle for after that initial one because else nodejs might throw uncatcheable exception. Also in general having an error handler on the server (at least if we have created that) is good, also for WS
But fine with adding later

Apollon77 (approved):

packages/nodejs/src/environment/NodeJsEnvironment.ts
lauckhart: The fs stuff is admittedly a little hacky.  Currently the only things that set location are in nodejs and they set to a fs location independent of the Storage implementation. 

We need file storage for config files, sockets and should also really have a lockfile.  Plus ota firmware...  So really we should probably just have a proper fs abstraction.

I added the "resolve" method on storage too so this is just a fallback.  Let me see if I can think of something simple that will make it cleaner.

packages/nodejs/src/net/NodeJsHttpEndpoint.ts
lauckhart: Yeah I guess we could, I'm not sure when node would emit such an error event so crashing didn't seem entirely inappropriate.  Difficult to know how to deal with an error with zero context...  But maybe better to just log, will need to deal with aftermath regardless

Apollon77: I think logging for now is fine and then we need to see the real cases

packages/nodejs/src/environment/NodeJsEnvironment.ts
Apollon77: In fact location is "set manually" anyway ... so maybe have a "isFileSystemBacked" or such for now as a flag to set at the same place and false by default and check that ... and error out on the above case? maybe as intermediate solution

lauckhart: I changed to use root.path environment variable which is what it should have been using originally.

Apollon77: I think that makes a lot of sense!

packages/nodejs/src/net/NodeJsHttpEndpoint.ts
lauckhart: I added code to log a warning with the error message in these cases.  It sounds like Node may emit connection errors from failed connection attempts using this status code.  Which is different from "clientError" which I'm guessing is for errors from established connections.

PR #2526, opened 2025-10-06
Mqtt implementation

lauckhart:
MQTT implementation

This is a platform-independent MQTT implementation built on MQTT.js.

Depends on #2525

### DIFF
--- a/packages/mqtt/README.md
+++ b/packages/mqtt/README.md
@@ -1,6 +1,6 @@
 # MQTT implementation for matter.js
 
-This is a matter.js plugin that provides an MQTT implementation based on [MQTT.js](https://github.com/mqttjs/MQTT.js).
+This is a matter.js plugin that provides an MQTT client based on [MQTT.js](https://github.com/mqttjs/MQTT.js).
 
 To enable MQTT support you will need a broker that supports MQTT 5.  Import `@matter/mqtt` and enable MQTT by setting
 `mqttAddress` on the `RemoteServer` of your `ServerNode`.
--- a/packages/mqtt/README.md
+++ b/packages/mqtt/README.md
@@ -1,0 +1,9 @@
+# MQTT implementation for matter.js
+
+This is a matter.js plugin that provides an MQTT implementation based on [MQTT.js](https://github.com/mqttjs/MQTT.js).
+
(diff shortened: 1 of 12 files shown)


packages/mqtt/package.json
```diff
@@ -0,0 +1,75 @@
...
+        "@matter/tools": "*",
+        "safe-buffer": "^5.2.1"
+    },
+    "engines": {
+        "node": ">=20.19.0 <22.0.0 || >=22.13.0"
+    },
+    "files": [
+        "dist/**/*",
+        "src/**/*",
+        "require/**/*",
```
Apollon77: Does that exist?

packages/mqtt/package.json
```diff
@@ -0,0 +1,75 @@
...
+        "build": "matter-build",
+        "build-clean": "matter-build --clean"
+    },
+    "dependencies": {
+        "@matter/general": "*",
+        "mqtt": "^5.14.1"
+    },
+    "devDependencies": {
+        "@matter/tools": "*",
+        "safe-buffer": "^5.2.1"
```
Apollon77: Dev dep you use it in code. Should be normal dep or?

packages/mqtt/README.md
```diff
@@ -0,0 +1,9 @@
+# MQTT implementation for matter.js
+
+This is a matter.js plugin that provides an MQTT implementation based on [MQTT.js](https://github.com/mqttjs/MQTT.js).
```
Apollon77: Maybe add wording wise that this is an Mqtt client - yes implicitly clear because you set a remote server but maybe clarifies that.

packages/mqtt/package.json
lauckhart: no

packages/mqtt/package.json
lauckhart: Yeah I guess so, originally I was just importing by type, and MQTT.js had it as an actual dependency, but I guess now I'm calling Buffer.from so it should probably move

packages/mqtt/README.md
lauckhart: OK makes sense, if there were a pure-JS MQTT 5 broker I might've added that too just to make testing easy ;)

PR #2527, opened 2025-10-06
Streaming state component

lauckhart:
* Adds a new high-level `StateStream` component that provides a stream of changes for a node

* The stream provides direct access to the node/endpoint/behavior/attributes that changed

* It will emit changes both for ServerNode and any active ClientNodes

* A new `ChangeNotificationService` acts as the intermediary between `StateStream` and other components

* There are only two types of state events, "update" and "delete":

  - `update` provides a list of attributes and associated values for a single behavior

  - `delete` indicates that an endpoint (possibly a ClientNode) was deleted

* The stream supports limited filtering by node and/or cluster

* The stream also accepts initial data versions.  All data streams initially unless the data is at the specified version

* Also includes an update to `ServerNode` that will stop `MdnsService` if the node owns it (that is, it cannot be shared across nodes)

* Changes for Q attributes should adhere to the quiet broadcast schedule

* Change collection has a configurable coalescence interval to reduce the number of events emitted


packages/node/src/behavior/state/managed/Datasource.ts
```diff
@@ -443,7 +432,15 @@ function configureExternalChanges(internals: Internals) {
...
         }
 
-        const iterator = Object.keys(changes)[Symbol.iterator]();
+        const changedProps = Object.keys(changes);
+
+        const onChangePromise = internals.onChange?.(Array.from(changedProps));
+
+        const iterator = changedProps[Symbol.iterator]();
+
+        if (onChangePromise) {
```
Apollon77, line 441: should this also be


```suggestion
        if (MaybePromise.is(onChangePromise)) {
```

?

Apollon77 (approved):
Cool ... can we also get the attributeId into the changes and the cluster ids?
Most current use cases that report changes to other sources (like Openhab, iobroker and also the python server) use the ID level for that.

Apollon77:
@lauckhart Ahh right now I see ... yes ok it is change stream with all details so can easiely be filtered on need. alright

packages/node/src/behavior/state/managed/Datasource.ts
lauckhart: It's either a Promise or undefined so shouldn't need MaybePromise

PR #2539, opened 2025-10-08
Node.js WebSocket server implementation

lauckhart:
This is a node.js-specific WebSocket server implementation.  It may run on Node.js-compatible runtimes but only if they faithfully emulate the legacy Node.js HTTP server API.  Other runtimes tend to have a more modern WebSocket Server API which will require a dedicated adapter.

Unlike HTTP however this adapter will be slightly more complicated on Deno and Bun because their WebSocket implementations still look more evented than streamed.  They won't be as bad as the Node.js one, though, and won't require a third-party dependency.


.vscode/launch.json
```diff
@@ -102,7 +102,7 @@
                 "${workspaceFolder}/support/*/build/esm/**/*.js"
             ],
             "name": "Run shell",
-            "cwd": "${workspaceFolder}/packages/nodejs-shell",
+            "cwd": "${workspaceFolder}/support/codegen/packages/nodejs-shell",
```
Apollon77: is that right?

.vscode/launch.json
```diff
@@ -131,7 +131,7 @@
                 "${workspaceFolder}/support/*/build/esm/**/*.js"
             ],
             "name": "Run CLI tool",
-            "cwd": "${workspaceFolder}/packages/cli-tool",
+            "cwd": "${workspaceFolder}/support/codegen/packages/cli-tool",
```
Apollon77: same here

packages/nodejs-ws/package.json
```diff
@@ -0,0 +1,82 @@
...
+        "*": {
+            ".": [
+                "/dist/cjs/index.d.ts"
+            ]
+        }
+    },
+    "publishConfig": {
+        "access": "public"
+    },
+    "optionalDependencies": {
```
Apollon77: could you please move that up top the other deps to have them in one place? Thanks

.vscode/launch.json
lauckhart: No, codegen bug, fixed

packages/nodejs-ws/package.json
lauckhart: hah ok, that's where npm put it

PR #2541, opened 2025-10-08
Remote control servers for non-Matter protocols

lauckhart:
* The goal here is to support a variety of use cases:

    - Standard control channel via TCP or UNIX sockets to control devices and controllers from the local system

    - Offer a variety of higher-level interfaces for integration with 3rd party environments

    - Offer a standard channel for direct communication with mobile and web apps

    - High-level protocol appropriate for backchannel communication with test suites

* Adds a generic `ApiResource` object model that maps a generic API built on method + resource path command abstraction

* Implements a new `RemoteServer` behavior for managing pluggable protocols implementing a `RemoteInterface` protocol

* Adds three remote interfaces, HTTP (REST + RPC over HTTP), WebSockets and MQTT

* Goal of each API is to expose full Matter semantics while making APIs appear "native" to the external protocol

* A secondary goal is to align API semantics across disparate protocols and with matter.js API to:

    - Reduce the number of API "shapes" we need to document and devs need to learn

    - Reduce maintenance burden via shared implementation

* These are tested to varying degrees, current automated tests are only for WebSockets

### DIFF
--- a/packages/node/src/behavior/system/remote/api/Envelope.ts
+++ b/packages/node/src/behavior/system/remote/api/Envelope.ts
@@ -1,0 +1,110 @@
+/**
+ * @license
+ * Copyright 2022-2025 Matter.js Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { LocalActorContext } from "#behavior/context/index.js";
+import { RootSupervisor } from "#behavior/supervision/RootSupervisor.js";
+import { Bytes } from "#general";
+import { DataModelPath, Schema } from "#model";
+import { StatusResponse, TlvOfModel } from "#types";
+
+/**
+ * Api data envelope packages used for request and response.
+ *
+ * This allows for transparent conversion between our three "native" formats -- JS values, serialized JSON and
+ * serialized TLV.
+ */
+export class Envelope<T = unknown> {
+    #schema: Schema;
+    #js?: T;
+    #json?: string;
+    #tlv?: Bytes;
+
+    constructor({ schema, js, json, tlv }: Envelope.Definition<T>) {
+        this.#schema = schema;
+        this.#js = js;
+        this.#json = json;
+        this.#tlv = tlv;
+    }
+
+    /**
+     * Validate against the schema.
+     */
+    validate(path?: DataModelPath) {
+        if (!path) {
+            path = DataModelPath(this.#schema.path);
+        }
+        RootSupervisor.for(this.#schema).validate?.(this.js, LocalActorContext.ReadOnly, { path });
+    }
+
+    /**
+     * Convert a {@link js} value to {@link JSON}.
+     *
+     * This acts as a deep copy and optimizes JSON access.  If TLV becomes a priority then we can make this conversion
+     * configurable.
+     */
+    convertToJson() {
+        if (this.#js !== undefined) {
+            void this.json;
+        }
+        this.#js = undefined;
+    }
+
+    /**
+     * Native JS format.
+     */
+    get js() {
+        if (this.#js === undefined) {
+            if (this.#json) {
+                try {
+                    this.#js = JSON.parse(this.#json) as T;
+                } catch (e) {
+                    if (e instanceof SyntaxError) {
+                        throw new StatusResponse.FailureError(`Unparseable JSON: ${e.message}`);
+                    }
+                    throw e;
+                }
+            } else if (this.#tlv) {
+                this.#js = TlvOfModel(this.#schema).decode(this.#tlv) as T;
+            } else {
+                this.#js = null as T;
+            }
+        }
+        return this.#js;
+    }
+
+    /**
+     * JSON format.
+     */
+    get json() {
+        if (this.#json === undefined) {
+            this.#json = JSON.stringify(this.js);
+        }
+        return this.#json;
+    }
+
+    /**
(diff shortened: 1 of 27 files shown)


packages/node/src/behavior/system/remote/api/Envelope.ts
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    /**
+     * Native JS format.
+     */
+    get js() {
+        if (this.#js === undefined) {
+            if (this.#json) {
+                try {
+                    this.#js = JSON.parse(this.#json) as T;
```
Apollon77, line 62: How we handle bigints here? Any chance?

packages/node/src/behavior/system/remote/api/Envelope.ts
```diff
@@ -0,0 +1,110 @@
...
+        }
+        return this.#js;
+    }
+
+    /**
+     * JSON format.
+     */
+    get json() {
+        if (this.#json === undefined) {
+            this.#json = JSON.stringify(this.js);
```
Apollon77, line 83: same here

packages/node/src/behavior/system/remote/api/RemoteRequest.ts
```diff
@@ -0,0 +1,88 @@
+/**
+ * @license
+ * Copyright 2022-2025 Matter.js Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { isObject } from "#general";
+import { StateStream } from "#node/index.js";
```
Apollon77, line 8: maybe try to use a more exact import?

packages/node/src/behavior/system/remote/interfaces/MqttInterface.ts
```diff
@@ -0,0 +1,413 @@
...
+    NotImplementedError,
+} from "#general";
+import { any, CommandModel, Metatype, Schema } from "#model";
+import { StateStream } from "#node/integration/StateStream.js";
+import { StatusResponse } from "#types";
+import { Api } from "../api/Api.js";
+import { ApiPath } from "../api/ApiPath.js";
+import { Envelope } from "../api/Envelope.js";
+import { LocalResponse } from "../api/LocalResponse.js";
+import { RemoteRequest } from "../api/RemoteRequest.js";
```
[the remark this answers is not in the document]
Apollon77: Ha, more "duplicate"

packages/node/src/behavior/system/remote/interfaces/MqttInterface.ts
```diff
@@ -0,0 +1,413 @@
...
+    const supervisor = RootSupervisor.for(schema);
+    return supervisor.cast(js);
+}
+
+function jsToPayload(schema: Schema, js: unknown): Bytes | string {
+    if (js === undefined || js === null || js === "") {
+        return Bytes.empty;
+    }
+
+    switch (schema.effectiveMetatype) {
```
Apollon77, line 375: *int64 as string?

packages/node/src/node/ServerNode.ts
```diff
@@ -211,6 +212,7 @@ export namespace ServerNode {
         SessionsBehavior,
         EventsBehavior,
         ControllerBehavior,
+        RemoteServer,
```
Apollon77, line 215: Really automatically add by default and also have it enabled by default? Because right now we would (unexpectedly) expose a matter socket by default and have that enabled when I understand the code right.

I would think it is better to have it selectively added for now (also regarding memory and consumption such)

packages/node/src/behavior/system/remote/api/Envelope.ts
lauckhart: If they are serialized as strings and target an integer field then they should be cast to number or bigint depending on size.

packages/node/src/behavior/system/remote/api/Envelope.ts
lauckhart: Not handled here, I replaced JSON.stringify with a function that is bigint & bytes aware.

packages/node/src/behavior/system/remote/api/RemoteRequest.ts
lauckhart: yes

packages/node/src/behavior/system/remote/interfaces/MqttInterface.ts
lauckhart: New "asJson" function does that on serialization for bigints only

packages/node/src/node/ServerNode.ts
lauckhart: Switching to individual servers-per-protocol as discussed, will add a "plugin" behavior later to allow configuration-based behavior injection

packages/node/src/behavior/system/remote/interfaces/MqttInterface.ts
```diff
@@ -0,0 +1,413 @@
...
+                break;
+            }
+
+            js = payload;
+            break;
+
+        default:
+            if (Bytes.isBytes(payload)) {
+                try {
+                    payload = Bytes.toString(payload);
```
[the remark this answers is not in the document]
lauckhart: This was indeed a bug but code has changed a bit so no longer relevant

PR #2544, opened 2025-10-09
Only delete parent Instance when matching

Apollon77:
Old behavior was that the parent was always deleted irrelevant if the relevant

### DIFF
--- a/packages/nodejs-ws/test/WebSocketTest.ts
+++ b/packages/nodejs-ws/test/WebSocketTest.ts
@@ -25,6 +25,9 @@ let tempFileNum = 0;
 
 let environment: Environment;
 
+// TODO The timeouts of 5s are needed when the test runs locally with more network interfaces because it uses the
+//  real network. We should change that to use a mock network layer instead.
+
 describe("WebSocket", () => {
     beforeEach(async () => {
         environment = new Environment("test", Environment.default);
@@ -45,7 +48,7 @@ describe("WebSocket", () => {
             await cx.send("asdf" as any);
 
             await cx.receiveError("invalid-action", "Request is not an object");
-        });
+        }).timeout(5000);
 
         it("for missing opcode", async () => {
             await using cx = await setup();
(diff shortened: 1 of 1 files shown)


packages/general/src/environment/Environment.ts
```diff
@@ -113,37 +112,39 @@ export class Environment {
...
+                serviceEvents.deleted.emit(instance);
+            }
+        } else if (this.#parent !== undefined) {
+            this.#parent.delete(type, instance);
         }
     }
 
     /**
      * Prevent this environment from automatically instantiating or retrieving a service from parent environment.
+     * If a local instance exists it is removed first.
```
lauckhart: This comment is wrong now but I think in general we should remove block().  My thinking:

* If the device is there, you have already overridden the parent so block() doesn't make sense (plus now it throws an error)

* If you call delete, semantically I think it's reasonable to say "make this service inaccessible in this environment".  That means 1.) delete it if it's there, and 2.) install a null so it will not be retrieved from the parent environment

You could argue that you might want to do a "delete any local service but do use inherited service" but that's not generally how environment is designed...  I think I'd just address that one if we ever need it.

packages/general/src/environment/Environment.ts
```diff
@@ -113,37 +112,39 @@ export class Environment {
...
-        if (serviceEvents) {
-            serviceEvents.deleted.emit(instance);
+            this.#deleted.emit(type, instance);
+
+            const serviceEvents = this.#serviceEvents.get(type);
+            if (serviceEvents) {
+                serviceEvents.deleted.emit(instance);
+            }
+        } else if (this.#parent !== undefined) {
+            this.#parent.delete(type, instance);
```
lauckhart: I think it would be better semantically if we treat a mutation as local to this environment, so don't cascade delete up.  Otherwise conceptually you might be mucking with an environment you don't own.

See also my comment on `block()`.

PR #2546, opened 2025-10-09
Non-Matter protocol servers (take 2)

lauckhart:
* The goal here is to support a variety of use cases:

    - Standard control channel via TCP or UNIX sockets to control devices and controllers from the local system

    - Offer a variety of higher-level interfaces for integration with 3rd party environments

    - Offer a standard channel for direct communication with mobile and web apps

    - High-level protocol appropriate for backchannel communication with test suites

* Adds a generic `ApiResource` object model that maps a generic API built on method + resource path command abstraction

* Adds three new servers, HttpServer (REST + RPC over HTTP), WebSocketServer and MqttServer

* Goal of each API is to expose full Matter semantics while making APIs appear "native" to the external protocol

* A secondary goal is to align API semantics across disparate protocols and with matter.js API to:

    - Reduce the number of API "shapes" we need to document and devs need to learn

    - Reduce maintenance burden via shared implementation

* These are tested to varying degrees, current automated tests are only for WebSockets

* Also adds packages/nodejs-ws to build


packages/general/src/util/String.ts
```diff
@@ -161,7 +161,7 @@ export function serialize(value: any) {
                 return "[]";
             }
 
-            const entries = Object.entries(value)
+            const entries = Object.entries(value as {})
```
Apollon77, line 164: maybe use a Record<string, unknown> here because{} is special typing wise

packages/node/src/behavior/system/remote/api/Envelope.ts
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    /**
+     * Native JS format.
+     */
+    get js() {
+        if (this.#js === undefined) {
+            if (this.#json) {
+                try {
+                    this.#js = JSON.parse(this.#json) as T;
```
Apollon77, line 64: Adding note here but applies to all JSON parse and JSON.stringify ... how we have BigInts here? They will be there!

packages/general/src/util/String.ts
lauckhart: In this case I do mean "any non-nullish value" which is just to satisfy `Object.vales` typing. I don't normally cast to something more detailed than required for use

packages/node/src/behavior/system/remote/api/Envelope.ts
lauckhart: They won't be produced here of course, but ValueSupervisor.cast covers all "primitive" types that don't have a native JSON presentation, including bigints, dates and bytes

For integer specifically, if a string appears in a position where schema specifies an integer, the cast will convert to `number` if the string represents an integer that can be encoded safely as a double and otherwise a bigint

PR #2549, opened 2025-10-10
Remote server examples

lauckhart:
Two new examples:

* Creating a node with a WebSocket control channel over UNIX socket

* Creating a node that publishes itself via MQTT

### DIFF
--- a/packages/general/src/environment/Environment.ts
+++ b/packages/general/src/environment/Environment.ts
@@ -77,21 +77,20 @@ export class Environment {
 
         // When null then we do not have it and also do not want to inherit from parent
         if (mine === undefined) {
-            let instance = this.#parent?.maybeGet(type);
+            const instance = this.#parent?.maybeGet(type);
             if (instance !== undefined && instance !== null) {
                 // Parent has it, use it
                 return instance;
             }
+        }
 
-            // ... otherwise try to create it
-            if ((type as Environmental.Factory<T>)[Environmental.create]) {
-                instance = (type as any)[Environmental.create](this) as T;
-                if (!(instance instanceof type)) {
-                    throw new InternalError(`Service creation did not produce instance of ${type.name}`);
-                }
-                this.set(type, instance);
-                return instance;
+        // ... otherwise try to create it. The create method must install it in the environment if needed
+        if ((type as Environmental.Factory<T>)[Environmental.create]) {
+            const instance = (type as any)[Environmental.create](this) as T;
+            if (!(instance instanceof type)) {
+                throw new InternalError(`Service creation did not produce instance of ${type.name}`);
             }
+            return instance;
         }
 
         throw new UnsupportedDependencyError(`Required dependency ${type.name}`, "is not available");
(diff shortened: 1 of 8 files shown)


packages/examples/src/mqtt-device/README.md
```diff
@@ -0,0 +1,5 @@
+# Example device with MQTT control enabled
+
+This example demonstrates how to implement a matter.js device that you can control via MQTT.
```
Apollon77, line 3: maybe state that it acts as a client and connects to a provided mqtt server? :-)

packages/examples/src/mqtt-device/MqttDevice.ts
```diff
@@ -0,0 +1,42 @@
...
+// This demonstrates a matter.js device that supports an MQTT API on Node.js
+
+import { MqttServer, ServerNode } from "@matter/main";
+import { OnOffLightDevice, OnOffLightRequirements } from "@matter/main/devices/on-off-light";
+
+// This plugin installs MQTT support for matter.js
+import "@matter/mqtt";
+
+/**
+ * Create node with MQTT APIs enabled.
```
Apollon77, line 17: maybe also here add an info on mqtt client role please

packages/examples/src/web-socket-device/README.md
```diff
@@ -0,0 +1,5 @@
+# Example device with WebSocket control enabled
+
+This example demonstrates how to implement a matter.js device that you can control via WebSockets.
```
Apollon77, line 3: Please also here state that it acts as a server. Usually would be cool to have the parameter to set the "bind address" also listed here in readme

packages/examples/package.json
```diff
@@ -61,6 +61,11 @@
         "@project-chip/matter.js": "*",
         "esbuild": "^0.25.9"
     },
+    "devDependencies": {
+        "@matter/nodejs-ble": "*",
```
Apollon77, line 63: ble is now duplicated in deps ... because already an optional one

packages/examples/package.json
```diff
@@ -61,6 +61,11 @@
...
+        "@matter/mqtt": "*",
```
Apollon77, line 64: why dev deps?

packages/examples/package.json
lauckhart: OK I moved to a required dep because it is imported unconditionally by DeviceNodeFull.ts and ControllerNode.ts.

packages/examples/package.json
lauckhart: Just because they aren't required by all examples, but I moved to normal dep

packages/examples/src/web-socket-device/README.md
lauckhart: Bind-to address is getting kind of redundant but I agree makes sense here.  "you can control via Websockets" makes server clear to me but the mention of listening address should also help clarify

packages/examples/src/mqtt-device/MqttDevice.ts
lauckhart: Would folks expect us to be embedding a broker?  I added language mentioning that an MQTT 5 broker is required but I think "and we are a client of that broker" is unnecessary.

packages/examples/src/mqtt-device/README.md
lauckhart: OK, replaced the "see implementation for details" with a copy of the details.

packages/examples/src/mqtt-device/MqttDevice.ts
Apollon77: No clue what they would expect :-) thats why client is fully ok for now and we should have clear info ... rest we see when people use it after release

PR #2566, opened 2025-10-13
Close OccurrenceManager on node shutdown

Apollon77:
To also ensure non-volatile stores gets closed properly. The PR also fixes that the promise of close is returned correctly.

@lauckhart The OccurrenceManager is initially loaded by "events" but removing it with the other node logic is too early to send outstanding events on shutdown. So I moved it there. Makes that sense?


packages/protocol/src/events/OccurrenceManager.ts
```diff
@@ -272,7 +272,7 @@ export class OccurrenceManager {
     }
 
     close(): MaybePromise<void> {
-        MaybePromise.then(this.#cull, () => this.#store.close());
```
lauckhart, line 275: Maybe add a note here that ServerEnvoronment takes care of this

PR #2575, opened 2025-10-15
Interaction rework

Apollon77:
This PR is reworking several aspects of the Client interactions and also reworks and sync logging in Server and client interactions. It touches:
* Fixing logformat to prevent "double spaces" on empty elements
* Format function returns in logging as Diagnostics and not values
* Rework Invoke to move encoding to when Request is build and not per invoke request (mainly for logging convenience), also adds meta data to transport object for decoding
* Adding Invoke options for a concrete and Wildcard invoke definition
* Adds needed "low level" invoke options to tweak the request
* Adds path logging based on Specifiers, use for Read, Write and Invoke
* Adjusts Write logic with "matter 1.4 rules" regarding chunking
* Adjusts ClientInteraction to have all checks and needed logic as InteractionClient had too (beside Group functionality, comes as separate PR!)
* Allow more Invoke options for ClientNodeInteraction
* Streamline logging in interactions and exchanges and case/pase
* Add a missing await

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -63,6 +63,23 @@ The main work (all changes without a GitHub username in brackets in the below li
 -   Other
     - Feature: For developers working with the matter.js repository on Macs and Windows, we now offer a devcontainer to facilitate native Linux development
 
+# 0.15.6 (2025-10-16)
+
+-   @matter/general
+    - Fix: Ensures services are not removed from parents when they are removed locally
+
+-   @matter/node
+    - Fix: Ensures OccurrenceManager is closed on node shutdown
+
+-   @matter/protocol
+    - Enhancement: Removes exchanges earlier when closing to prevent double session closures
+    - Enhancement: Ignores invoke responses that should be suppressed but delivered
+    - Fix: Fixes Fabric scoped event reads to ensure the filtering is applied correctly
+
+-   @project-chip/matter.js
+    - Fix: Initializes controller into an own environment instead using the provided environment
+    - Fix: Ensure that the PairedNode meta-data are updated and stored when changed
+
 ## 0.15.5 (2025-10-07)
 
 -   @matter/general


packages/node/src/node/server/InteractionServer.ts
```diff
@@ -442,28 +454,28 @@ export class InteractionServer implements ProtocolHandler, InteractionRecipient
             throw new StatusResponseError("No attributes or events requested", StatusCode.InvalidAction);
         }
 
-        logger.debug(
-            () =>
-                `Subscribe to attributes: ${
-                    attributeRequests?.map(path => this.#node.protocol.inspectPath(path)).join(", ") ?? "none"
-                }, events: ${eventRequests?.map(path => this.#node.protocol.inspectPath(path)).join(", ") ?? "none"}`,
+        logger.debug(() =>
```
lauckhart: Maybe either `via, "Subscription details"` as log message prefix, or conditionally merge with the message above?

packages/node/src/node/server/InteractionServer.ts
```diff
@@ -507,7 +519,7 @@ export class InteractionServer implements ProtocolHandler, InteractionRecipient
                 error instanceof MatterError ? error.message : error,
             );
             if (error instanceof StatusResponseError && !(error instanceof ReceivedStatusResponseError)) {
-                logger.info(`Sending status response ${error.code} for interaction error: ${error.message}`);
+                logger.info(`Status response » ${error.code} for interaction error: ${error.message}`);
```
lauckhart: Might replace "for interaction error: " with just "Error:".

For bare error messages I created #2578 so you can do `Diagnostic.errorMessage(error)` and it will print the error the same way as `Diagnostic.error` would except without the stack trace.

So this line could be something like:

```ts
logger.info(`Status »", Diagnostic.bold(error.code), exchange.via, "Error:", Diagnostic.errorMessage(error))
```

packages/node/src/node/server/InteractionServer.ts
```diff
@@ -610,10 +627,18 @@ export class InteractionServer implements ProtocolHandler, InteractionRecipient
     ) {
         const exchange = this.#context.exchangeManager.initiateExchange(session.peerAddress, INTERACTION_PROTOCOL_ID);
         const message = {} as Message;
-        logger.debug(
-            `Send DataReports to re-establish subscription ${subscriptionId} to `,
-            Diagnostic.dict({ isFabricFiltered, maxInterval, sendInterval }),
+
+        logger.info(
+            `Re-Establish subscription »`,
```
lauckhart: Here and below should write "re-establish" as "reestablish"

packages/protocol/src/action/request/Specifier.ts
```diff
@@ -97,6 +97,26 @@ export namespace Specifier {
           ? CMD
           : never;
 
+    export function commandOf<const C extends ClusterType, const CMD extends Specifier.Command<C>>(
```
lauckhart: Should rename other functions to have `Of` instead of `For`, or rename this to `commandFor`...  I think I like "Of" better but they should match

packages/protocol/src/action/request/Specifier.ts
```diff
@@ -144,3 +164,62 @@ export namespace Specifier {
...
+export function toWildcardOrHexPath(name: string, value: number | bigint | undefined) {
+    if (value === undefined) {
+        return "*";
+    }
+    return `${name ? `${name}:` : ""}0x${value.toString(16)}`;
+}
+
+/**
+ * Resolve a path into a human readable textual form for logging
+ * TODO: Add a Diagnostic display formatter for this
```
lauckhart, line 177: I added diagnostic display to `DataModelPath` to handle formatting of paths in general, and `ExpandedPath` to handle conversion of wire paths to `DataModelPath`.

You might add the "!" and "[ADD]" flags there or handle them separately, but otherwise would be good if we standardized on `ExpandedPath`.

packages/protocol/src/action/request/Write.ts
```diff
@@ -41,13 +52,25 @@ export function Write(optionsOrData: Write.Options | Write.Attribute, ...data: W
...
-        interactionModelRevision: options.interactionModelRevision ?? FALLBACK_INTERACTIONMODEL_REVISION,
-    };
+        moreChunkedMessages: false,
+        interactionModelRevision: options.interactionModelRevision ?? Specification.INTERACTION_MODEL_REVISION,
+
+        [Diagnostic.value]: () =>
+            Diagnostic.list(
+                data.map(entry => {
+                    const { version, value } = entry;
+                    return `${resolvePathForSpecifier(entry)} = ${Diagnostic.json(
```
lauckhart, line 68: So r.e. comment on `resolvePathForSpecifier`, use `ExpandedPath` instead in places like this

lauckhart (approved):
Nice, that covers a lot of details

packages/node/src/node/server/InteractionServer.ts
Apollon77: adjusted

packages/protocol/src/action/request/Specifier.ts
Apollon77: Added https://github.com/matter-js/matter.js/issues/2584 for it ... I would then also adjust the ProtocolService logging to that

PR #2596, opened 2025-10-18
JavaScript decorators for Matter semantics

lauckhart:
This was something of a heavy lift but I think the final result is pretty solid.  The decoration syntax is minimal and fairly straightforward.  The implementation has some complexity but I did my best to minimize.

Still TODO: Release notes and support for custom clusters (test and integrate TLV generation and add ClusterType generation).

* Includes decorator types, utilities and polyfills Symbol.metadata

* New `ClassSemantics` and `FieldSemantics` manage Matter semantics associated with JavaScript classes; these classes support iterative `Model` assembly either declaratively via decoration or programmatically via the API used by the decorators

* Adds various decorators for declaratively specifying type, qualities, etc. for classes and fields

* Change terminology for immutable models from "frozen" to "final"

* Add `Schema` function to retrieve the data model for a JavaScript class

* Includes decoration tests

* Replaces schema management in `Behavior` with the new metadata infrastructure

* Replaced manual schema for `CommissioningClient` with decorators; chosen for proof-of-concept because it's high-complexity, low-risk, and doesn't require TLV generation which is still untested

* Several other small tweaks are cosmetic or address fragility turned up by underlying changes to `Behavior`

### DIFF
--- a/packages/general/src/util/Promises.ts
+++ b/packages/general/src/util/Promises.ts
@@ -252,7 +252,7 @@ export const MaybePromise = {
                     //
                     //      https://github.com/typescript-eslint/typescript-eslint/issues/7276
                     //
-                    // eslint-disable-next-line @typescript-eslint/no-misused-promises
+                    // seems to be ok as of eslint 9.37.0
                     result = (result as Promise<T>).finally(onfinally);
                 } else {
                     result = result.then(
--- a/packages/node/src/behavior/system/network/NetworkClient.ts
+++ b/packages/node/src/behavior/system/network/NetworkClient.ts
@@ -20,8 +20,14 @@ export class NetworkClient extends NetworkBehavior {
     declare events: NetworkClient.Events;
 
     override initialize() {
-        this.reactTo(this.events.autoSubscribe$Changed, this.#handleSubscription);
-        this.reactTo(this.events.defaultSubscription$Changed, this.#handleChangedDefaultSubscription);
+        if (this.#node.isGroup) {
+            // Groups can never subscribe
+            this.state.autoSubscribe = false;
+            this.state.defaultSubscription = undefined;
+        } else {
+            this.reactTo(this.events.autoSubscribe$Changed, this.#handleSubscription, { offline: true });
+            this.reactTo(this.events.defaultSubscription$Changed, this.#handleChangedDefaultSubscription);
+        }
     }
 
     override async startup() {
--- a/packages/node/src/node/ClientGroup.ts
+++ b/packages/node/src/node/ClientGroup.ts
@@ -1,0 +1,36 @@
+/**
+ * @license
+ * Copyright 2022-2025 Matter.js Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+import type { ActionContext } from "#behavior/context/ActionContext.js";
+import { Interactable } from "#protocol";
+import { ServerNodeStore } from "#storage/index.js";
+import { ClientNode } from "./ClientNode.js";
+import { ClientGroupInteraction } from "./client/ClientGroupInteraction.js";
+
+export class ClientGroup extends ClientNode {
+    #interaction?: ClientGroupInteraction;
+
+    override get isGroup() {
+        return true;
+    }
+
+    override get interaction(): Interactable<ActionContext> {
+        if (this.#interaction === undefined) {
+            this.#interaction = new ClientGroupInteraction(this);
+        }
+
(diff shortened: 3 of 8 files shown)


packages/node/src/behavior/supervision/RootSupervisor.ts
```diff
@@ -160,8 +161,9 @@ export class RootSupervisor implements ValueSupervisor {
             // TODO: We should handle writable/fabric scoped being non-volatile already in the conformance interpreter
             if (
                 member.effectiveQuality.nonvolatile ||
```
Apollon77, line 163: hm ... non vloatile is also attribute-only, right?

packages/node/src/behavior/system/commissioning/CommissioningClient.ts
```diff
@@ -258,174 +274,198 @@ export class CommissioningClient extends Behavior {
...
-                FieldElement({ name: "activeInterval", type: "duration", constraint: "max 3600000" }),
-                FieldElement({ name: "activeThreshold", type: "duration", constraint: "max 65535" }),
-            ],
-        }),
-        FieldElement({ name: "tcpSupport", type: "uint8", quality: "N" }),
-        FieldElement({ name: "longIdleOperatingMode", type: "bool", quality: "N" }),
-    );
-}
+    export class PeerAddress implements ProtocolPeerAddress {
+        @field(fabricId)
```
Apollon77: fabric-idx, or? uups was wrong before too

packages/node/src/behavior/system/commissioning/CommissioningClient.ts
```diff
@@ -258,174 +274,198 @@ export class CommissioningClient extends Behavior {
...
+    }
+
+    /**
+     * The network address of a node.
+     */
+    export class NetworkAddress implements ServerAddress.Definition {
+        @field(string)
+        type: "udp" | "tcp" | "ble";
+
+        @field(string, nullable)
```
Apollon77: nullable or optional? ... and was required and not nullable before. also for some more

packages/node/src/behavior/system/commissioning/CommissioningClient.ts
```diff
@@ -258,174 +274,198 @@ export class CommissioningClient extends Behavior {
...
+         * The remote node's session intervals.
          */
-        sessionParameters?: Partial<SessionParameters>;
+        @field(SessionIntervals, nullable, nonvolatile)
+        sessionIntervals?: Partial<ProtocolSessionIntervals>;
 
         /**
          * TCP support bitmap.
          */
+        @field(uint8, nullable, nonvolatile)
```
Apollon77: can we use an official type for the bitmap here? Should be there in elements or?

packages/node/src/behavior/system/commissioning/CommissioningClient.ts
lauckhart: right

packages/node/src/behavior/supervision/RootSupervisor.ts
lauckhart: For matter yes, but for we also allow on fields to make them persistent...  Right now all state fields are persistent I think but that was more a mistake than anything...  I guess we could leave it that way but it would potentially mean extra writes if we have frequently changing state values

packages/node/src/behavior/system/commissioning/CommissioningClient.ts
lauckhart: Previously the model didn't match the type, I generally tweaked to match the type when that was the case

lauckhart: r.e. nullable vs optional, "nullable" seemed to be a better match for JS semantics but I'll switch to "mandatory" instead since it seems like "allow undefined" is a better default and otherwise I'd need to make all of the decorators set mandatory by default

packages/node/src/behavior/system/commissioning/CommissioningClient.ts
lauckhart: We don't have the bitmap defined anywhere afaik.  Definitely not generated, it's just a table in a random section of the spec

We could add it but then would need to add encode/decode support to `RemoteDescriptor` and ideally decorators for bitfields.

lauckhart: Anyway I think better for separate PR, this preserves semantics as they were previously

PR #2598, opened 2025-10-19
Legacy controller Rework to Use ServerNode Part #1

Apollon77:
This is a first part of rework to base the legacy controller on ServerNode. It uses a server node for all network parts (and so also allows to add OTA clusters in a second step). It still does not use "ClientNode" but still "PairedNode" and just grabs the networking from the node. But it's a first step.

It also:
* Enhances FabricAutority and ControllerBehavior to allow the same Controller configuration featureset as it was available in Legacy controller (e.g. custom defined Root Node Id and such)
* Initializes the PeerAddress Storage in serverEnvironment to avoid double initialization (was initialized in twi places before and also hard overwritten with a new class).
* it adds some optimizations around MDNS and advertisements vs node shutdowns
* Ensures that ControllerBehavior is loaded also for MatterController legacy usage
* Removed FabricAuthorityConfigurationProvider service and replace by a config which is loaded on need
* Also optimizes some imports to reduce some cycling imports.
* When exchange got closed stop resubmissions
* * includes a package-lock relock

I includes two "hacks" to allow the legacy controller still use their own (legacy) storages. This is added with TODOs to remove this again. It formally bears the risk that such a class is outside (parent) of the node-own environment. This should be covered by logic but still a risk

### DIFF
--- a/packages/protocol/src/mdns/MdnsClient.ts
+++ b/packages/protocol/src/mdns/MdnsClient.ts
@@ -221,7 +221,7 @@ export class MdnsClient implements Scanner {
         this.#updateListeningStatus();
     }
 
-    /** Update the status of we care about MDNS messages or not */
+    /** Update the status if we care about MDNS messages or not */
     #updateListeningStatus() {
         const formerListenStatus = this.#listening;
         // Are we interested in MDNS traffic or not?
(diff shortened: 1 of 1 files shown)


packages/node/src/behavior/system/controller/ControllerBehavior.ts
```diff
@@ -61,27 +74,29 @@ export class ControllerBehavior extends Behavior {
...
-                        return adminFabricLabel;
-                    }
-
-                    get fabricId() {
-                        return adminFabricId;
-                    }
-                })(),
+                    },
+                    adminFabricLabel,
+                    fabricId: adminFabricId,
```
lauckhart: IIRC these were getters so you could set the values on state manually at any time prior to the controller initializing, is they any reason not to support that?

packages/protocol/src/fabric/FabricAuthority.ts
```diff
@@ -25,20 +25,52 @@ const logger = Logger.get("FabricAuthority");
  * Configuration for fabrics controlled by a FabricAuthority.
  */
 interface FabricAuthorityConfiguration {
+    adminFabricLabel: string;
     adminVendorId?: VendorId;
+    nodeId?: NodeId;
```
lauckhart: Maybe admin prefix here too?

packages/protocol/src/fabric/FabricAuthority.ts
```diff
@@ -25,20 +25,52 @@ const logger = Logger.get("FabricAuthority");
...
+
+    get adminVendorId(): VendorId | undefined {
+        return this.#config.adminVendorId;
+    }
+
+    get nodeId(): NodeId | undefined {
+        return this.#config.nodeId;
+    }
+
+    get fabricIndex(): FabricIndex | undefined {
```
lauckhart: Should this be deprecated too?

lauckhart (approved):

support/chip-testing/src/handler/LegacyControllerCommandHandler.ts
```diff
@@ -136,6 +136,7 @@ export class LegacyControllerCommandHandler extends CommandHandler {
     }
 
     close() {
+        logger.info(`-----> Controller ${this.#identity} stopping`);
```
lauckhart: Is this left here accidentally?

lauckhart (approved):

PR #2599, opened 2025-10-19
Initial Abstraction for Client Group Nodes

Apollon77:
This PR adds a first abstraction for Client "Group" Nodes. The Client Node factory automatically initializes them when the provided address is a group nodeId. grou pnodes can not have a state because read/subscribe is not allowed, so we simpyl use a in mem storage for them.

Additionally the adjusted ClientGroupInteraction is adding all relevant checks to all interactions, like also implemented in InteractionClient.

### DIFF
--- a/packages/node/src/behaviors/access-control/AccessControlServer.ts
+++ b/packages/node/src/behaviors/access-control/AccessControlServer.ts
@@ -77,21 +77,14 @@ export class AccessControlServer extends AccessControlBehavior.with("Extension")
                 // Handle Backward compatibility to Matter.js before 0.9.1 and add the missing ACL entry if no entry was set
                 // so far by the controller
                 const fallbackAcl: AccessControlTypes.AccessControlEntry = {
-                    fabricIndex: fabric.fabricIndex,
                     privilege: AccessControlTypes.AccessControlEntryPrivilege.Administer,
                     authMode: AccessControlTypes.AccessControlEntryAuthMode.Case,
                     subjects: [fabric.rootNodeId],
                     targets: null, // entire node
+                    fabricIndex: fabric.fabricIndex,
                 };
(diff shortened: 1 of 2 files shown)


packages/node/src/behavior/system/network/NetworkClient.ts
```diff
@@ -20,8 +21,14 @@ export class NetworkClient extends NetworkBehavior {
     declare events: NetworkClient.Events;
 
     override initialize() {
-        this.reactTo(this.events.autoSubscribe$Changed, this.#handleSubscription);
-        this.reactTo(this.events.defaultSubscription$Changed, this.#handleChangedDefaultSubscription);
+        if ((this.#node as ClientGroup).isGroup) {
```
lauckhart: Maybe add ClientGroup.is()?

packages/node/src/node/client/ClientGroupInteraction.ts
```diff
@@ -0,0 +1,68 @@
...
+    SubscribeResult,
+    Write,
+    WriteResult,
+} from "#protocol";
+import { ClientNodeInteraction } from "./ClientNodeInteraction.js";
+
+export class ClientGroupInteraction extends ClientNodeInteraction {
+    /** Groups do not support reading or subscribing to attributes */
+    override read(_request: Read, _context?: ActionContext): ReadResult {
+        throw new ImplementationError("Groups do not support reading attributes");
```
lauckhart: I think a custom `InvalidGroupOperationError extends ImplementationError` or something is warranted

packages/node/src/behavior/system/network/NetworkClient.ts
Apollon77: Had the same idea while doing the dishes ;-))

PR #2619, opened 2025-10-23
Initial push for a Thermostat server default Implementation

Apollon77:
This PR implements Thermostat cluster server including all relevant features. MSCH (Mater Schedule Configuration is considered provisional and so not implemented for now).

The implementation contains  docs and also a bit of convenience extra logic when it comes to setting the runmode.

It includes several adjustments and fixes to correctly handle "Dynamic properties" from behaviors including managed types also in write cases.
It implements Atomic Write in a generic WriteHandler class and uses it for Presets as of now. The implementation of Atomic Write is partially bound to Thermostats for now because also the needed command and types is enhanced there because only used in this regard as of now. We generalize once more clusters use this.

### DIFF
--- a/packages/node/src/behavior/state/managed/ManagedReference.ts
+++ b/packages/node/src/behavior/state/managed/ManagedReference.ts
@@ -54,18 +54,18 @@ export function ManagedReference(
     const altKey = primaryKey === "id" ? (key === name ? undefined : name) : id;
     let value: unknown;
 
-    let properties: Val.Struct | undefined;
+    let dynamicContainer: Val.Struct | undefined;
     if ((parent.value as Val.Dynamic)[Val.properties]) {
-        properties = (parent.value as Val.Dynamic)[Val.properties](parent.rootOwner, session);
-        if (key in (properties as Container)) {
-            value = (properties as Container)[key];
-        } else if (altKey !== undefined && altKey in (properties as Container)) {
-            value = (properties as Container)[altKey];
+        dynamicContainer = (parent.value as Val.Dynamic)[Val.properties](parent.rootOwner, session);
+        if (key in (dynamicContainer as Container)) {
+            value = (dynamicContainer as Container)[key];
+        } else if (altKey !== undefined && altKey in (dynamicContainer as Container)) {
+            value = (dynamicContainer as Container)[altKey];
         } else {
-            properties = undefined;
+            dynamicContainer = undefined;
         }
     }
-    if (properties === undefined) {
+    if (dynamicContainer === undefined) {
         if (key in (parent.value as Container)) {
             value = (parent.value as Container)[key];
         } else if (altKey !== undefined) {
@@ -113,10 +113,10 @@ export function ManagedReference(
 
             // Now use change to complete the update
             this.change(() => {
-                if (properties) {
-                    (properties as Container)[key] = newValue;
-                    if (altKey !== undefined && altKey in properties) {
-                        delete (properties as Container)[altKey];
+                if (dynamicContainer) {
+                    (dynamicContainer as Container)[key] = newValue;
+                    if (altKey !== undefined && altKey in dynamicContainer) {
+                        delete (dynamicContainer as Container)[altKey];
                     }
                 } else {
                     (parent.value as Container)[key] = newValue;
@@ -131,7 +131,7 @@ export function ManagedReference(
             if (!parent.original) {
                 return undefined;
             }
-            if (properties !== undefined) {
+            if (dynamicContainer !== undefined) {
                 const origProperties = (parent.original as Val.Dynamic)[Val.properties](parent.rootOwner, session);
                 if (key in (origProperties as Container)) {
                     return (origProperties as Container)[key];
@@ -158,10 +158,10 @@ export function ManagedReference(
                 // In transactions, clone the value if we haven't done so yet
                 if (clone && value === this.original) {
                     const newValue = clone(value);
-                    if (properties !== undefined) {
-                        (properties as Container)[key] = newValue;
-                        if (altKey !== undefined && altKey in properties) {
-                            delete (properties as Container)[altKey];
+                    if (dynamicContainer !== undefined) {
+                        (dynamicContainer as Container)[key] = newValue;
+                        if (altKey !== undefined && altKey in dynamicContainer) {
+                            delete (dynamicContainer as Container)[altKey];
                         }
                     } else {
                         (parent.value as Container)[key] = newValue;
@@ -189,11 +189,11 @@ export function ManagedReference(
             }
 
             let value;
-            if (properties !== undefined) {
-                if (key in properties) {
-                    value = (properties as Container)[key];
-                } else if (altKey !== undefined && altKey in properties) {
-                    value = (properties as Container)[altKey];
+            if (dynamicContainer !== undefined) {
+                if (key in dynamicContainer) {
+                    value = (dynamicContainer as Container)[key];
+                } else if (altKey !== undefined && altKey in dynamicContainer) {
+                    value = (dynamicContainer as Container)[altKey];
                 }
             } else {
                 if (key in parent.value) {
--- a/packages/node/src/behaviors/thermostat/AtomicWriteHandler.ts
+++ b/packages/node/src/behaviors/thermostat/AtomicWriteHandler.ts
@@ -29,7 +29,9 @@ const logger = Logger.get("AtomicWriteHandler");
  * The implementation tries to be generic, but is currently only used by the Thermostat cluster, so the atomic write
  * types are imported from there.
  *
- * The logic requires that the cluster behavior implements the following additional events:
+ * The logic requires that the cluster behavior implements the following additional events as "pure Observable()" events,
+ * because the current implementation logic requires error thrown by the event handlers to signal validation failures to
+ * be thrown back to te emitter. This is not the case for official state events.
  * * `${attributeName}$AtomicChanging` - emitted when an attribute is changed as part of an atomic write, before the value
  *   is actually changed. Receives the new value, the old value and the action context as parameters.
  * * `${attributeName}$AtomicChanged` - emitted when an attribute is changed as part of an atomic write, after the value
--- a/packages/node/src/behaviors/thermostat/ThermostatServer.ts
+++ b/packages/node/src/behaviors/thermostat/ThermostatServer.ts
@@ -84,6 +84,9 @@ const schema = ThermostatBehaviorLogicBase.schema!.extend({
  * Important note: To access the current local temperature (including all calibrations applied) please use
  * this.internal.localTemperature because the localTemperature attribute in state might be null depending on the
  * configured features.
+ *
+ * TODO: Currently the general purpose "atomic write" Matter feature is only implemented in this specific cluster because
+ *  only used here so far. Also see information in AtomicWriteHandler.ts.
  */
 export class ThermostatBaseServer extends ThermostatBehaviorLogicBase {
     declare protected internal: ThermostatBaseServer.Internal;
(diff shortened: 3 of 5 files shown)


packages/node/src/behaviors/thermostat/ThermostatServer.ts
```diff
@@ -4,14 +4,994 @@
...
+    #setupValidations() {
+        // Validate existing values to match the constraints at initialization
+        this.#assertUserSetpointLimits("HeatSetpointLimit");
+        this.#assertUserSetpointLimits("CoolSetpointLimit");
+        this.#clampSetpointToLimits("Heat", this.state.occupiedHeatingSetpoint);
+        this.#clampSetpointToLimits("Heat", this.state.unoccupiedHeatingSetpoint);
+        this.#clampSetpointToLimits("Cool", this.state.occupiedCoolingSetpoint);
+        this.#clampSetpointToLimits("Cool", this.state.unoccupiedCoolingSetpoint);
+
+        if (this.events.absMinHeatSetpointLimit$Changing !== undefined) {
```
lauckhart: Could add "maybeReactTo" for this pattern. Personally though I'd probably make a map of validation functions keyed by attribute name then loop over the keys to install

packages/node/src/behaviors/thermostat/ThermostatServer.ts
```diff
@@ -4,14 +4,1000 @@
...
+        );
+    }
+
+    /**
+     * Adjust the state of the thermostat based on the new system mode.
+     * When the AutoMode feature is enabled and the current system mode is Auto, the thermostat running mode
+     * is adjusted instead of the system mode.
+     */
+    protected adjustState(newState: Thermostat.SystemMode) {
+        if (this.features.autoMode && this.state.systemMode === Thermostat.SystemMode.Auto) {
```
lauckhart: Lots of redundancy here.  Seems like the if statement should just select the state attribute and enum & place in variables.  Then everything else is the and

packages/node/src/behaviors/thermostat/ThermostatServer.ts
```diff
@@ -4,14 +4,1000 @@
...
+        }
+        if (this.state.minSetpointDeadBand < 0 || this.state.minSetpointDeadBand > 127) {
+            throw new ImplementationError("minSetpointDeadBand is out of valid range 0..127");
+        }
+        // Store these states internally to prevent changes
+        this.internal.minSetpointDeadBand = this.state.minSetpointDeadBand;
+        this.internal.controlSequenceOfOperation = this.state.controlSequenceOfOperation;
+    }
+
+    override setpointRaiseLower({ mode, amount }: Thermostat.SetpointRaiseLowerRequest) {
```
lauckhart: Consider breaking up into functions, one for each mode, so this method is just a single switch + function calls

packages/node/src/behaviors/thermostat/ThermostatServer.ts
```diff
@@ -4,14 +4,1000 @@
...
+ * It **does not** implement any logic to actually control a thermostat device or to react to temperature changes to e.g.
+ * adjust the ThermostatRunningMode when the Auto feature is supported. This is all up to the specific implementation.
+ *
+ * For local temperature or occupancy values we check if there is a local cluster available on the same endpoint and use
+ * them, alternatively raw measurements can be set in the states externalMeasuredIndoorTemperature and
+ * externallyMeasuredOccupancy. The OutdoorTemperature can be set directly on the attribute if supported.
+ * The RemoteSensing attribute need to be set correctly as needed by the developer to identify the measurement source.
  */
-export class ThermostatServer extends ThermostatBehavior {}
+export class ThermostatBaseServer extends ThermostatBehaviorLogicBase {
```
lauckhart, line 91: There's a ton of branching herein on mode.  Maybe consider a separate `ThermostatMode` class with fields like `isAllowed`, `setpointRange`, `defaults`, `requiresFeature`, `occupiedSetpoint`, `unoccupiedSetpoint`, `temperatureStatus` etc.

With a few metadata fields such as `systemMode`, `raiseLowerMode`, `runningMode` you could select the correct implementation, though you could also do that with switch statements.

And with functions like `setpointRaiseLower` (or maybe "moveSetpoint" if you want to cater to normal people), `adjustState`, `installValidation` etc. I think it'd organize everything nicely.

packages/node/src/behaviors/thermostat/ThermostatServer.ts
Apollon77: I added the maybeReatTo() ... with that it is the same "one liner" then a "push" into an array would be

packages/node/src/behaviors/thermostat/ThermostatServer.ts
```diff
@@ -4,14 +4,994 @@
...
+            throw new StatusResponse.ConstraintErrorError(
+                `min${scope} (${min}) must be less than or equal to max${scope} (${max})`,
+            );
+        }
+        if (max > absMax) {
+            throw new StatusResponse.ConstraintErrorError(
+                `max${scope} (${max}) must be less than or equal to absMax${scope} (${absMax})`,
+            );
+        }
+        if (absMax < min || absMax < absMin) {
```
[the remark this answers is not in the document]
Apollon77: fixed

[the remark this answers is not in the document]
Apollon77: fixed

packages/node/src/behaviors/thermostat/ThermostatServer.ts
Apollon77: Yes I agree. I also had a bit different structure at the beginning but then I had issues and in the end i ended up to "mimic" a lot of the logic of the SDK, so I decided to also leave that there to easy future adjustments. But I will cleanup a bit

Apollon77: And "setpointRaiseLower" is a Matter defined command :-))

packages/node/src/behavior/state/managed/ManagedReference.ts
```diff
@@ -51,10 +53,24 @@ export function ManagedReference(
     const key = primaryKey === "id" ? (id ?? name) : name;
     const altKey = primaryKey === "id" ? (key === name ? undefined : name) : id;
     let value: unknown;
-    if (key in (parent.value as Container)) {
-        value = (parent.value as Container)[key];
-    } else if (altKey !== undefined) {
-        value = (parent.value as Container)[altKey];
+
+    let properties: Val.Struct | undefined;
```
lauckhart: Maybe call this `dynamicContainer` or something, then it'll be a little clearer what the code's doing.

support/models/src/local/ThermostatOverrides.ts
```diff
@@ -27,5 +27,83 @@ LocalMatter.children.push({
             name: "ControlSequenceOfOperation",
             default: FieldValue.None,
         },
+
+        {
```
lauckhart, line 33: It would be nice to inject these into the command list if required in `Scope` like we do with `injectGlobalAttributes`.  I know this is just temporary and maybe this won't ever be used by anyone else but would be good to add notes here and elsewhere like there is in ThermostatServer.

packages/node/src/endpoint/Endpoint.ts
```diff
@@ -256,10 +264,37 @@ export class Endpoint<T extends EndpointType = EndpointType.Empty> {
...
+     * with the actual Behavior type.
+     *
+     * @param type the {@link Behavior} to patch
+     * @param values the values to change
+     */
+    async setStateOf(type: string, values: Val.Struct): Promise<void>;
+
+    async setStateOf(type: Behavior.Type | string, values: Val.Struct) {
+        if (typeof type === "string") {
+            const foundType = this.behaviors.active.find(b => b.id === type);
```
lauckhart: This should use `behaviors.supported[type]`.  Then you don't need the search and it will find behaviors that are not active so they can be activated by `agent.get` below

packages/node/src/endpoint/Endpoint.ts
```diff
@@ -256,10 +264,37 @@ export class Endpoint<T extends EndpointType = EndpointType.Empty> {
...
+            if (foundType === undefined) {
+                throw new ImplementationError(`Behavior ${type} is not active on by this endpoint`);
```
lauckhart: Typo here, would be good to use the same error text as `stateOf`

packages/node/src/behaviors/thermostat/ThermostatServer.ts
```diff
@@ -4,14 +4,1549 @@
...
+ * them, alternatively raw measurements can be set in the states externalMeasuredIndoorTemperature and
+ * externallyMeasuredOccupancy. The OutdoorTemperature can be set directly on the attribute if supported.
+ * The RemoteSensing attribute need to be set correctly as needed by the developer to identify the measurement source.
+ *
+ * The following custom events are provided:
+ * * calibratedTemperature$Changed: Emitted when the measured local temperature changes including any calibration applied. This event is mainly useful when the localTemperatureNotExposed feature is used.
+ *
+ * Important note: To access the current local temperature (including all calibrations applied) please use
+ * this.internal.localTemperature because the localTemperature attribute in state might be null depending on the
+ * configured features.
```
lauckhart, line 86: Maybe TODO or NOTE here that documents that we implement general-purpose "atomic" functionality inline in this cluster because it's the only cluster that uses it and we would like to wait and see how the feature evolves before making it general purpose, and/or link to `AtomicWriteHandler.ts` for more information.

packages/node/src/behaviors/thermostat/AtomicWriteHandler.ts
```diff
@@ -0,0 +1,410 @@
...
+/**
+ * Handles atomic write handling according to Matter definitions.
+ * The implementation tries to be generic, but is currently only used by the Thermostat cluster, so the atomic write
+ * types are imported from there.
+ *
+ * The logic requires that the cluster behavior implements the following additional events:
+ * * `${attributeName}$AtomicChanging` - emitted when an attribute is changed as part of an atomic write, before the value
+ *   is actually changed. Receives the new value, the old value and the action context as parameters.
+ * * `${attributeName}$AtomicChanged` - emitted when an attribute is changed as part of an atomic write, after the value
+ *   is actually changed. Receives the new value, the old value and the action context as parameters.
```
lauckhart, line 38: Is there any reason we need additional events?  Could we use `$Changing` for notification of atomic change and `$Changed` for notification after commit like normal?

Just thinking since you can't do a non-atomic mutation of an atomic attribute the `$Changing` event is kind of useless as a separate thing.

For `$AtomicChanged` it looks like you're using that for as an additional validation step.  We might consider `validateChanges()` method on the behavior or something where you can validate the state as a whole prior to commit (atomic or otherwise).  That seems like it would be general purpose; it would handle the atomic use case but also allow you to perform inter-field validation during normal commit without having to add a listener for every interdependent attribute.

lauckhart (approved):

packages/node/src/behavior/state/managed/ManagedReference.ts
Apollon77: good idea, renamed

support/models/src/local/ThermostatOverrides.ts
Apollon77: Right I missed the comment here. added

packages/node/src/endpoint/Endpoint.ts
Apollon77: kk, adjusted, good catch

packages/node/src/behaviors/thermostat/ThermostatServer.ts
Apollon77: added TODO

packages/node/src/behaviors/thermostat/AtomicWriteHandler.ts
Apollon77: In fact $Changing is called "incrementally" while writing chunks and can not do the full validation. $Changed is usually called at the end "after we have the final value" and is used for some more checks. The "problem" with the original events is that they are automatically installed by the behavior as OnlineEvent and handle errors differently. But for the Atomic Write logic I needed to have the errors being "thrown" like with a pure Observable. I add an info on why it is that way. Yes we could also replace that by methods to validate an incremental and a final but i found the events more convenient to use

PR #2620, opened 2025-10-23
Matter 1.4.2 support

lauckhart:


packages/model/src/aspects/Conformance.ts
```diff
@@ -608,7 +608,7 @@ function ParsedAst(conformance: Conformance, definition: string): Conformance.As
                 return { type: Conformance.Special.Desc };
             }
 
-            if (name === "null") {
+            if (name.toLowerCase() === "null") {
```
Apollon77: nit ... now with 2 times "to LowerCase" we could use a variable :-)

packages/node/test/endpoint/server/ProtocolServiceTest.ts
```diff
@@ -151,29 +151,6 @@ describe("ProtocolServiceTest", () => {
         await node.close();
     });
 
-    it("properly sanitize fabric sensitive reads", async () => {
```
Apollon77, line 154: why removed?

lauckhart: The field isn't marked S anymore and I couldn't find another one in the entire model. Didn't seem worth making a contrived cluster to test something that isn't used

packages/model/src/aspects/Conformance.ts
lauckhart: Oh hah didn't even notice that one.  Changed to a `switch` statement

packages/node/test/endpoint/server/ProtocolServiceTest.ts
Apollon77: lol, kk

PR #2639, opened 2025-10-28
Split entropy from crypto

lauckhart:
Requiring something called "crypto" just to get random numbers seemed a little awkward, so this splits out an "Entropy" abstraction from the Crypto abstraction.  Crypto extends Entropy for implementation convenience.


Apollon77:
I will update ThermostatServer tomorrow ... need to bring a pr anyway

Apollon77:
And maybe with mentioning in changelog

PR #2646, opened 2025-10-29
Add validation of integer bounds

lauckhart:
Enforces integer bounds during validation of supervised values.

### DIFF
--- a/packages/node/src/behavior/state/validation/assertions.ts
+++ b/packages/node/src/behavior/state/validation/assertions.ts
@@ -6,7 +6,7 @@
 
 import { Bytes, isObject } from "#general";
 import { SchemaErrorPath } from "#model";
-import { DatatypeError, Val } from "#protocol";
+import { DatatypeError, IntegerRangeError, Val } from "#protocol";
 
 export function assertNumber(value: Val, path: SchemaErrorPath): asserts value is number {
     if (Number.isFinite(value)) {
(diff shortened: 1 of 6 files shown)


packages/node/src/behavior/state/validation/assertions.ts
```diff
@@ -62,3 +62,59 @@ export function assertArray(value: Val, path: SchemaErrorPath): asserts value is
...
+    assertInt.notNullable[uintName] = createIntAssertion(uintName, 0n, unsignedUpperBound);
+    assertInt.nullable[uintName] = createIntAssertion(`nullable ${uintName}`, 0n, unsignedUpperBound - 1n);
+
+    const signedUpperBound = unsignedUpperBound / 2n;
+    const signedLowerBound = -signedUpperBound + 1n;
+    assertInt.notNullable[intName] = createIntAssertion(intName, signedLowerBound, signedUpperBound);
+    assertInt.nullable[intName] = createIntAssertion(`nullable ${intName}`, signedLowerBound + 1n, signedUpperBound);
+}
+
+function createIntAssertion(name: string, lowerBoundInclusive: bigint, upperBoundExclusive: bigint) {
```
Apollon77, line 96: Why you do  it inclusive vs exclusive and then need to calculate again just for the error message, why not staying inclusive on both bounds?

Apollon77 (approved):
And I think we should have that in changelog in any case :-)

packages/node/src/behavior/state/validation/assertions.ts
lauckhart: It's not calculating again for the error, it's an additional calculation.  Just makes initial calculations a little more complicated but you're right probably cleaner.

PR #2654, opened 2025-10-31
Subscription resiliency for ClientNode

lauckhart:
Adds a "sustain" option to `Subscribe`.  This keeps the subscription alive until closed.

This flag results in fairly significant mode shift for `ClientInteraction#subscribe`:

* If the subscription is not sustained, returns a `PeerSubscription`, which closes on timeout

* If sustained, returns a `SustainedSubscription` which does not close until explicitly aborted.  If the underlying `PeerSubscription` times out, it will establish a new `PeerSubscription` when the node is once again reachable

This commit includes various other bits that are more or less related:

* Wires proper abort support into `ClientInteraction`.  This generally means that closing the interaction won't hang the process regardless of the underlying connection state

* Fixes a `Datasource` bug that prevented refresh of existing references on external change

* Adds an optional `AbortSignal` to `ValueSupervisor.Session` and `InteractionSession` so we can make any async logic associated with local or remote action abortable

* Various updates to documentation and log messages

* Makes `ClientNodeInteraction` subscriptions sustained by default

* Refactors `NetworkRuntime` to use `AbortSignal` to signal closure

* Removes ID-based subscription cancellation from new API components.  Use an `AbortController` or `subscription.close()` instead

* Implements the "resubscribe on timeout" integration test for ClientNode; removes the "determines lack of ping" test because it is redundant

* Makes "auto subscribe" the default for ClientNode, not sure what happened there

* Converts `ClientInteraction` from an automatic environment component to one that you must install explicitly

* Removes redundancy from `ClientInteraction` and streamlines with `await using`


packages/node/src/behavior/system/network/NetworkClient.ts
```diff
@@ -48,36 +48,23 @@ export class NetworkClient extends NetworkBehavior {
...
-        await this.#handleSubscription(true);
     }
 
-    async #handleSubscription(desiredState = this.state.autoSubscribe) {
+    async #handleAutoSubscribeChanged(desiredState = this.state.autoSubscribe) {
         const { isDisabled } = this.state;
         const subscriptionDesired = desiredState && !isDisabled;
 
-        if (subscriptionDesired === this.internal.subscriptionActivated) {
+        if (subscriptionDesired === !!this.internal.defaultSubscription) {
```
Apollon77: For me there is a difference between "a default subscription definition exists" and "subscription is active".  Because ideally developers never touch the default subscription definition but only activate or deactivate it. It is inmyeyes unconvenient is developers that want to have more control over how the subscription happens need to set or clear the "default subscription" definition.

That's why the internal.subscriptionActivated was holding the state if there is currently a subscription active. And the "if" here is trying to determine if - when the developer sets autoSubscribe from false to true we want to establish a new subscription because he just activated it. And wre also discussed thats ok to have no way to "disable" a subscription

packages/node/src/behavior/system/network/NetworkClient.ts
```diff
@@ -199,8 +179,10 @@ export namespace NetworkClient {
         /**
          * If true, automatically subscribe to the provided default subscription (or all attributes and events) when
          * the node is started. If false, do not automatically subscribe.
+         *
+         * The subscription will activate or deactivate automatically if you change this property.
          */
-        autoSubscribe = false;
+        autoSubscribe = true;
```
Apollon77: This was also changed to false as by our discussion. In fact creating a client node via commissioning will set this to true by default.
Adding a client node by "forAddress" (so for communication to another node in the fabric) it used the false defaulkt. But sure we can also adjust that, but then need to change that in "forAddress" in PeerSet. In the end all clientnodes and so also all such subscriptions are configured by us anyway, so in my eyes the default do not matter too much, we just need to make sure it behaves right.

Apollon77: See CommissioningClient L184 for how it gets configured and enabled if needed by default when we commission. If you change the default this needs to be adjusted please to adjust the same goal as before.

And PeerSet#forAddress accordingly too to disable subscriptions by default there

packages/node/src/node/client/ClientEventEmitter.ts
```diff
@@ -56,26 +56,27 @@ export function ClientEventEmitter(node: ClientNode, structure: ClientStructure)
...
+            logger.warn(`Received event ${eventName} for unsupported cluster ${clusterName} on ${endpoint}`);
             return;
         }
 
-        const event = events[eventName];
-        if (event === undefined) {
-            logger.error(`Received unsupported event ${eventName} for cluster ${clusterName} on ${endpoint}`);
-            return;
-        }
+        logger.info(
```
Apollon77, line 72: info? :-)

packages/protocol/src/action/client/subscription/SustainedSubscription.ts
```diff
@@ -0,0 +1,195 @@
...
+
+    export const DefaultRetrySchedule: RetrySchedule.Configuration = {
+        // Protocol-level level happens at the exchange level and is faster; this is an application-level retry.  Retry
+        // more slowly so we do not hammer devices that are experiencing transient errors
+        initialInterval: Seconds(15),
+
+        // Similarly, we have an exchange.  If a device repeatedly fails to establish a subscription, give it plenty of
+        // time to recover.  It's even possible our subscription attempt is invalid for some reason, in which case we
+        // an aggressive interval would be particularly bad form
+        maximumInterval: Hours(1),
```
Apollon77, line 190: I think PairedNode used 10mins here as "best practice" following how Apple does it...

packages/node/src/behavior/system/network/NetworkClient.ts
lauckhart: This is testing this.internal.defaultSubscription, which is an `ActiveSubscription` if the default subscription is active, not this.state.defaultSubscription, which is optional configuration overrides for the default subscription

Which is understandably confusing, maybe what we should do is change this.state.defaultSubscription to this.state.subscriptionDefaults and have it apply to all subscription. Then in the common case it would still just act as configuration for the default subscription, but would have a different name and have the side benefit of allowing you to control how any subscribe() call behaves

packages/node/src/behavior/system/network/NetworkClient.ts
lauckhart: OK I may need a refresher on the forAddress use cases.  Agreed it's kind of irrelevant here because we can set in the factory functions. Probably we shouldn't have them rely on this default but would still be good to rationalize what the default should be

lauckhart: It would be good to get some tests in place for the specific use cases, I was hoping something would break if I undid something problematic...  I want sure where to look to see the "no subscription" flow

packages/node/src/behavior/system/network/NetworkClient.ts
Apollon77: Ahhh ... lol yeahh I was then confused by the naming ... I like the rename idea for this.state

packages/node/src/node/client/ClientEventEmitter.ts
lauckhart: Yeah, I figured an event is significant enough for "just above debug" log level, you don't think so?

packages/node/src/behavior/system/network/NetworkClient.ts
Apollon77: I agree, there are still tests missing, planned to add them with bindings/ota or such because there this use case is relevant.

When it comes to "a good default" that's honestly hard because in fact there are two distinct use cases of ClientNode in the end:
1. Controller usecase where you commission a device into a self-owned fabric. There you usually want to have autoSubscription
2. "Fabric internal inter node communication" for "a reason". These reasons can be
  * logical comm requirements (e.g. OTA), where it is about calling command or such, but no subscription needed
  * bindings (node or group) which is also implementation specific if an subscription is desired or not, would also use default false here
  
We could it even convert it into a mandatory property that needs to be set to avoid confusion and require the "use cases" to define it properly

packages/protocol/src/action/client/subscription/SustainedSubscription.ts
lauckhart: OK, I'm a little fuzzy on how this interacts with the other retry logic in ReconnectableExchange Provider. I was thinking that one does a faster cycle and then this one is kind of "ok the node is there but really doesn't want to talk to us, let's be real polite until it resets" -- but probably am overthinking it

packages/node/src/behavior/system/network/NetworkClient.ts
lauckhart: For bindings we maybe need to see it IRL, I was thinking that that "grab everything we have acls for" might be a good default, but if controllers assign acls too broadly then we might want "grab all attributes from the bound client clusters"...  Those're both autosubscribe = true but don't necessarily retrieve much data

Agreed that it might make sense as a mandatory parameter, we just can't do that here because of how state works unless we make it runtime error

Apollon77: You also need to think about node limitations. if you have a binding where a switch controls a light then in fact no state of the light is needed to be known  (unless the switch can show that state like innovelli), but usually you just send commands.
So bindings in my eyes we ensure the connection and thats it. Yes "Object wise" we can pre init what should be there based on infos of the binding but all rest is up to the developer.

packages/node/src/node/client/ClientEventEmitter.ts
Apollon77: ok, lets stay info for now ... we see if it hurts later

packages/protocol/src/action/client/subscription/SustainedSubscription.ts
Apollon77: As written in Discord: If we make sure that subsciption is also wired to "hey I discovered the node/got a new session/new connection" (aka reconnection) and then retries earlier that that 1h here is fine.

packages/node/src/behavior/system/network/NetworkClient.ts
lauckhart: Yes, we're just talking about the default here

packages/protocol/src/action/client/subscription/SustainedSubscription.ts
lauckhart: My thinking is that the retry loop shouldn't be in effect if we don't have an active session; we should instead be waiting on a "connection".  That's because 1.) a subscription is likely the only thing talking to the peer, and thus the only thing to drive discovery and session establishment, and 2.) then everything wakes up as soon as a session becomes available

lauckhart: As mentioned in discord, added a TODO that this needs to be worked out

packages/node/src/behavior/system/network/NetworkClient.ts
lauckhart: I'll just leave it as false for now given that it isn't really used anywhere anyway

packages/node/src/behavior/system/network/NetworkClient.ts
lauckhart: OK, I ended up renaming this.internal.defaultsubscription to this.internal.activeSubscription...  I didn't see any options that I'd actually want applied to all subscriptions so made internal name the slightly less accurate one

PR #2656, opened 2025-10-31
Enhance Value patcher to also accept attribute ids

Apollon77:
Reading state by Id is already implemented, now we also allow to use endpoint.set() to use alternative ids

### DIFF
--- a/packages/node/src/behavior/state/managed/values/ValuePatcher.ts
+++ b/packages/node/src/behavior/state/managed/values/ValuePatcher.ts
@@ -70,7 +70,7 @@ function StructPatcher(schema: ValueModel, supervisor: RootSupervisor): ValueSup
     // An object mapping name to default value (if any) for sub-structs
     const memberDefaults = {} as Record<string, Val.Struct>;
 
-    // An object mapping name to true iff member is an array
+    // An object mapping name to true if member is an array
     const memberArrays = {} as Record<string, boolean>;
 
     // An object mapping alt keys (numeric IDs) to member names, if numeric IDs are defined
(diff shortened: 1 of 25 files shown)


packages/node/src/behavior/state/managed/values/ValuePatcher.ts
```diff
@@ -105,14 +111,17 @@ function StructPatcher(schema: ValueModel, supervisor: RootSupervisor): ValueSup
             throw new WriteError(path, `cannot patch ${target} because it is not an object`);
         }
 
-        for (const key in changes as Val.Struct) {
+        for (let key in changes as Val.Struct) {
+            let newValue = (changes as Val.Struct)[key];
```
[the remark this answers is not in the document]
Apollon77: This is wanted because we want to read the value we got but process with the adjusted key name

packages/node/src/behavior/state/managed/values/ValuePatcher.ts
```diff
@@ -73,6 +73,9 @@ function StructPatcher(schema: ValueModel, supervisor: RootSupervisor): ValueSup
     // An object mapping name to true iff member is an array
     const memberArrays = {} as Record<string, boolean>;
 
+    // An object mapping alt keys (numeric IDs) to member names, if numeric IDs are defined
+    const memberAltKeys = {} as Record<string, string>;
```
lauckhart, line 77: Is there any reason to store the indirection from id‐>name->patcher vs just storing id->patcher in memberPatchers?

Apollon77: I had it that way at the beginning, but that means I need to fill in the "altKey" in all the helper structs, e.g. also memberArrays and memberDefaults and there I was unsure about potential effects .
Also i was unsure if target object can handle the alternative id too.

With all this I decided for the "a bit more simple approach" and just replace the key after getting the value and leave rest untouched

lauckhart: The underlying objects do handle IDs as keys but probably fine this way

lauckhart (approved):

PR #2657, opened 2025-10-31
Implement Scenesmanagement

Apollon77:
This PR is implementing ScenesManagment as by Spec 1.4.2. This includes:
* Scenes Management Cluster implementation
* enable all tests for Scenes
* Implement all the bound Adapter logics in Groups, OnOff, LevelControl and ColorControl
* Also add the Number/non-null number handling on TLV Level

### DIFF
--- a/packages/node/src/behaviors/scenes-management/ScenesManagementServer.ts
+++ b/packages/node/src/behaviors/scenes-management/ScenesManagementServer.ts
@@ -1085,7 +1085,7 @@ export namespace ScenesManagementServer {
     export type SceneAttributeData = { [key: string]: { [key: string]: boolean | number | bigint | null } };
 
     /** Scene Table Entry as decorated class for persistence */
-    export class ScenesTableEntry {
+    export class ScenesTableEntry implements Omit<ScenesManagement.LogicalSceneTable, "extensionFields"> {
         @field(groupId, mandatory)
         sceneGroupId!: GroupId;
 


packages/node/src/behaviors/level-control/LevelControlServer.ts
```diff
@@ -491,8 +502,18 @@ export class LevelControlBaseServer extends LevelControlBase {
...
-                    // Ensure we move to "on" level before initiating any transition
-                    this.handleOnOffChange(true);
+                    if (this.state.onLevel !== null) {
+                        // Ensure we move to "on" level before initiating any transition
+                        result = this.handleOnOffChange(true);
+                        this.state.blockOnOffCouplingOnce = true; // But block the second call by listener
+                        this.context.transaction.addParticipants({
+                            postCommit: () => {
+                                if (this.state.blockOnOffCouplingOnce) {
+                                    this.state.blockOnOffCouplingOnce = false;
```
lauckhart: I am not 100% sure but if you mutate the state in postCommit it may be that it never commits, because the "automatic commit" will already have run.  So then this change wouldn't take effect...  You might need to use `endpoint.act` to do this in a separate transaction

packages/types/src/tlv/TlvNumber.ts
```diff
@@ -36,13 +38,32 @@ import { TlvWrapper } from "./TlvWrapper.js";
...
         protected readonly lengthProvider: (value: T) => TlvLength,
-        protected readonly min?: T,
-        protected readonly max?: T,
+        readonly baseTypeMin: T,
+        readonly baseTypeMax: T,
+        min?: T,
+        max?: T,
     ) {
         super();
+        if (min !== undefined && min > baseTypeMin) {
```
lauckhart, line 53: Is this what you mentioned is not in the specification?  Maybe leave a comment that this is emulating CHIP and/or is going to be added to spec?

packages/model/src/common/index.ts
```diff
@@ -5,6 +5,7 @@
  */
 
 export * from "./DataModelPath.js";
+export * from "./Datatype.js";
```
lauckhart: Ack, this file is supposed to be deleted, must've been a bad merge.  Use e.g. `uint8.id` instead of `Datatype.uint8`.

packages/node/src/behaviors/scenes-management/ScenesManagementServer.ts
```diff
@@ -4,11 +4,1131 @@
...
+    ValueSigned16 = "valueSigned16",
+    ValueUnsigned32 = "valueUnsigned32",
+    ValueSigned32 = "valueSigned32",
+    ValueUnsigned64 = "valueUnsigned64",
+    ValueSigned64 = "valueSigned64",
+}
+
+/** Mapping from Datatypes to the AttributeValuePair field to use and expect */
+export const DataTypeToSceneAttributeDataMap: Record<string, AttributeValuePairDataFields | undefined> = {
+    [Datatype.bool]: AttributeValuePairDataFields.ValueUnsigned8,
```
lauckhart: R.e. comment on Datatype.ts, you can use`bool.id` instead of `Datatype.bool`.

If you wanted you could also populate this with a function to generate the mapping pretty easily using `metabase.byteSize` in the model and `startsWith("int")` to detect signed values...  But manual seems fine too.

packages/node/src/behaviors/scenes-management/ScenesManagementServer.ts
```diff
@@ -4,11 +4,1131 @@
...
+export class ScenesManagementServer extends ScenesManagementBase {
+    declare state: ScenesManagementServer.State;
+    declare protected internal: ScenesManagementServer.Internal;
+
+    override initialize() {
+        if (!this.state.sceneTableSize) {
+            this.state.sceneTableSize = 128; // Let's use that as a meaningful max for now if not specified by the developer
+        }
+
+        // When fabrics are removed we need to clean up our scenes accordingly
```
lauckhart: Didn't we implement automatic cleanup of all fabric-scoped data?

packages/node/src/behaviors/scenes-management/ScenesManagementServer.ts
```diff
@@ -4,11 +4,1131 @@
...
+    export type SceneAttributeData = { [key: string]: { [key: string]: boolean | number | bigint | null } };
+
+    /** Scene Table Entry including scene values and fabric index */
+    export interface ScenesTableEntry extends Omit<ScenesManagement.LogicalSceneTable, "extensionFields"> {
+        sceneValues: SceneAttributeData;
+        fabricIndex: FabricIndex;
+    }
+
+    /** Scene Table Entry as decorated class for persistence */
+    export class ScenesTableEntry {
```
lauckhart: One thing I've considered is just defining fields with "!" suffix and making a private constructor.  I *think* the decorations would still work though not 100% sure.  That would make implementation smaller and ensure class can only be used as an interface.

Anyway I think if you move the fields above here with a "!" the automatic decoration logic won't see them and you can get rid of the unsafe declaration merge complaint.

lauckhart (approved):

packages/types/src/tlv/TlvNumber.ts
Apollon77: no, was something else. This basically should match with your logic on the behavior level when it comes to definitions. This logic ensures that we also accept higher mins then basetype min

packages/model/src/common/index.ts
Apollon77: Adjusted and removed file

packages/node/src/behaviors/scenes-management/ScenesManagementServer.ts
Apollon77: Adjusted

packages/node/src/behaviors/scenes-management/ScenesManagementServer.ts
Apollon77: yes I did ... lol ... adjusted

packages/node/src/behaviors/scenes-management/ScenesManagementServer.ts
Apollon77: I tried

packages/node/src/behaviors/level-control/LevelControlServer.ts
Apollon77: I moved to internal

packages/node/src/behaviors/scenes-management/ScenesManagementServer.ts
Apollon77: ... and works

PR #2677, opened 2025-11-03
Proper subscription interval config for ClientNode

lauckhart:
* Configures subscription intervals based on physical device properties, similar but not identical to that of PairedNode

* ~Fixes new use of "ClientNode" as "generic behavior" which I hadn't actually tested yet and was broken because the generated class name doesn't actually end in "Client"; fixed that but added a proper flag on the constructors to check instead~ (extracted to separate PR)

* Updated codegen to support above

* Added some reasonable defaults to PowerSourceBehavior initialization

* Converted interval logic to work in `Duration`s instead of seconds

* Added tests for above new logic


packages/node/src/behavior/system/network/NetworkClient.ts
```diff
@@ -182,7 +178,7 @@ export namespace NetworkClient {
          *
          * Newly commissioned nodes default to true.
          */
-        autoSubscribe = false;
+        autoSubscribe = true;
```
Apollon77: Did we not discussed that in last PR? that the "true" is set by Commissioniongclient and so the default should stay false - or we also need to adjust PeerSet#forAddress :-)

packages/node/src/node/NodePhysicalProperties.ts
```diff
@@ -0,0 +1,90 @@
...
+    const network = endpoint.behaviors.typeFor(NetworkCommissioningClient);
+    if (network) {
+        const features = network.schema.supportedFeatures;
+        if (features.has("WI")) {
+            properties.wifiConnected = true;
+        }
+        if (features.has("TH")) {
+            properties.threadConnected = true;
+        }
+        if (features.has("ET")) {
```
Apollon77, line 52: We should consider it "ethernetconnected" when there is no Network cluster existing, then it is "connected to network already, not managed by matter" ... But that logic should more go then into the above method

packages/node/src/node/NodePhysicalProperties.ts
```diff
@@ -0,0 +1,90 @@
...
+            !features.has("WIRED") ||
+            endpoint.behaviors.elementsOf(powerSource).attributes.has("batChargeLevel")
+        ) {
+            if (endpoint.stateOf(PowerSourceClient).status === PowerSource.PowerSourceStatus.Active) {
+                properties.isBatteryPowered = true;
+            }
+        }
+    }
+
+    // Sleepy thread device
```
Apollon77, line 83: Old logic had the ICD check on root level because the cluster shall only be there ... So we do not need to check that for any endpoint :-) but in fact we can also do

packages/node/src/node/NodePhysicalProperties.ts
```diff
@@ -0,0 +1,90 @@
...
+    ) {
+        properties.isThreadSleepyEndDevice = true;
+    }
+
+    // Recurse into children
+    //
+    // Only consider the root endpoint and utility endpoints.  Clusters on application endpoints or sub-endpoints refer
+    // to a different device
+    for (const part of endpoint.parts) {
+        if (part.number !== 0 && part.type.deviceClass !== DeviceClassification.Utility) {
```
Apollon77: Are multiple device types aggregated here to include the proper details? i did not checked the relevant code there now

packages/protocol/src/peer/PhysicalDeviceProperties.ts
```diff
@@ -4,16 +4,16 @@
...
-const DEFAULT_SUBSCRIPTION_CEILING_WIFI_S = 60;
-const DEFAULT_SUBSCRIPTION_CEILING_THREAD_S = 60;
-const DEFAULT_SUBSCRIPTION_CEILING_THREAD_SLEEPY_S = 180;
-const DEFAULT_SUBSCRIPTION_CEILING_BATTERY_POWERED_S = 600;
+const DEFAULT_SUBSCRIPTION_FLOOR_DEFAULT = Seconds(1);
+const DEFAULT_SUBSCRIPTION_FLOOR_ICD = Instant;
+const DEFAULT_SUBSCRIPTION_CEILING_WIFI = Minutes(1);
+const DEFAULT_SUBSCRIPTION_CEILING_THREAD = Minutes(1);
+const DEFAULT_SUBSCRIPTION_CEILING_THREAD_SLEEPY = Minutes(3);
+const DEFAULT_SUBSCRIPTION_CEILING_BATTERY_POWERED = Minutes(5);
```
Apollon77: why not the 10mins as before?

packages/node/src/behavior/system/network/NetworkClient.ts
lauckhart: Oops bad merge, the last PR was extracted from this

packages/node/src/node/NodePhysicalProperties.ts
lauckhart: I figured the ethernet flag should be for ethernet and the "not managed by matter" bit comes from having no flags set

packages/protocol/src/peer/PhysicalDeviceProperties.ts
lauckhart: Because I'm bad at maths

lauckhart:
> @lauckhart Did I spotted all changes to before? So in fact "ethernet not true when no network is there" and the 10->5mins for battery devices?

In the original implementation I didn't see where it avoids descending into an aggregaotor or composing device, which seemed like it could cause problems

packages/node/src/node/NodePhysicalProperties.ts
lauckhart: I didn't realize this was a change, but if the cluster is on a composing Network Management endpoint (or an extension to another device type) shouldn't we count that?

packages/node/src/node/NodePhysicalProperties.ts
lauckhart: They're not aggregated, no, but I was assuming this is the primary device type, which should always be the most specific application type on application endpoints.  If that's not the case then we can fix that

packages/node/src/node/NodePhysicalProperties.ts
Apollon77: or we rename from "ethernet" to "somehow on network" because in the end such devices are usually wired or such

packages/node/src/node/NodePhysicalProperties.ts
Apollon77: no but ICD can only be on root ... so for me was more "i do not do that logic on every endpoint" optimization ... but as said ... fine with both ways

packages/node/src/node/NodePhysicalProperties.ts
Apollon77: Difficult, especially when you think about power source. in fact you can place a power source on root or on an extra endpoint or on the application endpoint, but then it would secondary anyway.
And TBH when we redesigning now, maybe lets ignore "bridged nodes" at all because they might have stuff that confuses us :-) (like "battery info for a sub thing)

packages/protocol/src/peer/PhysicalDeviceProperties.ts
Apollon77: lol :-)

Apollon77:
Re your comment (tht I mnow see after dding all above comments) ... yeahh ... we missed that. :)

packages/node/src/node/NodePhysicalProperties.ts
lauckhart: As discussed in Discord, doing it just on root is probably correct but leaving as is for now just to be extra safe

packages/node/src/node/NodePhysicalProperties.ts
lauckhart: Wait the issue I was referring to is the order of `deviceTypeList`.  If a utility device type was for some reason listed first then we wouldn't notice the application type.  I adjusted in separate PR to ensure we prefer the application device types, but that also turned up that I wasn't actually setting the device classification for ClientNode so fixed that too.

The point of this code is to ignore application endpoints and all of their children for these purposes.  So that skips aggregators and thus bridged nodes.

The comment "Cluster from application endpoint should not affect default" in `NodePhysicalProperties` is where this is tested, although it's running against `ServerNode` so missed the fact that it wouldn't have worked on `ClientNode`.

packages/node/src/node/NodePhysicalProperties.ts
lauckhart: I vote we leave `isEthernetConnected` for known ethernet and leave `isEthernetConnected === isWifiConnected === isThreadConnected === false` to indicate "somehow on network".  Could add an `isConnectedMystically` flag but I'm not sure there's a benefit

packages/node/src/node/NodePhysicalProperties.ts
Apollon77: But as said power source might be on an application endpoint. Then we would. It see it potentially

packages/node/src/node/NodePhysicalProperties.ts
```diff
@@ -0,0 +1,90 @@
...
+    const powerSource = endpoint.behaviors.typeFor(PowerSourceClient);
+    if (powerSource) {
+        const features = powerSource.schema.supportedFeatures;
+        if (
+            features.has("BAT") ||
+            // Perform additional checks because we've encountered devices with incorrect features
+            !features.has("WIRED") ||
+            endpoint.behaviors.elementsOf(powerSource).attributes.has("batChargeLevel")
+        ) {
+            if (endpoint.stateOf(PowerSourceClient).status === PowerSource.PowerSourceStatus.Active) {
```
Apollon77: Please also add the or "unspecified" status as in old logic because also based on "devices in the wold"

packages/node/src/node/NodePhysicalProperties.ts
lauckhart: PR discord added mains power discovery, only prune search for aggregators, and added a bunch of comments

packages/node/src/node/NodePhysicalProperties.ts
lauckhart: added

PR #2680, opened 2025-11-04
Implement VendorId Verification procedure and fix Fabric update logic

Apollon77:
This PR implements the Matter 1.4 2 VendorIdVerification prodecure for devices and partially (beside final certificate chain and DCL checks) controller side.
The controller side will be finished in a follow up PR when we add the DCL logic needed to gfetch certificates for controllers and OTA.

The PR also adds the relevant python test (OPCREDS 3.8) to the manually executed tests. This should fix itself after merge of https://github.com/project-chip/connectedhomeip/pull/41750 and a test container rebuild

Additionally the test execution of this PR also showed that we missed some details in updateFabric logic like adjusting announcements and ensuring the attribute change reporting. This all is fixed too

### DIFF
--- a/packages/matter.js/src/device/DeviceTypes.ts
+++ b/packages/matter.js/src/device/DeviceTypes.ts
@@ -997,7 +997,7 @@ export function getDeviceTypeDefinitionFromModelByCode(code: number): DeviceType
         name: `MA-${device.name.toLowerCase()}`,
         revision: 0,
         code: device.id,
-        deviceClass: capitalize(device.classification) as DeviceClasses,
+        deviceClass: capitalize(device.classification as string) as DeviceClasses,
         superSet: device.type
             ?.replace(/([A-Z])/g, "_$1")
             .substring(1)
--- a/packages/model/src/common/DeviceClassification.ts
+++ b/packages/model/src/common/DeviceClassification.ts
@@ -47,3 +47,20 @@ export enum DeviceClassification {
      */
     Node = "node",
 }
+
+export namespace DeviceClassification {
+    export function isUtility(classification: DeviceClassification | undefined) {
+        switch (classification) {
+            case DeviceClassification.Node:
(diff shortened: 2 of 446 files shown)


packages/protocol/src/certificate/kinds/Icac.ts
```diff
@@ -80,14 +80,19 @@ export class Icac extends OperationalBase<OperationalCertificate.Icac> {
 
         // The subject DN SHALL NOT encode any matter-rcac-id attribute.
         if ("rcacId" in subject) {
-            throw new CertificateError(`Ica certificate must not contain an rcacId.`);
+            throw new CertificateError(`Ica certificate must not contain a rcacId.`);
```
lauckhart: I think "an" is correct because racism starts with a phonetic vow

Apollon77: Ok, I change back

packages/node/src/behavior/system/commissioning/CommissioningServer.ts
```diff
@@ -123,14 +123,16 @@
...
+                            `Failsafe owns a different fabricIndex then ${failsafe.forUpdateNoc ? "updated" : "added"}: ${failsafe.fabricIndex} vs. ${fabricIndex}`,
                         );
                     }
                 }
             } else if (fabricAction === FabricAction.Removed) {
                 // Removed fabric with active Failsafe are ignored but should match the failsafe one
                 if (failsafe.fabricIndex !== fabricIndex) {
-                    throw new MatterFlowError("Failsafe owns a different fabricIndex then removed.");
+                    throw new MatterFlowError(
+                        "Failsafe owns a different fabricIndex then removed: ${failsafe.fabricIndex} vs. ${fabricIndex}",
```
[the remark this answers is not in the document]
Apollon77: fixed

PR #2692, opened 2025-11-06
Add logic to read ASN certs

Apollon77:
This PR adds logic to read and parse ASN certificates into our object structure.

All works fine and we can also verify and the Tlv based certs we had.

One "problem" is still left that I did not got solved: I also added 3 certs from the Matter specs test vectors and they work fine and we can perfectly also regenerate the same ASN.1 structure and data. But verifying the signature is failing for them.
Reason ... unclear. OpenSSL tell they are correct.

### DIFF
--- a/packages/node/src/behavior/system/commissioning/RemoteDescriptor.ts
+++ b/packages/node/src/behavior/system/commissioning/RemoteDescriptor.ts
@@ -96,10 +96,10 @@ export namespace RemoteDescriptor {
 
         if (pairingHint !== undefined) {
             result.PH = pairingHint;
-        }
 
-        if (pairingInstructions !== undefined) {
-            result.PI = pairingInstructions;
+            if (pairingInstructions !== undefined) {
+                result.PI = pairingInstructions;
+            }
         }
 
         if (sessionIntervals !== undefined) {
--- a/packages/protocol/src/advertisement/PairingHintBitmap.ts
+++ b/packages/protocol/src/advertisement/PairingHintBitmap.ts
@@ -59,7 +59,7 @@ export const PairingHintBitmap = {
      * The Device will enter Commissioning Mode when reset button is pressed for N seconds. The exact value of N SHALL
      * be made available via PI key.
      */
-    pressRestButtonForNumberOfSeconds: BitFlag(8),
+    pressResetButtonForNumberOfSeconds: BitFlag(8),
 
     /**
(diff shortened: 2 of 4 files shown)


packages/protocol/src/certificate/kinds/X509Base.ts
```diff
@@ -341,10 +346,390 @@ export abstract class X509Base<CT extends X509Certificate> {
...
+        if (!extensions) {
+            throw new CertificateError("Invalid extensions structure");
+        }
+
+        for (const ext of extensions) {
+            const { [DerKey.Elements]: extElements } = ext;
+            if (!extElements || extElements.length < 2) continue;
+
+            const oid = extElements[0][DerKey.Bytes];
+            const oidHex = Bytes.toHex(oid);
```
lauckhart: Might be a slight improvement to operate with bignums instead of strings

packages/protocol/src/certificate/kinds/X509Base.ts
```diff
@@ -341,10 +346,390 @@ export abstract class X509Base<CT extends X509Certificate> {
...
+                        const hexString = Bytes.toString(valueBytes);
+                        value = BigInt("0x" + hexString);
+                        break;
+                    }
+                    case "icacId":
+                    case "rcacId":
+                    case "vvsId": {
+                        // 8-byte hex string -> BigInt, but convert to number if it fits
+                        const hexString = Bytes.toString(valueBytes);
+                        const bigIntValue = BigInt("0x" + hexString);
```
lauckhart, line 433: Use Bytes.toBigInt()

packages/protocol/src/certificate/kinds/X509Base.ts
```diff
@@ -341,10 +346,390 @@ export abstract class X509Base<CT extends X509Certificate> {
...
+            let valueIndex = 1;
+            if (extElements.length > 2 && extElements[1][DerKey.TagId] === DerType.Boolean) {
+                valueIndex = 2; // Skip critical flag
+            }
+
+            const valueOctetString = extElements[valueIndex][DerKey.Bytes];
+            const valueNode = DerCodec.decode(valueOctetString);
+
+            switch (oidHex) {
+                case "551d13": // basicConstraints
```
lauckhart: A namespace of constants would be nice

PR #2700, opened 2025-11-07
Subscription data adjustments

Apollon77:
This is adding more checks, a reasonable maximum for the maxCeiling value and some tweaks around Subscriptions


packages/types/src/protocol/messages/TlvSubscribeRequest.ts
```diff
@@ -21,10 +21,10 @@ export const TlvSubscribeRequest = TlvObject({
     keepSubscriptions: TlvField(0, TlvBoolean),
 
     /** The requested minimum interval boundary floor in seconds. */
-    minIntervalFloorSeconds: TlvField(1, TlvUInt16),
+    minIntervalFloor: TlvField(1, TlvUInt16),
```
lauckhart: Is this a good idea?  It seems confusing that we have fields called "minIntervalFloor" that are sometimes a `Duration` (so milliseconds) and sometimes a number in seconds depending on where they come from.  Seems like it'd be prone to error.

Apollon77: Ok, I change back,rigtj ... just thought better to be near to spec

lauckhart: Ah, yes good to match matter names but in this case I think there's a good reason not to

lauckhart (approved):

PR #2701, opened 2025-11-07
Adds basic checks for Large Message commands

Apollon77:
as defined in 1.4.2 spec. Right now we always throw errors because we do not support it


packages/node/src/node/integration/ProtocolService.ts
```diff
@@ -502,8 +502,20 @@ function clusterTypeProtocolOf(backing: BehaviorBacking): ClusterTypeProtocol |
...
+                } = member;
+
+                const command = {
+                    id: id as CommandId,
+                    responseId,
+                    requestTlv,
+                    responseTlv,
+                    limits,
+                    name,
+                    largeMessage: largeMessage ? largeMessage : undefined,
```
lauckhart: I think this should go into limits instead.  Timed is there too.  Then it would get passed around for free without an additional field

packages/node/src/node/client/PeerBehavior.ts
```diff
@@ -168,7 +172,7 @@ function generateType(analysis: ShapeAnalysis, baseType: Behavior.Type): Cluster
         isCloned = true;
     }
 
-    function implementCommand(command: string) {
+    function implementCommand(command: string, largeMessage: boolean) {
```
lauckhart: Hard to tell for sure but I don't think we should be passing metadata in here...  The command implementation can get this from the schema when invoked so the logic doesn't need to run for every command generated, only those invoked

packages/node/src/node/client/PeerBehavior.ts
```diff
@@ -182,6 +186,7 @@ function generateType(analysis: ShapeAnalysis, baseType: Behavior.Type): Cluster
                             fields,
                         }),
                     ],
+                    largeMessage,
```
lauckhart: I think this may be misplaced as it is a property of the individual commands. In the server context I think it means "the channel supports large messages".  In the client context we should be able to determine "this action requires a channel that supports large messages" by examining the constituent commands.

Right now we don't have that information because this flag isn't in the ClusterType Command object, but I think it would be better to just leave out until we can fix that. Especially since there is other validation we don't do on the client, instead relying on error handling in the peer

packages/node/src/node/integration/ProtocolService.ts
Apollon77: But that would mix quality and access then, really sure?

packages/node/src/node/client/PeerBehavior.ts
Apollon77: The problem is that the command execution logic does not use schema. That uses ClusterType.commands stuff which is "old style" and only has the "tlv schema". So that would mean enhancing generation to have is in the Cluster type stuff
or what do you mean?

packages/node/src/node/client/PeerBehavior.ts
Apollon77: Yes and no ... because if you want to send multiple commands in one message then we need to know as soon as one of the command needs that because we only send one thing, or?
Right now the official API bound way only includes one command, so only a dev would send multiple commands ... But sure lets discuss how we want to have it.

Apollon77: If we say that devs just pump in commands and we decide based on device capabilities how to really send them out and such then yes we need to move it

packages/node/src/node/integration/ProtocolService.ts
lauckhart: Yes I think that's ok

Apollon77: kk

packages/node/src/node/client/PeerBehavior.ts
lauckhart: The command functions can use `this.schema` to access the schema when they execute

packages/node/src/node/client/PeerBehavior.ts
lauckhart: Yes, we'd need to loop over all commands to check for L quality

I was specifically thinking of the multi-invoke case...  We should enforce this limit there too but with current design it would be "we'll enforce this limit if you tell us about it" vs "we'll enforce this limit if matter requires it"

Apollon77: As discussed added TODO instead logic for new client logic for now

PR #2705, opened 2025-11-08
ClientNode structure management updates

lauckhart:
* Add `mutex.lock()`.  This uses resource management (Disposable) to make locking more natural in async functions

* Fix bug in "|" operator conformance applicability assessment

* Fix bug in `DecodedBitmap` detection of feature map attributes

* Add `EncodedBitmap` which is the inverse of `DecodedBitmap`

* Remove various circular deps in `@client/node`

* Rename `createType` to `ClusterBehaviorType` and move into separate file from utility functions & types

* Add "client" mode to `ClusterBehaviorType` that disables some generation features that only apply for server behaviors

* Add `commandFactory` parameter to `ClusterBehaviorType` that allows you to generated commands as something other than `Behavior.unimplemented`

* Unconditionally generate all command methods, not just those that are conformant, so there is always a function present for behaviors retrieved via *Client

* Added `maybeStateOf` as convenience on `Endpoint`

* Removed `EndpointInitializer#finalizeType`.  Instead, `ClientEndpointInitializer` instruments cluster behaviors as necessary when it creates the backing

* Modified `PeerBehavior` factory to support two kinds of shapes:

  * "discovered" is a base behavior + metadata; this is the original use case

  * "known" provides a behavior that is instrumented "as is"; this is the "add support for this cluster because I know it exists" use case

* Add support to `ClientStructure` to erase and delete endpoints when they are removed

* Modify `ClientStructure` so it defers installing and deleting endpoints until an entire subscription response is consumed; previously a new endpoint would appear partially initialized and wouldn't be useful

* Lock `node.lifecycle.mutex` when mutating `ClientNode` endpoints to ensure changes serialize

* Add new events to `Peers` that make it more convenient to instrument `ClientNode` endpoints:

  * `node.peers.endpointInstalled(SomeEndpoint)` returns an observable that emits when a new endpoint is installed of the specified type

  * `node.peers.clusterInstalled(SomeClusterBehavior)` returns an obserable when a new endpoint is installed that supports the specified cluster

* Added support overrides to the `PeerBehavior` cache key, without which we would have incorrect cache hits

* Modify `Peers` to erase and remove the `ClientNode` when a peer broadcasts the "leave" event for the correct fabric

* Added `fabric.leave()` to support proper shutdown when leaving a fabric; this ensures events emit (previously the "leave" event was dropped) and the exchange is closed properly

* Modify `OperationalCredentialsServer` to exit the fabric using `fabric.leave()`

* Added `ClientNode` tests for new functionality

* Added `fabrics.forDescriptor()` utility function to retrieve a fabric with its root public key and fabric ID (fabric IDs are only guaranteed unique in the scope of a fabric authority)


Apollon77 (approved):
Great optimizations

PR #2728, opened 2025-11-15
Expose DCL client functionality via matter.js shell

Apollon77:
Enhance the nodejs-shell with a new "dcl" sub-area that exposes all query methods from the DCL client in packages/protocol/src/dcl. This allows users to query the CSA Distributed Compliance Ledger (DCL) for Matter compliance information directly from the shell.

Features:
- fetch-root-certificates: List all approved PAA root certificates
- fetch-root-certificate: Get detailed info for a specific certificate
- fetch-model: Get device model information by VID/PID
- fetch-model-versions: List available software versions for a device
- fetch-model-version: Get detailed info for a specific software version
- Optional --test flag to use test DCL instead of production (default)

All commands support both decimal and hex (0x prefix) input for VID/PID parameters and provide formatted JSON output.

### DIFF
--- a/packages/general/src/log/Diagnostic.ts
+++ b/packages/general/src/log/Diagnostic.ts
@@ -335,11 +335,14 @@ export namespace Diagnostic {
     /**
      * Convert a number or bigint to a hex string which is prefixed by "0x" for logging purposes
      * @param value The number or bigint to convert
-     * @param padding Optional padding width (pads with leading zeros)
+     * @param padding Optional padding width (pads with leading zeros). When padding is specified, the "0x" prefix is omitted for cleaner aligned output.
      */
     export function hex(value: number | bigint, padding?: number) {
         const hexString = value.toString(16);
-        return `0x${padding !== undefined ? hexString.padStart(padding, "0") : hexString}`;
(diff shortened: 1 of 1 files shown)


packages/general/src/log/Diagnostic.ts
```diff
@@ -334,9 +334,12 @@ export namespace Diagnostic {
...
     /**
      * Convert a number or bigint to a hex string which is prefixed by "0x" for logging purposes
+     * @param value The number or bigint to convert
+     * @param padding Optional padding width (pads with leading zeros)
      */
-    export function hex(value: number | bigint) {
-        return `0x${value.toString(16)}`;
+    export function hex(value: number | bigint, padding?: number) {
+        const hexString = value.toString(16);
+        return `0x${padding !== undefined ? hexString.padStart(padding, "0") : hexString}`;
```
lauckhart: In a nice aligned presentation the 0x is going to be noise, I think it'd leave it off if padding.  Padded presentation should make it fairly clear it's hex

Apollon77: i adjust that

PR #2734, opened 2025-11-16
Adjust Thermostat logic

Apollon77:
We wrongly loked for local Measurement servers but these are clients and will be added when we add bindings

### DIFF
--- a/packages/node/src/behaviors/thermostat/ThermostatServer.ts
+++ b/packages/node/src/behaviors/thermostat/ThermostatServer.ts
@@ -6,9 +6,6 @@
 
 import { ActionContext } from "#behavior/context/ActionContext.js";
 import { ValueSupervisor } from "#behavior/supervision/ValueSupervisor.js";
-import { OccupancySensingServer } from "#behaviors/occupancy-sensing";
-import { TemperatureMeasurementServer } from "#behaviors/temperature-measurement";
-import { OccupancySensing } from "#clusters/occupancy-sensing";
 import { Thermostat } from "#clusters/thermostat";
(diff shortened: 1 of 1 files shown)


packages/node/src/behaviors/thermostat/ThermostatServer.ts
```diff
@@ -415,32 +415,22 @@ export class ThermostatBaseServer extends ThermostatBehaviorLogicBase {
...
-            }
-            this.reactTo(
-                this.agent.get(TemperatureMeasurementServer).events.measuredValue$Changed,
-                this.#handleMeasuredTemperatureChange,
+
+        // TODO: Check for installed Bindings for TemperatureMeasurement cluster when supported and use the value
+
+        if (this.state.externalMeasuredIndoorTemperature === undefined) {
+            logger.warn(
+                "No local TemperatureMeasurement cluster available and externalMeasuredIndoorTemperature state not set. Setting localTemperature to null",
```
lauckhart, line 423: In a little confused here, why mention a local cluster?  It doesn't seem like we check one anymore

Honestly though reading a value from a local cluster seems like a nice feature even if it's not contemplated in the spec.  If we were to keep that support I'd make the endpoint configurable

lauckhart (approved):

packages/node/src/behaviors/thermostat/ThermostatServer.ts
Apollon77: Oh yeahh I might need to adjust the text here too.
Formally "same endpoint" would be against spec but we could add it as a custom feature to provide an endpoint "locally" in the node. I can adjust that ... but "convenience" feature

PR #2736, opened 2025-11-17
Integrate OTA and certificate DCL management

Apollon77:
This introduces two basic services to handle PAA certificates and OTA via DCL. It is not yes wired in but available via shell for basic management. More integration comes in follow up PRs.

### DIFF
--- a/packages/general/test/crypto/StandardCryptoTest.ts
+++ b/packages/general/test/crypto/StandardCryptoTest.ts
@@ -185,19 +185,5 @@ describe("StandardCrypto", () => {
                 });
             });
         });
-
-        describe("Node.js-only Algorithms (Should Throw in StandardCrypto)", () => {
-            it("throws NotImplementedError for SHA-512/224 (ID 10)", async () => {
-                await expect(crypto.computeHash(testData, "SHA-512/224")).to.be.rejectedWith(/Unrecognized/);
-            });
-
-            it("throws NotImplementedError for SHA-512/256 (ID 11)", async () => {
-                await expect(crypto.computeHash(testData, "SHA-512/256")).to.be.rejectedWith(/Unrecognized/);
-            });
-
-            it("throws NotImplementedError for SHA3-256 (ID 12)", async () => {
-                await expect(crypto.computeHash(testData, "SHA3-256")).to.be.rejectedWith(/Unrecognized/);
-            });
-        });
     });
 });


packages/nodejs-shell/src/shell/cmd_ota.ts
```diff
@@ -0,0 +1,552 @@
...
+                    async argv => {
+                        const { keyname, vid, pid, mode } = argv;
+
+                        let deletedCount: number;
+
+                        if (keyname) {
+                            // Delete by keyname
+                            deletedCount = await theNode.otaService.delete({
+                                filename: keyname,
+                            });
```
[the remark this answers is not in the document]
Apollon77: adjusted

packages/general/src/crypto/Crypto.ts
```diff
@@ -29,6 +29,42 @@ export const CRYPTO_EC_KEY_BYTES = 32;
 export const CRYPTO_AUTH_TAG_LENGTH = 16;
 export const CRYPTO_SYMMETRIC_KEY_LENGTH = 16;
 
+/**
+ * Hash algorithms identified by IANA Hash Function identifiers.
+ * Based on FIPS 180-4 Section 6.2 and FIPS 202.
+ *
+ * The enum values are the FIPS-defined algorithm IDs.
+ */
+export enum HashAlgorithm {
```
lauckhart: Use of a TS enum is non-idiomatic and adds runtime cost.  Use of the FIPS ID as primary identifier is very non-idiomatic and seems unnecessary.

Can we instead use an idiomatic string type that adheres to the WebCrypto standards for hash names with extensions for hashes that WebCrypto doesn't support?

If we need the FIPS mapping then fine to use an enum for that, but that should be `HashAlgorithmFipsId` or something, but not the key we pass everywhere.

packages/general/src/crypto/NodeJsStyleCrypto.ts
```diff
@@ -28,6 +30,19 @@ import type * as NodeJsCryptoApi from "node:crypto";
 // Ensure we don't reference global crypto accidentally
 declare const crypto: never;
 
+/**
+ * Mapping of hash algorithms to Node.js crypto algorithm names.
+ * Node.js supports all algorithms.
+ */
+const NODE_CRYPTO_ALGORITHMS: Record<HashAlgorithm, string> = {
```
lauckhart: If algorithm identifiers are strings then we should be able to do this conversion programmatically without a manual mapping table (using toLowerCase + replace)

packages/general/src/crypto/NodeJsStyleCrypto.ts
```diff
@@ -114,22 +129,34 @@ export class NodeJsStyleCrypto extends Crypto {
...
-    computeSha256(
+    computeHash(
+        algorithm: HashAlgorithm,
         data: Bytes | Bytes[] | ReadableStreamDefaultReader<Bytes> | AsyncIterator<Bytes>,
     ): MaybePromise<Bytes> {
-        const hasher = this.#crypto.createHash(CRYPTO_HASH_ALGORITHM);
+        // Validate algorithm
+        const nodeAlgorithmName = NODE_CRYPTO_ALGORITHMS[algorithm];
+        if (!nodeAlgorithmName) {
+            throw new NotImplementedError(`Unsupported hash algorithm ${HASH_ALGORITHM_NAMES[algorithm] ?? algorithm}`);
```
lauckhart: Maybe better to let the underlying platform handle this assessment, especially since I suspect there will be "node.js-style crypto" that doesn't support some of these algorithms

packages/general/src/crypto/NodeJsStyleCrypto.ts
```diff
@@ -114,22 +129,34 @@ export class NodeJsStyleCrypto extends Crypto {
...
+            // Handle streaming data (ReadableStreamDefaultReader or AsyncIterator)
             let iteratorFunc: () => Promise<IteratorResult<Bytes>>;
             if ("read" in data && typeof data.read === "function") {
                 iteratorFunc = data.read.bind(data);
             } else if ("next" in data && typeof data.next === "function") {
                 iteratorFunc = data.next.bind(data);
             } else {
-                throw new ImplementationError("Invalid data type for computeSha256");
+                throw new ImplementationError(
+                    `Invalid data type for computeHash with algorithm ${HASH_ALGORITHM_NAMES[algorithm]}`,
```
lauckhart: This is an example of where numeric enum kind of pollutes code and makes things more verbose...  So `HASH_ALGORITHM_NAMES[algorithm]` instead of just `algorithm`

packages/general/src/crypto/Crypto.ts
```diff
@@ -59,12 +95,23 @@ export abstract class Crypto extends Entropy {
     abstract decrypt(key: Bytes, data: Bytes, nonce: Bytes, aad?: Bytes): Bytes;
 
     /**
-     * Compute the SHA-256 hash of a buffer.
+     * Compute a cryptographic hash using the specified algorithm.
      */
-    abstract computeSha256(
+    abstract computeHash(
```
lauckhart, line 90: Given that the vast majority of calls are SHA-256, what if we make the algorithm second parameter with a default of "SHA-256"?

packages/general/src/crypto/StandardCrypto.ts
```diff
@@ -97,14 +108,29 @@ export class StandardCrypto extends Crypto {
         });
     }
 
-    computeSha256(buffer: Bytes | Bytes[] | ReadableStreamDefaultReader<Bytes> | AsyncIterator<Bytes>) {
+    computeHash(
+        algorithm: HashAlgorithm,
+        buffer: Bytes | Bytes[] | ReadableStreamDefaultReader<Bytes> | AsyncIterator<Bytes>,
+    ) {
+        // Check if Web Crypto supports this algorithm
+        const webCryptoName = WEB_CRYPTO_ALGORITHMS[algorithm];
```
lauckhart: Similar to comment in node.js version...  Let's let WebCrypto do this for us.  Not sure in the node case but for WebCrypto use of `NotSupportedError` is normative so we just need to check for that and wrap with a `NotImplementedError`

packages/nodejs-shell/src/shell/cmd_cert.ts
```diff
@@ -0,0 +1,176 @@
...
+                            describe: "Filter by vendor ID (hex format like 0xFFF1 or decimal)",
+                            type: "string",
+                        });
+                    },
+                    async argv => {
+                        const { vendorId: vendorIdStr } = argv;
+
+                        await theNode.start();
+                        if (!theNode.certificateService) {
+                            throw new Error("Certificate service not initialized");
```
lauckhart: Maybe an accessor for certificateService that throws if the service is unavailable?  Maybe `certificateService` and `maybeCertificateService` (if you need the latter).  Or could use `env.get(CertificateService)`

packages/protocol/src/certificate/kinds/Certificate.ts
```diff
@@ -473,10 +473,19 @@ export namespace X509Base {
         return result;
     }
 
+    /** These extensions are minimum required for all Matter certificates. */
+    export const REQUIRED_PAA_EXTENSIONS = ["basicConstraints", "keyUsage", "subjectKeyIdentifier"];
```
lauckhart, line 477: Given the way we're using this file maybe we could just call "X509" instead of "X509Base".  It's a namespace as much as a base class

packages/protocol/src/dcl/DclCertificateService.ts
```diff
@@ -0,0 +1,583 @@
...
+    Timer,
+} from "#general";
+import { Paa } from "../certificate/kinds/AttestationCertificates.js";
+import { DclClient, MatterDclError } from "./DclClient.js";
+import { DclPkiRootCertificateSubjectReference } from "./DclRestApiTypes.js";
+
+const logger = Logger.get("DclCertificateService");
+
+// GitHub repository for development/test certificates
+const GITHUB_OWNER = "project-chip";
```
lauckhart, line 29: We have a github client in `support/codegen/src/util/github.ts`.  Maybe we could move that into `@matter/general` so it can be shared?

packages/general/src/util/MockFetch.ts
```diff
@@ -0,0 +1,186 @@
...
+            responseData,
+            options,
+        });
+    }
+
+    /**
+     * Install this mock as the global fetch implementation.
+     * Must be called before the code under test executes fetch calls.
+     */
+    install() {
```
lauckhart, line 82: We could also consider passing in `fetch` for components that require it.  Would also allow us to instrument

packages/protocol/test/util/fetch-mock.ts
```diff
@@ -0,0 +1,186 @@
...
+ * const fetchMock = new FetchMock();
+ * fetchMock.addResponse("/api/data", { result: "success" });
+ * fetchMock.install();
+ *
+ * // Your code that calls fetch("/api/data") will now receive the mocked response
+ *
+ * fetchMock.uninstall();
+ * ```
+ */
+export class FetchMock {
```
lauckhart: Cool, I would move into general, rename to `MockFetch`, and make it extend `Callable` so you can just invoke like normal fetch

packages/react-native/src/crypto/ReactNativeCrypto.ts
```diff
@@ -121,6 +124,29 @@ export class ReactNativeCrypto extends StandardCrypto {
...
+        data: Bytes | Bytes[] | ReadableStreamDefaultReader<Bytes> | AsyncIterator<Bytes>,
+    ) {
+        // Block unsupported algorithms
+        if (
+            algorithm === HashAlgorithm.SHA512_224 ||
+            algorithm === HashAlgorithm.SHA512_256 ||
+            algorithm === HashAlgorithm.SHA3_256
+        ) {
+            throw new NotImplementedError(
+                `Hash algorithm ${HASH_ALGORITHM_NAMES[algorithm]} (${HashAlgorithm[algorithm]}) is not supported in ReactNativeCrypto. Use Node.js-based crypto implementation.`,
```
lauckhart: Suggesting use of node.js crypto here I don't think makes sense since if they encounter this error node.js is not an option

lauckhart (approved):

packages/general/src/crypto/Crypto.ts
Apollon77: Ok I adjust that. was basically my idea that way but other way around fine too

Apollon77: No mapping needed anymore

packages/protocol/src/dcl/DclCertificateService.ts
Apollon77: moving and enhancing.

lauckhart (approved):
Looks good!

PR #2738, opened 2025-11-18
Protocol reorganization #1

lauckhart:
This is a large amount of refactoring of protocol package.  High-level goal is to simplify and make it easier to reason about state.

Individual commits have more detail.  Would be reasonable to review by commit, and I could also potentially break out into separate PRs although I worry I may end up in merge and/or CI hell that way.

Here are broad strokes of some of the more major changes:

* Organize logic and data related to a specific peer in a single object that represents the peer.  This is a WIP, although the primary part that remains is reworking network state management which is the end goal I'm shooting for

* Factors away ChannelManager.  Now if you have a session you can use it unless it's closed.  This is how we used sessions logically but under the hood we needed to continually bounce between sessions and channels

* Centralize protocol mocking.  The was primarily to reduce other refactoring effort as we have evolved a number of different mock systems with overlapping responsibility.  Also a WIP


packages/protocol/src/session/Session.ts
```diff
@@ -70,7 +73,15 @@ export interface SessionParameters extends SessionIntervals {
 
 export type SessionParameterOptions = Partial<SessionParameters>;
 
+export class NonOperationalSession extends ImplementationError {
```
Apollon77: Should have Error in the name as ending?!

packages/protocol/src/session/Session.ts
```diff
@@ -218,11 +229,16 @@ export abstract class Session {
...
-        closeAfterExchangeFinished?: boolean,
-        flushSubscriptions?: boolean,
-    ): Promise<void>;
+    async destroy(
+        _sendClose?: boolean,
+        _closeAfterExchangeFinished?: boolean,
+        _flushSubscriptions?: boolean,
+    ): Promise<void> {
+        if (this.#channel) {
+            await this.#channel.close();
```
Apollon77, line 315: not 100% sure if this is right ... because no clue if maybe then acks or such things fail?

packages/protocol/test/bdx/BdxTest.ts
```diff
@@ -87,7 +87,7 @@ describe("BdxTest", () => {
                         expect(receivedData).to.deep.equal(data);
                     },
                 });
-            });
+            }).timeout(1e9);
```
Apollon77: leftover?

packages/matter.js/src/MatterController.ts
```diff
@@ -535,9 +535,9 @@ export class MatterController {
         return this.#node!.env.get(InteractionClientProvider).connect(this.#fabric!.addressOf(peerNodeId), options);
     }
 
-    createInteractionClient(peerNodeIdOrChannel: NodeId | MessageChannel, options: PeerConnectionOptions = {}) {
-        if (peerNodeIdOrChannel instanceof MessageChannel) {
-            return this.#node!.env.get(InteractionClientProvider).getInteractionClientForChannel(peerNodeIdOrChannel);
+    createInteractionClient(peerNodeIdOrChannel: NodeId | SecureSession, options: PeerConnectionOptions = {}) {
```
Apollon77: if it is not "channel" any lre then maybe adjust the variable name "peerNodeIdOrChannel"

packages/protocol/src/session/Session.ts
```diff
@@ -29,7 +29,7 @@ import { SessionParameters } from "./SessionParameters.js";
 
 export class NonOperationalSession extends ImplementationError {
     constructor(session: Session) {
-        super(`Session ${session.name} has no channel assigned`);
+        super(`Session ${session.name} has ended`);
```
Apollon77: formally could also have never been initialized ... but yeah common usecase is ended

packages/protocol/src/peer/ControllerCommissioner.ts
```diff
@@ -382,9 +382,6 @@ export class ControllerCommissioner {
             }
             throw e;
         }
-
-        await unsecureSession.destroy();
```
Apollon77, line 386: Where the insecuresession is closed now? SHould that go into "finally"? (and when I see now ... yeahhh missing in error case before too)

packages/matter.js/src/MatterController.ts
```diff
@@ -426,7 +426,7 @@ export class MatterController {
 
     async disconnect(nodeId: NodeId) {
         this.#construction.assert();
-        return this.#peers!.disconnect(this.#fabric!.addressOf(nodeId));
+        return this.#peers?.get(this.#fabric!.addressOf(nodeId))?.delete();
```
Apollon77: That should be peer.close() ... because we do not want to forget the node on a simple "disconnect"

packages/protocol/src/peer/Peer.ts
```diff
@@ -84,13 +86,14 @@ export class Peer {
...
-        await this.close();
+        logger.info("Removing", Diagnostic.strong(this.toString()));
+        await this.close(false);
         await this.#context.deletePeer(this);
         await this.#context.sessions.deleteResumptionRecord(this.address);
     }
 
     /**
-     *
+     * Remove the peer
```
Apollon77: is the comment really correct? because basically only sessions for the peer are removed, or?

packages/node/src/behavior/system/commissioning/CommissioningServer.ts
```diff
@@ -407,6 +403,8 @@ export namespace CommissioningServer {
         fabricsChanged = Observable<[fabricIndex: FabricIndex, action: FabricAction]>();
         enabled$Changed = AsyncObservable<[context: ActionContext]>();
     }
+
+    export type FabricAction = "added" | "deleted" | "updated";
```
Apollon77, line 407: Should add a "breaking" note in changelog for that just to be sure to not confuse devs

packages/protocol/src/session/Session.ts
lauckhart: yeah

packages/protocol/src/session/Session.ts
lauckhart: Yeah we could consider changing, it does show up in kind of weird places when there are bugs

packages/protocol/src/session/Session.ts
lauckhart: I haven't finished working through exchange management yet so not 100% sure there may not be a dangling promise related to ACKs remaining...  But the goal is that this is not invoked so long as there is still anything in flight

packages/protocol/test/bdx/BdxTest.ts
lauckhart: yes, removed

packages/protocol/src/session/Session.ts
lauckhart: I consolidated with SessionClosedError so this one has gone away.

packages/matter.js/src/MatterController.ts
lauckhart: done

packages/matter.js/src/MatterController.ts
lauckhart: fixed

packages/protocol/src/peer/ControllerCommissioner.ts
lauckhart: I made the `catch` block close the session.  I added a comment r.e. the success case; in that case the channel is moved from the unsecured session to the secure session, so the unsecured session shouldn't be closed

packages/protocol/src/peer/Peer.ts
lauckhart: It looks like the comment is unfinished but this does more than remove sessions; it stops discovery and reconnection, terminates sessions and removes it from the PeerSet

packages/node/src/behavior/system/commissioning/CommissioningServer.ts
lauckhart: ok, external folks use this?

Apollon77: We at least had it mentioned in the migration docs, so maybe

packages/general/src/util/Observable.ts
```diff
@@ -735,6 +738,18 @@ export class ObserverGroup {
         this.#defaultTarget = target;
     }
 
+    on<T extends any[], R>(
```
Apollon77, line 800: We should also have this Commit and "change" as potentially breaking in the changelog because it might affect users

packages/node/src/node/ClientNode.ts
```diff
@@ -115,10 +116,16 @@ export class ClientNode extends Node<ClientNode.RootEndpoint> {
...
+        this.lifecycle.change(EndpointLifecycle.Change.Destroying);
+
         if (this.lifecycle.isCommissioned) {
             this.statusUpdate("decommissioning");
 
-            await this.act("decommission", agent => agent.commissioning.decommission());
+            try {
+                await this.act("decommission", agent => agent.commissioning.decommission());
+            } catch (e) {
+                logger.error(`Error decommissioning ${this}:`, e);
```
Apollon77: hm .. but did we not said that  "decommission" is the decommissioning which errors when failing and delete is "without? Now you combined that again and catch the error?

packages/protocol/src/protocol/ExchangeManager.ts
```diff
@@ -149,8 +149,8 @@ export class ExchangeManager {
                 if (this.#closing) return;
                 const initiatorNodeId = packet.header.sourceNodeId ?? NodeId.UNSPECIFIED_NODE_ID;
                 session =
-                    this.#sessionManager.getUnsecureSession(initiatorNodeId) ??
-                    this.#sessionManager.createInsecureSession({
+                    this.#sessionManager.getUnsecuredSession(initiatorNodeId) ??
```
Apollon77: Re the other place wehere an Unsecure session is not deleted because the channel is moved to a secureSession ... whats with this then here? Would the unsecure session still be detected here?

packages/protocol/src/interaction/InteractionMessenger.ts
```diff
@@ -316,17 +315,27 @@ export class InteractionServerMessenger extends InteractionMessenger {
...
+            if (error instanceof NoResponseTimeoutError) {
+                logger.info(this.exchange.via, error);
+                return;
+            }
+
+            let errorStatusCode = Status.Failure;
             if (error instanceof StatusResponseError) {
-                logger.info(`Sending status response ${error.code} for interaction error: ${error.message}`);
+                logger.info(
+                    this.exchange.via,
```
Apollon77, line 330: maybe add the direction string as first element like for others here to show that we send?

packages/general/src/util/String.ts
```diff
@@ -326,3 +326,17 @@ export function serializeToJs(value: unknown) {
     parts.push("}");
     return parts.join("");
 }
+
+export namespace hex {
+    export function fixed(value: number | bigint, width: number) {
```
Apollon77, line 116: Maybe then also Diagnostic.hex() can use this?

packages/node/src/node/server/ServerSubscription.ts
```diff
@@ -590,31 +590,32 @@ export class ServerSubscription extends Subscription {
...
+
+        this.#changeHandlers.close();
 
         if (flush) {
             await this.#flush();
         }
+
+        this.#updateTimer.stop();
+        this.#sendDelayTimer.stop();
+        this.isClosed = true;
```
Apollon77: already set in L 610

packages/protocol/src/peer/Peer.ts
Apollon77: yeahh maybe updating makes sense

packages/protocol/src/session/NodeSession.ts
```diff
@@ -47,9 +47,13 @@ export class NodeSession extends SecureSession {
...
     readonly #attestationKey: Bytes;
     #caseAuthenticatedTags: CaseAuthenticatedTag[];
-    #isClosing = false;
     readonly supportsMRP = true;
     readonly type = SessionType.Unicast;
+    readonly #closedByPeer = AsyncObservableValue();
+    #isPeerLost = false;
+
+    // TODO - currently server specific; should move out or be integrated with client
+    readonly #subscriptions = new BasicSet<Subscription>();
```
Apollon77, line 58: Yeahh formally subscriptions are "by peer" and should use "the current active session that peer has - or open a new one if none is there"

packages/protocol/src/session/SessionManager.ts
```diff
@@ -632,30 +646,44 @@ export class SessionManager {
...
+        }
+
         await this.#subscriptionUpdateMutex;
 
-        await this.#storeResumptionRecords();
         const closePromises = this.#sessions.map(async session => {
-            await session?.end(false);
+            await session.closeSubscriptions(true);
+
+            // TODO - some CHIP tests (CASERecovery for one) expect us to exit without closing the session and will fail
```
Apollon77, line 683: i think this is also the "shutdown" case right?  Sending all session closes would potentially lead to a full retransmission cycle before actually shutting down (or factory resetting) the device ... 
And this comment is in general about" why not sending session close on "all close" cases

packages/protocol/src/peer/PeerAddress.ts
```diff
@@ -5,6 +5,7 @@
  */
 
 import { FabricIndex, GroupId, NodeId } from "#types";
+import { DiagnosticPresentation } from "@matter/general";
```
Apollon77: #general? :-)

packages/protocol/src/protocol/ExchangeManager.ts
```diff
@@ -116,31 +122,42 @@ export class ExchangeManager {
...
+            this.#deleteTransport(listener);
+        }
 
         for (const protocol of this.#protocols.values()) {
-            await protocol.close();
+            this.#workers.add(protocol.close());
         }
 
-        await this.#workers;
+        {
```
Apollon77, line 156: Ohhh yeeaaahh ... such intermediate blocks might become handy ... but also strange ... i think I need a bit to get familiar with that. :-)

packages/general/src/environment/RuntimeService.ts
```diff
@@ -4,26 +4,29 @@
...
 import { asError } from "#util/Error.js";
+import { Lifetime } from "#util/Lifetime.js";
 import { Diagnostic } from "../log/Diagnostic.js";
 import { DiagnosticSource } from "../log/DiagnosticSource.js";
 import { Logger } from "../log/Logger.js";
-import { Construction, type Constructable } from "../util/Construction.js";
-import { Cancellable, Destructable, Lifecycle } from "../util/Lifecycle.js";
-import { Multiplex } from "../util/Multiplex.js";
+import { type Constructable } from "../util/Construction.js";
+import { Destructable, Lifecycle } from "../util/Lifecycle.js";
```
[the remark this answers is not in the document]
Apollon77: That should be a false positive because else linting would have complained

Apollon77:
Awesome, especially also the Lifetime tracking sounds interesting. 

Other that that mainly minor comments

Apollon77 (approved):

packages/general/src/util/Observable.ts
lauckhart: Added

packages/protocol/src/protocol/ExchangeManager.ts
lauckhart: Here's alternatives and why I didn't choose them:

```ts
// Downsides: Adds a stack frame, perhaps slightly noisier
await closing.join("workers", async () => {
    await this.#workers;
});

// Downsides: Messy, perhaps more error prone, only works for single promise
const closing = closing.join("workers");
await this.#workers.finally(() => closing.close());

// Downsides: Messy, perhaps more error prone
const closing = closing.join("workers");
try {
    await this.#workers;
} finally {
    closing.close();
}
```

packages/protocol/src/session/SessionManager.ts
lauckhart: In this case we could probably terminate MRP retries early, but in general terminating a session without notifying the peer seems like bad form.  In the CASERecovery case for example the test takes considerably longer because the test does a full retransmission cycle on the old session before giving up and establishing a new session

packages/protocol/src/session/NodeSession.ts
lauckhart: I updated the comment and already stripped almost everything out of Subscription.ts

packages/general/src/util/String.ts
lauckhart: Ah, forgot about Diagnostic.hex...  Tried to move it to this but since the padding is optional there it isn't really compatible

packages/protocol/src/interaction/InteractionMessenger.ts
lauckhart: Good idea on direction string, added.  In general I'd like to leave the "via" as the first element as that puts it in the same place on the lines and its color makes it easy to see the beginning of the rest of the message

packages/protocol/src/protocol/ExchangeManager.ts
lauckhart: Hmm, maybe, yes.  I added an explicit "detach channel" method so it can still be closed without interfering with the channel, and made unsecured session deregister itself when the channel detaches

packages/node/src/node/ClientNode.ts
lauckhart: Hmm, I was thinking that `decommission` decommissions gracefully, while `delete` drops the peer without informing it.  Don't remember considering error handling, but if I said it then it must be right ;) Will remove

packages/node/src/node/server/ServerSubscription.ts
lauckhart: apparently took care of this in subsequent refactoring

packages/general/src/environment/RuntimeService.ts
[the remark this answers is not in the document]
lauckhart: It's because it's only referenced by a comment, so TS views it as "used" but CodeQL does not.  Changed to a type import

packages/protocol/src/session/Session.ts
lauckhart: yeah pretty sure we're all good here

packages/protocol/src/peer/ControllerCommissioner.ts
lauckhart: I also ended up adding a call to explicitly take the channel then added close here in a finally so it's less confusing

packages/protocol/src/session/SessionManager.ts
```diff
@@ -694,16 +694,6 @@ export class SessionManager {
         );
     }
 
-    updateAllSubscriptions() {
```
Apollon77, line 723: Where this was moved to? because should be used

packages/node/src/node/server/ServerSubscription.ts
```diff
@@ -685,7 +685,7 @@ export class ServerSubscription implements Subscription {
         const session = RemoteActorContext({
             activity: (exchange as NodeActivity.WithActivity)[NodeActivity.activityKey],
             fabricFiltered: request.isFabricFiltered,
-            message: {} as Message,
+            message: { packetHeader: { messageId: 123 } } as Message,
```
Apollon77: test leftover?

packages/protocol/src/session/SessionManager.ts
Apollon77: Yeah, but "thats the reality right now". But we can try to send , more close infos and see how stuff behaves. Open for "being better" :-)

Apollon77: And as said: if this would just confluct with "non TC_*" tests then we can also skip them knowing why

packages/protocol/src/protocol/ExchangeManager.ts
Apollon77: Makes sense.

packages/protocol/src/session/SessionManager.ts
lauckhart: On main it's not used and the update() call that it invokes on Subscription is empty

packages/node/src/node/server/ServerSubscription.ts
lauckhart: no I actually thought this was test code when I edited

PR #2740, opened 2025-11-20
Introduce DclVendorInfoService

Apollon77:
... and some naming cleanups on Certificate service. The new DclVendorInfoService fetches all VendorIds from the Dcl and provides central access to them.

### DIFF
--- a/packages/protocol/src/dcl/DclVendorInfoService.ts
+++ b/packages/protocol/src/dcl/DclVendorInfoService.ts
@@ -190,7 +190,7 @@ export class DclVendorInfoService {
     /**
      * Close the service and stop periodic updates.
      */
-    close() {
+    async close() {
         if (this.#closed) return;
         this.#closed = true;
 


packages/protocol/src/dcl/DclClient.ts
```diff
@@ -40,6 +41,35 @@ export class DclClient {
         this.#baseUrl = this.production ? DCL_PRODUCTION_URL : DCL_TEST_URL;
     }
 
+    async #fetchPaginatedJson<ItemT>(
+        path: string,
+        paginatedField: string,
+        options?: DclClient.Options,
+    ): Promise<ItemT[]> {
```
lauckhart, line 48: Maybe an async generator here?

lauckhart (approved):

packages/protocol/src/dcl/DclClient.ts
Apollon77: I also thought about it but for now we always need all data andn it not that many. So the "loop" is easier in a central place than in potentially multiple

PR #2741, opened 2025-11-21
Add shared Environment services with tracked participants

Apollon77:
More details see the added comments in Environment.ts For shared services we now allow registering them by providing participant references and these will be tracked and the service will close when the last participant closes.


packages/general/src/environment/Environment.ts
```diff
@@ -134,13 +233,24 @@ export class Environment {
...
+        const participants = instance !== undefined ? this.participants(type) : undefined;
+        this.delete(type, instance, participant); // delete and block inheritance
         if (instance !== undefined) {
+            if (participant !== undefined && participants !== undefined && participants !== null) {
+                participants.delete(participant);
+                if (participants.size > 0) {
+                    // still in use
+                    return;
+                }
+            }
```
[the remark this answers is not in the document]
Apollon77: Correct, but I change differently

packages/matter.js/src/PaseCommissioner.ts
```diff
@@ -79,7 +79,7 @@ export class PaseCommissioner {
         const { certificateAuthorityConfig: rootCertificateData, fabricConfig: fabricConfig } = this.options;
 
         try {
-            await this.environment.load(MdnsService);
+            await this.environment.load(MdnsService, this);
         } catch {
             logger.debug("No networking available, using only BLE");
         }
```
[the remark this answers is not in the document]
Apollon77: Thats ok, if the particpnat. is not registered a close will do nothing

packages/general/src/environment/Environment.ts
```diff
@@ -90,32 +140,81 @@ export class Environment {
...
             return instance;
         }
 
         throw new UnsupportedDependencyError(`Required dependency ${type.name}`, "is not available");
     }
 
+    participants(type: Environmental.ServiceType): Set<any> | undefined | null {
+        const { instance: mine, participants } = this.#services?.get(type) ?? {};
+        if (mine !== undefined) {
+            return participants instanceof Set ? participants : undefined;
```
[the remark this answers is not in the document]
Apollon77: adjusted the signature

packages/general/src/environment/Environment.ts
```diff
@@ -30,10 +30,33 @@ const logger = Logger.get("Environment");
...
+    #services?: Map<
+        Environmental.ServiceType,
+        {
+            /** The instance of the service, null when deleted to block from parents inheritance */
+            instance: Environmental.Service | null;
+
+            /**
+             * Set of participants using the service, undefined when not yet accessed, null when accessed without participants
+             */
+            participants?: Set<any> | null;
```
lauckhart: Dependents might be a better name. And maybe {} instead of any

packages/general/src/environment/Environment.ts
```diff
@@ -134,13 +233,23 @@ export class Environment {
 
     /**
      * Remove and close an environmental service.
+     * Ensure to call close with the same participant as used in get to properly track usage.
```
lauckhart: It feels like we are using an "any" as a fairly fragile key here...  I'm thinking maybe we should have something like this:

```ts
const services = environment.asDependent();

const mdns = services.get(MdnsService);
// etc.

// do stuff with services

services.close();

// now services are no longer valid
```

...so the way you indicate you are dependent on a service is also the way you indicate you are no longer dependent on the service.  And you wouldn't need to free dependencies one-by-one on close

Also the interface returned by `asDependent` could be shared by Environment so the method access wouldn't need to change.  You kind of would get dependency tracking for free

packages/general/src/environment/Environment.ts
```diff
@@ -65,19 +88,46 @@ export class Environment {
...
+                    `Service ${typeName} was initialized ${serviceUsedWithParticipants ? "with" : "without"} participants but is being accessed ${hasParticipant ? "with" : "without"} participants`,
+                );
+            }
+        }
+    }
+
     /**
      * Access an environmental service.
+     * Optionally track usage by a participant, which delays closing till all participants are gone.
+     * @param participant optional participant requesting the service, if provided the service tracks usage by the participant
```
lauckhart: Why is it necessary for this to be modal?  If environment knows there is a dependency then could it just ignore close calls without a dependency?  And conversely, if I ask for a service already available in "untracked" mode could we just add a dependency at that point?

packages/general/src/environment/Environment.ts
```diff
@@ -90,32 +140,81 @@ export class Environment {
...
+        // Validate participant mode consistency
+        const hasParticipant = participant !== undefined;
+        if (localInstance !== undefined && localInstance !== null) {
+            // We have a local instance, validate participant mode consistency
+            this.#assertParticipantUsage(hasParticipant, participants, type.name);
+
+            if (hasParticipant) {
+                // Remove the participant and only delete if no participants remain
+                serviceData?.participants?.delete(participant);
+                if (
```
lauckhart: Curious if `if (serviceData?.participants?.size)` seems unreasonable here, personally would be an easier read

packages/node/src/behavior/system/network/ServerNetworkRuntime.ts
```diff
@@ -74,7 +74,7 @@ export class ServerNetworkRuntime extends NetworkRuntime {
             const port = this.owner.state.network.operationalPort;
             const options = this.owner.state.commissioning.mdns;
             const crypto = this.owner.env.get(Crypto);
-            const { server } = this.owner.env.get(MdnsService);
+            const { server } = this.owner.env.get(MdnsService, this.owner);
```
lauckhart: Hmm with idea r.e. "asDependent" you could have the "services" be the thing you request services from, then you don't need to figure out what to use as the key

packages/nodejs-shell/src/MatterNode.ts
```diff
@@ -114,6 +115,15 @@ export class MatterNode {
 
     async close() {
         await this.commissioningController?.close();
+        if (this.environment.has(DclOtaUpdateService)) {
```
lauckhart: Yeah `this.services.close()` seems like a nice upgrade over this block.  Kind of like closing an ObserverGroup vs unregistering individual observers

packages/general/src/environment/Environment.ts
Apollon77: Adjusted along this idea, please re-review

packages/general/src/environment/SharedEnvironmentServices.ts
```diff
@@ -0,0 +1,141 @@
...
+ *
+ * Tracks which services are in use and ensures services are only closed when all
+ * consumers have released them. This enables safe sharing of services across multiple
+ * consumers without premature cleanup.
+ *
+ * Created via {@link Environment.asDependent} and automatically registered at the root
+ * environment for centralized lifecycle management. All shared services are accessed
+ * at the root level regardless of which environment creates the instance.
+ */
+export class SharedEnvironmentServices {
```
lauckhart: Maybe add ServiceProvider interface here to ensure that subset of API matches `Environment`

packages/general/test/environment/EnvironmentTest.ts
```diff
@@ -0,0 +1,999 @@
...
+            const fromChild1 = child.get(TestService);
+            const fromParent2 = parent.get(TestService);
+            const fromChild2 = child.get(TestService);
+
+            // All should return the exact same instance
+            expect(fromParent1).to.equal(service);
+            expect(fromChild1).to.equal(service);
+            expect(fromParent2).to.equal(service);
+            expect(fromChild2).to.equal(service);
+            expect(fromParent1).to.equal(fromChild1);
```
lauckhart: These checks are redundant

packages/general/test/environment/EnvironmentTest.ts
```diff
@@ -0,0 +1,999 @@
...
+        this.closed = true;
+    }
+}
+
+describe("Environment", () => {
+    describe("basic service management", () => {
+        it("creates and retrieves a service", () => {
+            const env = new Environment("test");
+            const service = new TestService();
+            env.set(TestService, service);
```
lauckhart, line 55: These three lines repeat a lot.  Maybe put env and service in variables set up by a before each hook.  The few tests that need a different pattern can just ignore

packages/general/test/environment/EnvironmentTest.ts
```diff
@@ -0,0 +1,999 @@
...
+
+            env.close(TestService);
+
+            expect(service.closed).to.be.true;
+            expect(env.has(TestService)).to.be.false;
+        });
+    });
+
+    describe("dependent tracking - get()", () => {
```
lauckhart: Split this into nested describes, one for "dependent tracking" and "get" nested inside...  Then nest other "dependent tracking" blocks too

Then move setup of things like dependent1, dependent2 etc. into scope variables maintained by hooks within the "dependent tracking" or "get" blocks depending on how broadly they are shared

packages/general/test/environment/EnvironmentTest.ts
```diff
@@ -0,0 +1,999 @@
...
+            const service = new TestService();
+            parent.set(TestService, service);
+
+            const fromParent1 = parent.get(TestService);
+            const fromChild1 = child.get(TestService);
+            const fromParent2 = parent.get(TestService);
+            const fromChild2 = child.get(TestService);
+
+            // All should return the exact same instance
+            expect(fromParent1).to.equal(service);
```
lauckhart: How about an `expectAllEqual` utility?  Then `expectAllEqual(service, fromParent1, fromparent2, fromChild1, fromChild2)`

packages/general/test/environment/EnvironmentTest.ts
lauckhart, line 1: I paid more attention here than usual to tests because copied code tends to make future refactoring a PITA

packages/node/src/behaviors/general-diagnostics/GeneralDiagnosticsServer.ts
```diff
@@ -327,8 +327,10 @@ export class GeneralDiagnosticsServer extends Base {
     }
 
     async #updateNetworkList() {
-        const mdnsService = this.env.get(MdnsService);
+        const services = this.env.asDependent();
```
lauckhart: Ah so with this usage pattern we should add a `[Symbol.dispose]` so you can "using" it

lauckhart (commented):
Cool looks great

packages/general/src/environment/SharedEnvironmentServices.ts
Apollon77: Good idea, done

packages/node/src/behaviors/general-diagnostics/GeneralDiagnosticsServer.ts
Apollon77: added

PR #2757, opened 2025-11-24
Enhance Group resource validations

Apollon77:
The Group Key management cluster can contain groups in either groupTable and groupKeyMap. Ideally these are in sync, but formally not needed. This change adds missing resource validation.

This was found in Cert test G-2.2 which is a manual test

### DIFF
--- a/packages/node/src/behaviors/group-key-management/GroupKeyManagementServer.ts
+++ b/packages/node/src/behaviors/group-key-management/GroupKeyManagementServer.ts
@@ -179,6 +179,16 @@ export class GroupKeyManagementServer extends GroupKeyManagementBehavior {
                 );
             }*/
         }
+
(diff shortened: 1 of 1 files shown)


packages/node/src/behaviors/group-key-management/GroupKeyManagementServer.ts
```diff
@@ -179,6 +179,16 @@ export class GroupKeyManagementServer extends GroupKeyManagementBehavior {
                 );
             }*/
         }
+
+        // We also need to add the groups from the group table to the count of groups per fabric
+        for (const entry of this.state.groupTable) {
```
lauckhart, line 184: Would be nice to add a GroupKeyManagementServerTest for this...

lauckhart (approved):

Apollon77:
Logic was correct,

PR #2763, opened 2025-11-25
Client Node Logic enhancements

Apollon77:
This change includes
* Ensure that events are processed when reading "read" results after the attributes, also ensure that clusters are created before emitting them on first read.
* Use an "DataUpdate" mutex for data update syncing on a node to not conflict with the lifecycle "startup" mutex
* log attribute/event status entries at least
* stop client nodes together with the  servernode lifecycle - start commented out as discussed with a TODO
* track when a clientNode startup is in progress to not try to start the device again for the first read (because that would run into a block)
* Adjust erase/delete methods to decommission/delete because erase is also used internally
* Fix Read chunking logic by cutting potentially chunked arrays and decode in next chunk run


packages/node/src/node/client/ClientStructure.ts
```diff
@@ -27,7 +27,7 @@
 import { ReadScope, type Read, type ReadResult } from "#protocol";
 import { DatasourceCache } from "#storage/client/DatasourceCache.js";
 import { ClientNodeStore } from "#storage/index.js";
-import type { AttributeId, ClusterId, ClusterType, CommandId, EndpointNumber } from "#types";
+import { AttributeId, ClusterId, ClusterType, CommandId, EndpointNumber, Status } from "#types";
```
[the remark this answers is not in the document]
Apollon77: I fix differently

packages/node/src/node/client/ClientStructure.ts
```diff
@@ -297,30 +337,43 @@ export class ClientStructure {
...
+                    cluster.behavior = undefined;
+                }
             }
-        }
 
-        if (cluster.behavior && AttributeList.id in attrs.values) {
-            if (!isDeepEqual(cluster.attributes, attrs.values[AttributeList.id])) {
-                cluster.behavior = undefined;
+            if (cluster.behavior && AcceptedCommandList.id in updates.values) {
+                if (!isDeepEqual(cluster.attributes, updates.values[AttributeList.id])) {
```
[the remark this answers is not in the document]
Apollon77: Ohh thats a nice finding ... @lauckhart I fixed it too

packages/node/src/node/ClientNode.ts
```diff
@@ -70,6 +72,10 @@ export class ClientNode extends Node<ClientNode.RootEndpoint> {
         return this.#matter;
     }
 
+    get dataUpdateMutex() {
```
lauckhart: Does this need to be on ClientNode?  We should generally try to avoid polluting the public interface

packages/node/src/node/ClientNode.ts
```diff
@@ -110,22 +116,42 @@ export class ClientNode extends Node<ClientNode.RootEndpoint> {
...
             this.statusUpdate("decommissioning");
 
             await this.act("decommission", agent => agent.commissioning.decommission());
         }
+        await this.delete();
+    }
 
-        await super.delete();
+    override async start() {
+        if (this.#startInProgress) {
```
lauckhart: Skylar this be in Node instead?  It doesn't really seem like it's ClientNode specific.  It basically protects from the beginning of the call to the point where #runtime is set

packages/node/src/node/client/ClientStructure.ts
```diff
@@ -168,10 +170,10 @@ export class ClientStructure {
      */
     async *mutate(request: Read, changes: ReadResult) {
         // Ensure mutations run serially and integrate properly with node lifecycle
-        using _lock = await this.#node.lifecycle.mutex.lock();
+        using _lock = await this.#node.dataUpdateMutex.lock();
```
lauckhart: Are we protected against lifecycle transitions conflicting with structure updates?  I can't think of where it might interfere but want to make sure. An alternative might be to avoid grabbing mutex if we're initializing when we know it's already held

packages/node/src/node/client/ClientStructure.ts
```diff
@@ -263,16 +273,47 @@ export class ClientStructure {
 
             // Update version but only if this was a wildcard read
             if (scope.isWildcard(endpointId, clusterId)) {
-                currentUpdates.values[DatasourceCache.VERSION_KEY] = change.version;
+                currentUpdates.values![DatasourceCache.VERSION_KEY] = change.version;
```
lauckhart: The "!" assertion is pretty far from where this is set. Maybe test above could be a type assertion so reasoning is more apparent and we don't accidentally break it?  Or just add an if to confirm it's really there

packages/node/src/node/client/ClientStructure.ts
```diff
@@ -263,16 +273,47 @@ export class ClientStructure {
...
 
         return currentUpdates;
     }
 
+    async #emitEvent(occurrence: ReadResult.EventValue, currentUpdates?: ClusterUpdates) {
+        const { endpointId, clusterId } = occurrence.path;
+
+        // If we are building updates to a cluster and the cluster/endpoint changes, apply the current update
+        // set
+        if (currentUpdates && (currentUpdates.endpointId !== endpointId || currentUpdates.clusterId !== clusterId)) {
```
lauckhart: Do events really come in in cluster order??  I assumed it would be chronological.  It seems like this logic might make assumptions about the implementation of the peer that might not be necessary. We could instead just queue events when the cluster is not initialize yet

packages/node/src/node/client/ClientStructure.ts
```diff
@@ -747,15 +821,19 @@ export class ClientStructure {
...
+        }
     }
 }
 
-interface AttributeUpdates {
+interface ClusterUpdates {
     endpointId: EndpointNumber;
     clusterId: ClusterId;
-    values: {
+    values?: {
```
lauckhart: Maybe rename to attributes since it's not in the container name anymore

packages/node/test/node/ClientNodeTest.ts
```diff
@@ -92,6 +92,7 @@ describe("ClientNode", () => {
         expect(ep1.state).deep.equals(EP1_STATE);
         const expectedEp1State = deepCopy(ep1.state);
 
+        //await peer1.set({ network: { autoSubscribe: false } });
```
lauckhart: What's this?

packages/node/test/node/ClientNodeTest.ts
lauckhart, line 1: Is there any new testing added to ensure susceptibility come up cleanly on restart?  If not that's likely to break again

packages/protocol/src/interaction/AttributeDataDecoder.ts
```diff
@@ -62,10 +63,41 @@ export type StructuredReadAttributeData = { [key: number]: { [key: number]: { [k
...
  */
-export function normalizeAndDecodeReadAttributeReport(data: TypeFromSchema<typeof TlvAttributeReport>[]): {
+export function normalizeAndDecodeReadAttributeReport(
+    data: TypeFromSchema<typeof TlvAttributeReport>[],
+    leftoverAttributeReports?: TypeFromSchema<typeof TlvAttributeReport>[],
+): {
     attributeData: DecodedAttributeReportValue<any>[];
     attributeStatus: DecodedAttributeReportStatus[];
 } {
+    // TODO Simplify this again into reach streaming once old Client API is gone
```
lauckhart: I don't know what this means, maybe some more detail on what needs to happen?

lauckhart (approved):

packages/node/src/node/ClientNode.ts
Apollon77: I can move to ClientStructure

packages/node/src/node/client/ClientStructure.ts
Apollon77: The problem was that this lifecycle mutex is used to protect the startup, so the initial "read/subscribe" done "during "start lead to  a blocking behavior. So your idea was to separate it into an extra mutex :-)

packages/node/src/node/ClientNode.ts
Apollon77: moved

packages/node/src/node/client/ClientStructure.ts
Apollon77: I had a bit more thoughts. Why do we need to lock here? Because basically if there is really a normal read interfering with a running subscription that could have several side effects. Should we maybe move that to the place where we install/rebuild clusters to ensure this does not overlap? Here is blocks the whole read (multi chunk wise)

packages/node/src/node/client/ClientStructure.ts
Apollon77: No it is directly som4 live above now ... so should be fine

packages/node/src/node/client/ClientStructure.ts
lauckhart: > The problem was that this lifecycle mutex is used to protect the startup, so the initial "read/subscribe" done "during "start lead to a blocking behavior. So your idea was to separate it into an extra mutex :-)

Yes, we just need to make sure that ClientStructure is closed or whatever before the node tears down its structure

In the case of a mutex calling "close" on it will ensure nobody is or will be using it

lauckhart: > I had a bit more thoughts. Why do we need to lock here? Because basically if there is really a normal read interfering with a running subscription that could have several side effects. Should we maybe move that to the place where we install/rebuild clusters to ensure this does not overlap? Here is blocks the whole read (multi chunk wise)

Yeah this is mainly protecting against edge cases that are hard to reason about.  Maybe we should just kill the lock altogether and focus on places where we can identify actual race conditions

packages/node/src/node/client/ClientStructure.ts
lauckhart: Yeah looks fine, I missed it

packages/node/src/node/client/ClientStructure.ts
Apollon77: Ok I removed the mutex for now and lets see when we get issues

packages/node/test/node/ClientNodeTest.ts
Apollon77: a leftover, removed

packages/node/test/node/ClientNodeTest.ts
Apollon77: The one old testcase verified already that the node, when re-instanciated, has the right data without a start. So this test validates this.

Because there is no automatic start I also did not added a test for this, but the one test that tests the restart will fail as soon as we start trying to connect, and if we add that we need to add tests for it, so for now i guess we are covered with what the functionality supports

packages/node/src/node/client/ClientStructure.ts
Apollon77: likely not ... right .. I adjusted

packages/node/test/node/ClientNodeTest.ts
Apollon77: But I added a test todo

packages/node/src/node/client/ClientStructure.ts
Apollon77: Renamed back

PR #2768, opened 2025-11-26
Automatically convert epoch-s/us values

### LINKED ISSUE #2767: Link in support/{codegen,models}/README.md is broken
In [support/codegen/README.md](https://github.com/matter-js/matter.js/blob/main/support/codegen/README.md) and [support/models/README.md](https://github.com/matter-js/matter.js/blob/main/support/models/README.md) there is a link to [support/packages/model/README.md](https://github.com/matter-js/matter.js/blob/main/support/packages/model/README.md), but it leads to a 404. I could not found which file it was suppose to link to.

I was interested in finding out what is the source used for code generation.

Apollon77:
Matter uses "1.1.2000"-based epoch values in some cases, we now  convert these automatically. There is no need to do anything. Just continue to use 1970-based epochs.

also fixes #2767

### DIFF
--- a/packages/types/src/tlv/TlvNumber.ts
+++ b/packages/types/src/tlv/TlvNumber.ts
@@ -18,7 +18,7 @@ import {
     INT8_MAX,
     INT8_MIN,
     ImplementationError,
-    Timestamp,
+    Seconds,
     UINT16_MAX,
     UINT24_MAX,
     UINT32_MAX,
@@ -235,13 +235,14 @@ export const TlvPosixMs = TlvUInt64;
 export const TlvSysTimeUs = TlvUInt64;
 export const TlvSysTimeMS = TlvUInt64;
 
-/** Seconds from Unix epoch (1970-01-01) to Matter epoch (2000-01-01) */
-export const MATTER_EPOCH_OFFSET_S = Timestamp(Days(10_957));
+/** Milliseconds from Unix epoch (1970-01-01) to Matter epoch (2000-01-01) */
+export const MATTER_EPOCH_OFFSET = Days(10_957);
 
-/**
- * Microseconds from Unix epoch (1970-01-01) to Matter epoch (2000-01-01)
- */
-export const MATTER_EPOCH_OFFSET_US = BigInt(MATTER_EPOCH_OFFSET_S * 1_000_000);
+/** Seconds from Unix epoch (1970-01-01) to Matter epoch (2000-01-01) */
+export const MATTER_EPOCH_OFFSET_S = Seconds.of(MATTER_EPOCH_OFFSET);
+
+/** Microseconds from Unix epoch (1970-01-01) to Matter epoch (2000-01-01) */
+export const MATTER_EPOCH_OFFSET_US = BigInt(MATTER_EPOCH_OFFSET * 1_000);
 
 /**
  * TLV Schema for Epoch time in seconds since Matter epoch (2000-01-01). You can just use the normal unix epoch


lauckhart (approved):

packages/node/src/behaviors/switch/SwitchServer.ts
```diff
@@ -30,13 +30,13 @@ const schema = SwitchServerBase.schema.extend({
     children: [
         FieldElement({
             name: "longPressDelay",
-            type: "epoch-s",
+            type: "uint32",
```
lauckhart: Maybe these should be "duration"?

packages/node/src/behavior/state/validation/ValueValidator.ts
```diff
@@ -221,11 +221,22 @@ function createSimpleValidator(
...
 
 function createIntegerValidator(schema: ValueModel, supervisor: RootSupervisor, name: string) {
+    const nullable = schema.effectiveQuality.nullable;
     let assertion;
-    if (schema.effectiveQuality.nullable) {
-        assertion = assertInt.nullable[name];
-    } else {
-        assertion = assertInt.notNullable[name];
+    // Let's check schema type specific assertations first
+    if (schema.type === "epoch-s" || schema.type === "epoch-us") {
```
lauckhart, line 228: Ideally these would probably be "date" metatype instead of integer which would change to entirely different visitation.  Probably too involved though because date handling would then be more modifications...  Maybe could just leave a TODO

packages/node/src/behavior/state/validation/assertions.ts
```diff
@@ -94,6 +95,30 @@ for (let i = 1n; i < 9n; i++) {
...
+    `nullable epoch-s`,
+    BigInt(MATTER_EPOCH_OFFSET_S),
+    BigInt(UINT32_MAX + MATTER_EPOCH_OFFSET_S) - 1n,
+);
+
+// Add special epoch-us case
+assertInt.notNullable["epoch-us"] = createIntAssertion(
+    "epoch-us",
+    MATTER_EPOCH_OFFSET_US,
+    UINT64_MAX + MATTER_EPOCH_OFFSET_US,
```
lauckhart, line 115: Ideally we'd encode as a ms float value which is how we handle high-precision time everywhere else...  Then convert for wire same as we do for epoch offset.  Then would only need one epoch

Not sure if loss of prison would be an issue though...  And probably not practical without metatype change though, so maybe just a TODO for that somewhere too

packages/types/src/tlv/TlvNumber.ts
```diff
@@ -228,8 +229,58 @@ export const TlvPercent = TlvUInt8.bound({ max: 100 });
...
 
 // Time Number types
-export const TlvEpochUs = TlvUInt64;
-export const TlvEpochS = TlvUInt32;
 export const TlvPosixMs = TlvUInt64;
 export const TlvSysTimeUs = TlvUInt64;
 export const TlvSysTimeMS = TlvUInt64;
+
+/** Seconds from Unix epoch (1970-01-01) to Matter epoch (2000-01-01) */
+export const MATTER_EPOCH_OFFSET_S = 10957 * 24 * 60 * 60;
```
lauckhart: Could this be a Timestamp instead?  So something like `MATTER_EPOCH_OFFSET = Timestamp(Days(10_957))`?

packages/node/src/behaviors/switch/SwitchServer.ts
Apollon77: right. changed

PR #2813, opened 2025-12-09
Add support for intermediate certificate

ArtemisMucaj:
Hey,

For testing purposes I need to use an ICAC when commissioning with matter.js  however it isn't supported, there's a TODO in the CertificateAuthority.ts file.

This PR adds support for it.

* FabricAuthority passes ICAC when available during fabric creation
* ControllerCommissioningFlow also includes ICAC in AddNOC command
* LegacyControllerCommandHandler returns ICAC in certificate responsees

I added a few tests to cover the PR. If you want to discuss the contents of the MR or want me to adjust the way I've implemented the feature I can do so :)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -95,6 +95,7 @@ The main work (all changes without a GitHub username in brackets in the below li
     - Fix: Ensures to only include the MaxTcpMessageSize in Session parameters when TCP is enabled
     - Fix: Fixes the used ACL level for wildcard writes
     - Fix: (@ArtemisMucaj) Fixes noc trust chain verification; verify against both rcac and icac
+    - Feature: (@ArtemisMucaj) Add support to pass / generate an intermediate CA in the CertificateAuthority class, when used the NOCs are signed with the ICAC cert.
 
 - @matter/react-native
     - Fix: (Luxni) Update UDP, BLE and Crypto usage to work with React Native


packages/protocol/src/certificate/CertificateAuthority.ts
```diff
@@ -22,14 +22,31 @@ import {
...
 /**
  * Manages the root key pair for a fabric owned by a local node.
- * TODO: Add support for (optional) ICACs
+ *
+ * Supports optional Intermediate Certificate Authority (ICAC) for 3-tier PKI hierarchy.
+ * When ICAC is enabled, the certificate chain becomes: RCAC -> ICAC -> NOC instead of RCAC -> NOC.
+ *
+ * Configuration:
+ * - ica: Enable/disable ICAC generation. Defaults to false (2-tier PKI).
+ * - signingCert: Which certificate signs NOCs. Must be "ica" when ica=true.
```
Apollon77: I know that your CA does it that way but do we really need to support that? It is very uncommon and honestly if there is an ICAC in my eyes it should be used. I only see theoretical cases to not use it but to provide it.

packages/protocol/src/certificate/CertificateAuthority.ts
```diff
@@ -39,6 +56,9 @@ export class CertificateAuthority {
     #rootCertBytes?: Bytes;
     #nextCertificateId = BigInt(1);
     #construction: Construction<CertificateAuthority>;
+    #ica!: boolean;
```
Apollon77: Why do you needed to enfore existence?  Better make option and  add this.construction.assert() before using it because then better to enforce when using or returning it.

packages/protocol/src/certificate/CertificateAuthority.ts
```diff
@@ -187,14 +267,155 @@ export class CertificateAuthority {
...
+    };
+
     export type Configuration = {
         rootCertId: bigint;
         rootKeyPair: BinaryKeyPair;
         rootKeyIdentifier: Bytes;
         rootCertBytes: Bytes;
         nextCertificateId: bigint;
+        ica?: boolean;
+        signingCert?: SigningCertType | string;
```
Apollon77: when handed in as configuration I would only use the enum type because this is then done by code, or?!

packages/protocol/src/certificate/CertificateAuthority.ts
```diff
@@ -187,14 +267,155 @@ export class CertificateAuthority {
...
+            return value as SigningCertType;
+        }
+        throw new InternalError(
+            `Invalid signingCert value: "${value}". Must be "${SigningCertType.Root}" or "${SigningCertType.Intermediate}"`,
+        );
+    }
+
+    #validateConfiguration(): void {
+        if (this.#signingCert === SigningCertType.Intermediate && !this.#ica) {
+            throw new InternalError(
```
Apollon77: InternalError or better ImplementationError? 

meeaning difference: Internal means "matter.js developers need to check that because something more internal broke" vs ImplementationError mans "dear developer, check the data you used because likely it is something about how you used the system which is wrong" ...

packages/protocol/src/certificate/CertificateAuthority.ts
```diff
@@ -187,14 +267,155 @@ export class CertificateAuthority {
...
+        if (this.#isValidStoredIcacCertificate(certValues)) {
+            this.#icacState = {
+                certId: BigInt(certValues.icacCertId as bigint | number),
+                keyPair: PrivateKey(certValues.icacKeyPair as BinaryKeyPair),
+                keyIdentifier: certValues.icacKeyIdentifier as Bytes,
+                certBytes: certValues.icacCertBytes as Bytes,
+            };
+
+            logger.info("ICAC data detected in storage and loaded.");
+        } else if (this.#ica) {
```
Apollon77: should this case not also reset the #ica flag?

packages/protocol/src/certificate/CertificateAuthority.ts
ArtemisMucaj: True, IMO it doesn't make sense neither so I can adjust to sign with the ICA when one exists if it's fine to you :) I wanted to leave the door open after what we discussed last time (that it's allowed),

Apollon77: That the question ... If a developer provides stuff then in my eyes I would say he provides whats needed and not anything. But right validation needs to accept both - as your PR yesterday. So for a a CertAuthority where the developer controls what he uses there I would say "we  take whats needed" and if an icac is there we use it ... if not then root. makes sense?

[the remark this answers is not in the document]
ArtemisMucaj: Fixed, the signingCert option was removed.

packages/protocol/src/certificate/CertificateAuthority.ts
```diff
@@ -187,14 +267,155 @@ export class CertificateAuthority {
...
+
+        if (this.#icacState) {
+            data.icacCertId = this.#icacState.certId;
+            data.icacKeyPair = this.#icacState.keyPair.keyPair;
+            data.icacKeyIdentifier = this.#icacState.keyIdentifier;
+            data.icacCertBytes = this.#icacState.certBytes;
+        }
+
+        return data;
+    }
```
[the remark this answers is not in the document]
ArtemisMucaj: Fixed, option is persisted in storage.

packages/protocol/src/certificate/CertificateAuthority.ts
ArtemisMucaj: Fixed, I removed the signingCert option, if ICAC exists we sign NOCs with it.

packages/protocol/src/certificate/CertificateAuthority.ts
ArtemisMucaj: Fixed, option was removed.

packages/protocol/src/certificate/CertificateAuthority.ts
ArtemisMucaj: Fixed, option removed.

packages/protocol/src/certificate/CertificateAuthority.ts
ArtemisMucaj: Fixed :)

packages/protocol/src/certificate/CertificateAuthority.ts
ArtemisMucaj: Fixed, I reset the flag after detecting if present in storage now.

packages/protocol/src/certificate/CertificateAuthority.ts
```diff
@@ -48,49 +65,64 @@ export class CertificateAuthority {
...
 
+            if (this.#isValidStoredRootCertificate(certValues)) {
+                const hasIcacInStorage = this.#isValidStoredIcacCertificate(certValues);
+                if (hasIcacInStorage) {
+                    this.#intermediateCert = true;
+                } else if (typeof certValues.intermediateCert === "boolean") {
+                    this.#intermediateCert = certValues.intermediateCert;
+                } else {
+                    this.#intermediateCert = false;
+                }
```
ArtemisMucaj: I think it'd be best to throw an ImplementationError here if intermediateCert from storage is true but ICAC props are missing since in this code path we're not reading from config

packages/protocol/src/certificate/CertificateAuthority.ts
```diff
@@ -48,49 +65,64 @@ export class CertificateAuthority {
...
+
+                this.#loadFromStorage(certValues);
+
+                // Reset intermediateCert to false if set but no ICAC data exists
+                if (this.#intermediateCert && !hasIcacInStorage) {
+                    logger.warn(
+                        "intermediateCert=true in configuration but no valid ICAC data in storage. Resetting to 2-tier PKI.",
+                    );
+                    this.#intermediateCert = false;
+                }
```
ArtemisMucaj: If an error is thrown above this can be removed IMO

packages/protocol/src/certificate/CertificateAuthority.ts
Apollon77: yes makes sense

packages/protocol/src/certificate/CertificateAuthority.ts
Apollon77: makes sense too

packages/protocol/src/certificate/CertificateAuthority.ts
```diff
@@ -187,14 +275,126 @@ export class CertificateAuthority {
...
+        intermediateCert?: boolean;
+    };
+
     export type Configuration = {
         rootCertId: bigint;
         rootKeyPair: BinaryKeyPair;
         rootKeyIdentifier: Bytes;
         rootCertBytes: Bytes;
         nextCertificateId: bigint;
+        intermediateCert?: boolean;
```
Apollon77, line 379: Do we need this in this case because we have it above in the "Partial" variant ... and when user provides certs then he either provides the icacs or not ... we will not create an icac if he would set that to "true" and did not provided it

packages/protocol/src/certificate/CertificateAuthority.ts
```diff
@@ -187,14 +275,126 @@ export class CertificateAuthority {
...
+
+interface IcacProps {
+    certId: bigint;
+    keyPair: PrivateKey;
+    keyIdentifier: Bytes;
+    certBytes: Bytes;
 }
 
 export namespace CertificateAuthority {
+    export type PartialConfiguration = {
```
Apollon77: Maybe rename to "CreateConfiguration" or "ConfigurationForCreate" because this is only relevant when we create it

support/chip-testing/src/handler/LegacyControllerCommandHandler.ts
```diff
@@ -549,6 +549,15 @@ export class LegacyControllerCommandHandler extends CommandHandler {
             certSigningRequest,
         );
 
+        if (ca.icacCert !== undefined) {
```
Apollon77: I think you can combine that in the below return by just returning "ICAC: ca.icacCert" which becomes undefined and so should be still "not there"

Apollon77 (commented):
Thanks a lot for your support and adjustments. I added some last comments. It really looks great!

Please also add and entry to the changelog that the CertAuthority now supports provided ICACs when providing device or controller certificates. ANd add your name as I did for your other change if you like.

Also if you like join the discord if you have other questions or want to discuss use cases. It would be also great to get some feedback how easy (or hard?) it was to do changes in our code base and what/if we can improve.

packages/protocol/src/certificate/CertificateAuthority.ts
ArtemisMucaj: The configuration flag is also in storage (as asked by a copilot prior review) and loaded so yes it's needed here. 

It felt odd to have in the **create** method's param two completely different objects when receiving the Configuration from store and when asking the class to generate (CreationConfiguration -> with intermediateCert boolean field).

ArtemisMucaj: I'll resolve this for now, if you want me to change something feel free to re open :)

Apollon77: All good. I might look at that when I have it in my IDE in front of me :-)

PR #2816, opened 2025-12-10
Test + updates for unknown client clusters

lauckhart:


packages/node/src/endpoint/Agent.ts
```diff
@@ -79,7 +79,7 @@ export class Agent {
      * Property access is available in TypeScript when the set of behaviors is defined statically.
      */
     get<T extends Behavior.Type>(type: T) {
-        let behavior = this.#behaviors[type.id];
+        let behavior = this.#behaviors[typeof type === "string" ? type : type.id];
```
Apollon77: can type be a string? The call signature does not allow that ... do we also need to enhance that?

lauckhart: Oh, I started to add support then changed my mind, forgot to revert

PR #2819, opened 2025-12-10
Prepare CI adjustments for upcoming Testrunner changes

Apollon77:
When https://github.com/project-chip/connectedhomeip/pull/42321 lands we need to adjust how we run tests in CI too

The run will fail before the linked PR lands


enkiusz:
The testrunner changes PR has been merged.

PR #2820, opened 2025-12-10
Subscription and Exchange adjustments + maxEventNumber for NetworkClient and some more

Apollon77:
This PR contains several fixes and adjustments around Subscriptions and Exchanges. key things are:
* In ClientSubscriptions: when an initial subscription datareport was empty we only consumed the initial report but did not read further and so the initial update was never sending the statusResponse, so device thought we are dead and so subscription was dead.
* in mutate we consumed the chunk iterator and then yielded it to the outside ... so all code listening to that "updated" method were getting an "empty" iterator 🙂 I now recollect the data in an array and yield this to stay compatible
* The PR also adds storing the maxEventNumber now for the NetworkClient as received by subscriptions. Usage of this to limit new subscriptions comes in a follow up PR
* Adjust the close handling of MessageExchanges because i saw issues in details where the changes from some weeks ago were in the end too invasive. We should still resubmit messages that are unacked also when we closed exchanges, unless we are the initiator.
* Add correct handling of maxPeerResponseTime in ClientInteraction
* better type the result of a subscription based on the sustain flag
* Groups invokes are always suppessreponse=true
* Correctly handle "leftoverData" when processing subscription updates

### DIFF
--- a/packages/protocol/src/action/client/ClientInteraction.ts
+++ b/packages/protocol/src/action/client/ClientInteraction.ts
@@ -31,6 +31,7 @@ import {
 } from "#general";
 import { InteractionClientMessenger, MessageType } from "#interaction/InteractionMessenger.js";
 import { Subscription } from "#interaction/Subscription.js";
+import { PeerAddress } from "#peer/index.js";
 import { ExchangeProvider } from "#protocol/ExchangeProvider.js";
 import { SecureSession } from "#session/SecureSession.js";
 import { Status, TlvAttributeReport, TlvNoResponse, TlvSubscribeResponse, TypeFromSchema } from "#types";


packages/node/src/node/client/ClientEventEmitter.ts
```diff
@@ -40,18 +40,26 @@ const warnedForUnknown = new Set<ClusterId | `${ClusterId}-${EventId}`>();
...
+            await node.act(async agent => {
                 // Current ActionContext is not writable, could skip act() but meh, see TODO above
                 //agent.context.priority = occurrence.priority;
                 event.emit(occurrence.value, agent.context);
+
+                const network = agent.get(NetworkClient);
+                if (occurrence.number > network.state.maxEventNumber) {
+                    await agent.context.transaction.addResources(network);
+                    await agent.context.transaction.begin();
+                    network.state.maxEventNumber = occurrence.number;
```
lauckhart, line 60: Should we do this lazily?  E.g. are there high volume events and are they generally treated as idempotent?

packages/protocol/src/protocol/MessageExchange.ts
```diff
@@ -217,6 +217,10 @@ export class MessageExchange {
         return this.#closed;
     }
 
+    get considerClosed() {
```
lauckhart, line 220: This is how I was using `isClosing`, should we get rid of that?

packages/node/src/node/client/ClientEventEmitter.ts
Apollon77: Currently Events are not that many. I think I now got 15 on a device that runs since 3 months without connecting to it ... and events are reset on any restart anyway ... so for now that's ok I guess. But sure I could use a BasicMutex here or such to offload it from the response but still track the promises?

packages/protocol/src/protocol/MessageExchange.ts
Apollon77: I check and remove isClosing likely.

packages/node/src/node/client/ClientEventEmitter.ts
lauckhart: k let's leave as is for now

Apollon77: In theory they can be a lot ... but then it is also chunked so in fact it spreads .... lets monitor that

PR #2827, opened 2025-12-12
Pairednode becomes ClientNode backed

Apollon77:
This PR is the first big step to bring the Legacy "PairedNode" controller API more near together with the ClientNode/Peer based new API.

This basically means:
* All client nodes/peers are now handled by the ServerNode "peers" logic. All former cached node data, the fabric and certificates are initially migrated on first run. Currently the migrated data are not deleted (code for this commented out), we do this after some more test feedback
* The PairedNode still manages reconnections and subscriptions, but all uses the new APIs already under the hood.
* The PairedNode basically now offers both ways for data access: The Old way using ClusterClients and the new API (documentation follows)
* Interaction client is now also "backed" by the interactable from the relevant node
* ClientNodeInteraction now uses the same logic as PairedNode before to determine if requests should be queued or not and queues them (if thread connected device).
* Some classes that are only used by legacy Classes are now moved from protocol to Node (intermediate step).
* Several more optimizations in new and old logics

### DIFF
--- a/.github/workflows/build-test.chip.yml
+++ b/.github/workflows/build-test.chip.yml
@@ -110,7 +110,7 @@ jobs:
 
       - name: Upload chip artifact
         if: needs.workflow-conditions.outputs.build-needed == 'true'
-        uses: actions/upload-artifact@v5
+        uses: actions/upload-artifact@v6
         with:
           name: chip-image
           path: /tmp/chip.tar


packages/general/src/util/Semaphore.ts
```diff
@@ -0,0 +1,242 @@
...
+
+/**
+ * A work slot that must be released when work is complete.
+ */
+export interface WorkSlot extends Disposable {
+    /**
+     * Release the slot manually.
+     * This is called automatically when using `using` syntax.
+     */
+    release(): void;
```
lauckhart: Maybe "close" for consistency?

packages/general/src/util/Semaphore.ts
```diff
@@ -0,0 +1,242 @@
...
+    async obtainSlot(abort?: Abort.Signal): Promise<WorkSlot> {
+        if (this.#closed) {
+            throw new AbortedError("Queue is closed");
+        }
+
+        // Normalize signal
+        const signal = abort ? ("signal" in abort ? abort.signal : abort) : undefined;
+
+        // Check if already aborted
+        if (signal?.aborted) {
```
lauckhart: You can use throwIfAborted()

packages/general/src/util/Semaphore.ts
```diff
@@ -0,0 +1,242 @@
...
+        return promise;
+    }
+
+    /**
+     * Grant a slot immediately (when capacity is available).
+     */
+    #grantSlot(): WorkSlot {
+        this.#runningCount++;
+        logger.debug(
+            "Slot granted. Running:",
```
lauckhart: Should use Diagnostic.dict, although I think maybe this is going to be too chatty as is, maybe these log messages are just temporary?

packages/general/src/util/Semaphore.ts
```diff
@@ -0,0 +1,242 @@
...
+
+        // Grant the slot to the next waiter
+        const slot = this.#grantSlot();
+        next.resolve(slot);
+    }
+
+    /**
+     * Clear the queue.
+     *
+     * @param reject - If true, reject all pending slot requests with an AbortedError
```
lauckhart: What's the use case for clearing without rejecting?  I don't think we should be create unsettled promises. I would maybe remove this and leave close()

If a caller doesn't want to deal with abort exceptions then they need to terminate their workers separately from this interface

packages/general/src/util/Semaphore.ts
```diff
@@ -0,0 +1,242 @@
...
+     * @returns A disposable work slot
+     * @throws AbortedError if the abort signal is triggered before a slot is obtained
+     */
+    async obtainSlot(abort?: Abort.Signal): Promise<WorkSlot> {
+        if (this.#closed) {
+            throw new AbortedError("Queue is closed");
+        }
+
+        // Normalize signal
+        const signal = abort ? ("signal" in abort ? abort.signal : abort) : undefined;
```
lauckhart: One patternto consider is to use Abort.  This will simplify the code, but more importantly:

I would also create an #abort: Abort member.  This would allow you to create an Abort here like const abort = new Abort({ abort: [ abort, this.#abort ] }).

Then, close() just needs to call this.#abort() and any queued tasks will abort automatically.  So we don't end up with multiple paths to terminate waiting...  Either you're aborted or not

packages/general/src/util/Semaphore.ts
```diff
@@ -0,0 +1,242 @@
...
+        const entry = {
+            resolve: resolver,
+            reject: rejecter,
+            signal,
+            abortListener: undefined as (() => void) | undefined,
+        };
+
+        // Set up abort listener if signal provided
+        if (signal) {
+            entry.abortListener = () => {
```
lauckhart: How does this logic trigger the next queued task?  I'd suggest #releaseSlot in a finally so you're sure it's released

packages/general/src/util/Semaphore.ts
```diff
@@ -0,0 +1,242 @@
...
+
+        const entry = {
+            resolve: resolver,
+            reject: rejecter,
+            signal,
+            abortListener: undefined as (() => void) | undefined,
+        };
+
+        // Set up abort listener if signal provided
+        if (signal) {
```
lauckhart: If you have a local `abort` as described above then you can just convert the promise to `abort.race(promise)` and use a finally for cleanup.

packages/general/src/util/Semaphore.ts
```diff
@@ -0,0 +1,242 @@
...
+                }
+                rejecter(signal.reason ?? new AbortedError());
+            };
+            signal.addEventListener("abort", entry.abortListener, { once: true });
+        }
+
+        logger.debug("Queueing slot request at position", this.#queue.length + 1);
+        this.#queue.push(entry);
+
+        return promise;
```
lauckhart: Rather than returning this promise I'd just await it here, with a catch that calls `#releaseSlot()`

packages/general/src/util/Semaphore.ts
```diff
@@ -0,0 +1,242 @@
...
+ */
+export class Semaphore {
+    readonly #delay: Duration;
+    readonly #queue = new Array<{
+        resolve: (slot: WorkSlot) => void;
+        reject: (reason?: any) => void;
+        abortListener?: () => void;
+        signal?: AbortSignal;
+    }>();
+    #delays = new Set<Timer>();
```
lauckhart: Hmm, why are there multiple delays?  Seems like we could just queue entries that are delayed and have a single delay timer to dequeue the next worker and restart the delay if there's another worker pending

packages/general/src/util/Semaphore.ts
```diff
@@ -0,0 +1,242 @@
...
+            this.#runningCount,
+            "Delayed:",
+            this.#delayCount,
+            "Queued:",
+            this.#queue.length,
+        );
+
+        if (this.#delay > 0) {
+            // Keep the slot blocked during delay, then release
+            const delay = Time.getTimer("Queue delay", this.#delay, () => {
```
lauckhart: This I think ends up being a "delay-per-slot", which is interesting, but I think semantically we probably want more of a "delay between workers", which a single delay timer as mentioned above would offer

packages/general/src/util/Semaphore.ts
```diff
@@ -0,0 +1,242 @@
...
+     * Get the number of currently active slots.
+     */
+    get running() {
+        return this.#runningCount;
+    }
+
+    /**
+     * Close the queue and reject all pending slot requests.
+     */
+    close(): void {
```
lauckhart, line 225: This should be async and wait until all queue slots are released before returning.

packages/general/test/util/SemaphoreTest.ts
```diff
@@ -0,0 +1,821 @@
...
+            resolveBlocker();
+
+            await Promise.all([t2, t3, t4]);
+
+            // Tasks should execute in order they were added
+            expect(executionOrder).deep.equals([1, 2, 3, 4]);
+        });
+    });
+
+    describe("abort handling", () => {
```
lauckhart, line 255: Need a test to confirm that abort() triggers next queued task

packages/general/src/util/Semaphore.ts
```diff
@@ -0,0 +1,242 @@
...
+
+import { AbortedError } from "#MatterError.js";
+import { Duration } from "#time/Duration.js";
+import { Time, Timer } from "#time/Time.js";
+import { Instant } from "#time/TimeUnit.js";
+import { Logger } from "../log/Logger.js";
+import { Abort } from "./Abort.js";
+import { createPromise } from "./Promises.js";
+
+const logger = Logger.get("PromiseQueue");
```
lauckhart: Need to update module name

packages/protocol/src/peer/ControllerCommissioner.ts
lauckhart, line 1: So we will no longer support commissioning without using `Node`?  I guess that's probably OK given that controllers tend to be heavier resource users anyway but I worry that @matter/node is getting pretty big and it's harder to keep good separation of concerns.

If we plan to rewrite commissioning using `Interactable` directly could we leave this stuff in @matter/protocol?  Would it be possible to leave for now so we can decide later?

packages/matter.js/src/cluster/client/AttributeClient.ts
lauckhart, line 1: Can this "legacy" stuff go into `matter.js` instead?  Or if still needed for commissioning flows could it stay with that?  Wondering about the value of an interim move

packages/matter.js/src/cluster/client/InteractionClient.ts
```diff
@@ -0,0 +1,1136 @@
...
+
+    /** Use an extended Message Response Timeout as defined for FailSafe cases which is 30s. */
+    useExtendedFailSafeMessageResponseTimeout?: boolean;
+
+    /** Skip request data validation. Use this only when you know that your data is correct and validation would return an error. */
+    skipValidation?: boolean;
+};
+
+export class InteractionClientProvider {
+    readonly #owner: ServerNode;
```
lauckhart, line 122: Is this maybe what caused all of this stuff to jump from protocol into node?  Taking a quick scan it doesn't seem like it uses too much of node, maybe a small interface to act as an adapter would suffice?

packages/node/src/node/client/ClientNodeInteraction.ts
```diff
@@ -42,15 +46,43 @@ export class ClientNodeInteraction implements Interactable<ActionContext> {
         }
     }
 
+    get #queue() {
+        const queue = this.#node.owner?.peers.queue;
```
lauckhart: Could queuing logic instead go into `ClientInteraction`?  Ideally `ClientNodeInteraction` is just an adapter with minimal logic that is specific to `ClientNode`...  The bulk of protocol logic then living in `@matter/protocol`

packages/node/src/node/client/ClientStructure.ts
```diff
@@ -238,9 +238,17 @@ export class ClientStructure {
     }
 
     /** Reference to the default subscription used when the node was started. */
```
lauckhart: this comment seems wrong

packages/node/src/node/client/Peers.ts
```diff
@@ -101,6 +109,13 @@ export class Peers extends EndpointContainer<ClientNode> {
         }
     }
 
+    get queue() {
```
lauckhart: As a low-level protocol concern I don't think this is appropriate in the high-level semantic interface for interacting with peers

packages/node/src/node/client/Peers.ts
```diff
@@ -117,16 +132,19 @@ export class Peers extends EndpointContainer<ClientNode> {
...
      */
     locate(options?: Discovery.Options) {
         return new InstanceDiscovery(this.owner, options);
     }
 
     /**
-     * Employ discovery to find a set of commissionable nodes.
+     * Employ discovery to find a set of commissionable nodes, the options can be used to limit the discovered devices
+     * (e.g. just a specific vendor).
+     * TODO: Allow to provide an array of options for multiple discoveries (e.g. from a Multi QR code).
```
lauckhart: Rather than an array of Options we should probably extend the union type that Options extends to include `{ identifiers: CommissionableDeviceIdentifiers[] }`...  Then in the "multi" case you could do `discover({ identifiers: [...] })`.  But still only have a single timeout and keep the interface simple in the common case

packages/node/src/node/client/Peers.ts
```diff
@@ -242,17 +262,14 @@ export class Peers extends EndpointContainer<ClientNode> {
      * If required, installs a listener in the environment's {@link InteractionServer} to handle subscription responses.
      */
     #configureInteractionServer() {
-        if (this.#closed || this.size > 0 || !this.owner.env.has(InteractionServer)) {
+        if (this.#closed || this.#subscriptionHandler !== undefined || !this.owner.env.has(InteractionServer)) {
```
lauckhart: Maybe lets rename to `#installSubscriptionHandler`.  That's all this does as of now and then this test will make more sense

packages/node/src/node/client/Peers.ts
```diff
@@ -281,6 +298,9 @@ export class Peers extends EndpointContainer<ClientNode> {
     }
 
     #onExpirationIntervalElapsed() {
+        if (this.#closed) {
```
lauckhart, line 301: Ack how can this happen?  Is there something async between setting `#closed` and stopping the timer?  Should reorder then

packages/node/src/node/Node.ts
```diff
@@ -130,6 +133,17 @@ export abstract class Node<T extends Node.CommonRootEndpoint = Node.CommonRootEn
             await this.#runtime.construction.ready;
             await this.act("network startup", agent => agent.get(NetworkBehavior).startup());
         } catch (e) {
+            // If runtime instance got created, tear it down
+            if (this.#runtime) {
+                this.#environment.delete(NetworkRuntime, this.#runtime);
+                try {
+                    await this.#runtime.close();
+                } catch {
+                    // Ignore all errors that might, likely we cannot tear down because construction never completed
```
lauckhart, line 144: Should log a warning or create a `SuppressedError`...  Just so when things are really fubar there's at least a trace as to why

packages/protocol/src/action/client/subscription/ClientSubscribe.ts
```diff
@@ -5,4 +5,9 @@ export interface ClientSubscribe extends Subscribe {
      * If true the subscription is virtualized and the underlying subscription is reestablished when lost.
      */
     sustain?: boolean;
+
+    /**
+     * If true, the request will be queued over all peers of the node
+     */
+    queued?: boolean;
```
lauckhart: It feels wrong to push this decision to the subscriber.  I could see maybe having an override here for very advanced cases, but shouldn't the protocol stack be deciding this?  I was thinking something like `clientNetworkingLimits: { queueSize: 100, threadQueueSize: 5 }` in `NetworkServer#state`

Which I realize isn't in scope for this PR but wondering about sticking it here...  Maybe could add @deprecated if for back-compat?

packages/protocol/src/action/client/ClientRead.ts
```diff
@@ -0,0 +1,14 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { Read } from "#action/request/Read.js";
+
+export interface ClientRead extends Read {
+    /**
+     * If true, the request will be queued over all peers of the node
+     */
+    queued?: boolean;
```
lauckhart: If we need this in public interface let's make a "ClientRequest" interface and have these guys all extend that in addition to their current base type.  Or have the base type extend directly and get rid of the `Client` specialization, and note in jsdoc that not all implementations need support queueing

packages/protocol/src/action/request/Invoke.ts
```diff
@@ -49,15 +76,18 @@ export function Invoke(
...
-    if ("commands" in optionsOrData) {
+    if ("commands" in optionsOrData || "invokeRequests" in optionsOrData) {
         options = optionsOrData;
-        commands = [...optionsOrData.commands, ...commands];
+        if (optionsOrData.commands !== undefined) {
+            commands = [...optionsOrData.commands, ...commands];
+        }
     } else {
-        commands = [optionsOrData, ...commands];
+        commands = [optionsOrData as Invoke.CommandRequest<any>, ...commands];
```
lauckhart: This cast is necessary because the type is no longer tight enough to act as a discriminator, should go away if Invoke definition is fixed

packages/protocol/src/action/request/Invoke.ts
```diff
@@ -123,7 +168,10 @@ export function Invoke(
 export namespace Invoke {
     export interface Definition {
         /** List of commands to invoke */
-        commands: Invoke.CommandRequest<any>[];
+        commands?: Invoke.CommandRequest<any>[];
```
lauckhart: This cannot be optional.  If we need a separate "invokeRequests" then `Definition` needs to be a union type where one or the other is set

(but hopefully we don't need that, see next comment)

packages/protocol/src/action/request/Invoke.ts
```diff
@@ -123,7 +168,10 @@ export function Invoke(
...
+
+        /** List of prepared raw commands to invoke */
+        invokeRequests?: InvokeCommandData[];
```
lauckhart: It looks like the only place this is used is the adapter in `InteractionClient`...  That adapter appears to have everything necessary to construct a `CommandRequest` directly, it's just doing some of the work of assembling the commands itself

Since this screws up the public interface I think it makes sense to remove this and modify `InteractionClient` to use this interface as designed, which should reduce complexity without too much effort.  I'll add further detail in `InteractionClient`

packages/node/src/node/client/legacy/InteractionClient.ts
```diff
@@ -0,0 +1,1136 @@
...
+            }
+            if (requestSchema.isFabricScoped && request.fabricIndex === undefined) {
+                request.fabricIndex = FabricIndex.NO_FABRIC;
+            }
+        }
+
+        if (!skipValidation) {
+            requestSchema.validate(request);
+        }
+        const commandFields = requestSchema.encodeTlv(request);
```
lauckhart: Rather than encoding yourself I think you can just provide the TLV and other parameters in an `Invoke` and let `ClientInteraction` take care of this for you with its existing interface.  `Invoke` is designed in terms of "bare" matter values + TLV schema such as you have here

packages/protocol/src/action/request/Invoke.ts
```diff
@@ -231,8 +280,13 @@ export namespace Invoke {
         return data;
     }
 
-    export function commandOf<const R extends CommandRequest>(request: R): ClusterType.Command {
+    export function commandOf<const C extends Specifier.Cluster, R extends CommandRequest<C>>(
+        request: Omit<R, "cluster"> & { cluster?: C },
+    ): ClusterType.Command {
         if (typeof request.command === "string") {
+            if (request.cluster === undefined) {
+                throw new InternalError("Cluster must be defined to resolve command by name");
```
lauckhart: This is duplicative of the following message, maybe `Specifier.clusterFor` needs undefined handling added?

packages/protocol/src/action/request/Invoke.ts
```diff
@@ -241,9 +295,9 @@ export namespace Invoke {
             if (command === undefined) {
                 throw new MalformedRequestError(`Cluster ${cluster.name} does not define command ${request.command}`);
             }
-            return command as Specifier.CommandFor<Specifier.ClusterOf<R>, R["command"]>;
+            return command;
         }
-        return request.command as Specifier.CommandFor<Specifier.ClusterOf<R>, R["command"]>;
+        return request.command;
```
lauckhart: It looks like this may have been intended to provider better typing for the result, is it possible to remove the `ClusterType.Command` return type on the function instead?

packages/protocol/src/peer/PhysicalDeviceProperties.ts
```diff
@@ -78,7 +78,7 @@ export namespace PhysicalDeviceProperties {
         }
         if (maxIntervalCeiling < defaultCeiling) {
             logger.debug(
-                `${description}: maxIntervalCeilingSeconds ideally is ${defaultCeiling}s instead of ${maxIntervalCeiling}s due to device type`,
+                `${description}: maxIntervalCeilingSeconds ideally is ${Duration.format(defaultCeiling)}s instead of ${Duration.format(maxIntervalCeiling)}s due to device type`,
```
lauckhart: Should take the "s" suffix out too since the formatter will add correct suffix

lauckhart (commented):
cool!

packages/protocol/src/action/client/ClientRead.ts
Apollon77: I did the ClientRequest interface

packages/protocol/src/action/client/subscription/ClientSubscribe.ts
Apollon77: I agree that we should change this. I added a TODO and also deprecate. For now I need it that the Controller works "compatible".

packages/node/src/node/client/ClientNodeInteraction.ts
Apollon77: For me honestly ClientInteraction is about the "protocol level" of communication and on that level nothing mentions or cares about optimizing "multi node communication". This is pure "higher level" things that are implementation choice frommmy perspective and only makes sense to consider when look at it as a controller managing multiple nodes with potentially different transport protocols.
So that's why I decided to put that here.

Also in my eyes this does not really blow up the logic here

I added a TODO to reconsider this choice when we reconsider where to best place this when thinking about it. But even if we introduce something on "network" level as you mentioned in an other comment it would be "node" focussed, or?

packages/node/src/node/client/Peers.ts
Apollon77: I agree. I just added it generically without really meaning a concrete implementation. I adjusted the todo

packages/node/src/node/client/Peers.ts
Apollon77: It happens because the "this.#expirationInterval.cancel()" on the sleep just cancels the sleep but still executes the logic afterwards. Si this if is there to prevent that the logic gets executed.

We could cancel with a reason that it throws but then also need to catch and log it or such.

How we should manage it?

packages/node/src/node/client/Peers.ts
Apollon77: As written in another feedback in my eyes this is higher level "custom logic" which is implementation choice. Matter protocol does not mention any of that - it is really "multi node optimization" and for me "peers" was exactly this level tbh.

And it is basically not just "networking", it is "interaction" based (meaning to ensure that multiple messages of one interaction are executed "together" or not.

Open to discuss.

packages/protocol/src/peer/ControllerCommissioner.ts
Apollon77: I planned to move it back once gotten rid of the InteractionClient.

So ok here my "problem":
* When I move this back then I also need to move back InteractionClient because an interact would not work here because this class gets directly instanciated ... and InteractionClient needs all the other moved *Client classes 
* When I move back ControllerCommissioningFlow then I also need ClusterClient and the *Client classes moving too.

So from my perspective the only way is to revert the move and leave that legacy stuff still in protocol. We can do that if you do not see an other way. WDYT

packages/node/src/node/client/legacy/InteractionClient.ts
Apollon77: ahh yeeahh me blond... sorry

packages/protocol/src/action/request/Invoke.ts
Apollon77: removed

packages/protocol/src/action/request/Invoke.ts
Apollon77: reverted

packages/protocol/src/action/request/Invoke.ts
Apollon77: reverted

packages/node/src/node/client/ClientNodeInteraction.ts
Apollon77: Adjusted as discussed

packages/node/src/node/client/Peers.ts
Apollon77: Adjusted as discussed

packages/general/src/util/Semaphore.ts
Apollon77: The queue is just the future. All tasks already started and running can not be aborted, so there is no slot assigned so also nothing to release. i added comment and moved the rejector into the "if" so that it only rejects if the task was not yes fulfilled.

packages/general/src/util/Semaphore.ts
Apollon77: Why? This promise fulfills when the slot is granted - before that there is no reserved slot to release. And afterwards it can not fail here.

packages/general/test/util/SemaphoreTest.ts
Apollon77: but why it should? abort only aborts future/unstarted tasks ... so when ready then the queue is checked

packages/general/src/util/Semaphore.ts
Apollon77: Hm ... how to ensure that this is not ending in any deadlock or such? Do we really should add this? We ensure that we are not giving out mre slots, so for me the work here is done by making sure to not add more

packages/protocol/src/peer/ControllerCommissioningFlow.ts
```diff
@@ -320,28 +323,57 @@ export class ControllerCommissioningFlow {
...
+                    continue;
+                }
+                attributeMap.set(key, value);
+            }
+        }
+        return [...attributeMap.values()];
+    }
+
+    // TODO improve response typing
+    async #invokeCommand(request: ClientInvoke) {
```
lauckhart: It looks like you could make this take an Invoke.CommandRequest (maybe with an option for timed request).  Then the individual invokes would each be 6 lines shorter

packages/protocol/src/peer/ControllerCommissioner.ts
lauckhart: I like the solution you came up with :)

packages/general/src/util/Semaphore.ts
lauckhart: awaiting here is for a better stack trace but right on the other point, I was thinking the promise resolves when the slot is released

packages/general/src/util/Semaphore.ts
lauckhart: OK duh makes sense

packages/general/src/util/Semaphore.ts
lauckhart: I was thinking in terms of "close" resulting in the semaphore being truly unused, vs. "still alive just can't grow".  The "half closed" state makes sense too but not sure if it should be the only option

The main reason to have it is that a component that has semaphored workers would be able to say `await semaphore.close()` and then proceed confident that all workers have stopped.  Without that you would need to track workers separately in another construct to ensure all workers are complete

Avoiding deadlock is just a.) cancel workers if necessary before closing semaphore, and b.) do not await semaphore close in a worker

PR #2838, opened 2025-12-15
Multi controller example

bhargavnagotha9:
Added multi controller within same fabric functionality example.

### DIFF
--- a/packages/examples/package.json
+++ b/packages/examples/package.json
@@ -40,7 +40,7 @@
         "matter-air-quality-sensor": "matter-run src/device-air-quality-sensor/AirQualitySensorDeviceNode.ts",
         "matter-smoke-co-alarm": "matter-run src/device-smoke-co-alarm/SmokeCoAlarmDeviceNode.ts",
         "matter-controller": "matter-run src/controller/ControllerNode.ts",
-        "matter-cmd": "matter-run src/multi-controller/ControllerCommand.ts",
+        "matter-shared-fabric": "matter-run src/multi-controller/SharedFabricController.ts",
         "bundle-device": "esbuild src/device-onoff/DeviceNode.ts --bundle --platform=node --conditions=esbuild --external:@stoprocent/bleno --external:@stoprocent/bluetooth-hci-socket --sourcemap --minify --keep-names --outfile=build/bundle/DeviceNode.cjs",
         "matter-device-bundled": "node --enable-source-maps build/bundle/DeviceNode.cjs"
     },
--- a/packages/examples/src/multi-controller/SharedFabricController.ts
+++ b/packages/examples/src/multi-controller/SharedFabricController.ts
@@ -6,14 +6,15 @@
  */
 
 /**
- * Simple command-based Matter controller for multi-controller scenario.
- * Sends commands directly without interactive mode.
+ * Shared Fabric Controller - Matter controller using external ICAC certificates.
+ * Demonstrates multi-controller scenarios where controllers share the same fabric
+ * without re-commissioning, using a 3-tier PKI hierarchy (RCAC -> ICAC -> NOC).
  *
  * Usage:
- *   npm run matter-cmd toggle    - Toggle the device
- *   npm run matter-cmd on        - Turn device ON
- *   npm run matter-cmd off       - Turn device OFF
- *   npm run matter-cmd read      - Read OnOff state
+ *   npm run matter-shared-fabric toggle    - Toggle the device
+ *   npm run matter-shared-fabric on        - Turn device ON
+ *   npm run matter-shared-fabric off       - Turn device OFF
+ *   npm run matter-shared-fabric read      - Read OnOff state
  */
 
 import { Bytes, Crypto, Environment, Logger, LogLevel, StorageService, Time } from "@matter/main";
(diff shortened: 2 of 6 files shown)


packages/examples/src/multi-controller/SharedFabricController.ts
Apollon77, line 1: maybe we foind a better name?Controller command feels like very generic and we also have two other controller examples. In fact this is a special one that shows certificate specialities, right?

packages/examples/package.json
```diff
@@ -40,6 +40,7 @@
         "matter-air-quality-sensor": "matter-run src/device-air-quality-sensor/AirQualitySensorDeviceNode.ts",
         "matter-smoke-co-alarm": "matter-run src/device-smoke-co-alarm/SmokeCoAlarmDeviceNode.ts",
         "matter-controller": "matter-run src/controller/ControllerNode.ts",
+        "matter-cmd": "matter-run src/multi-controller/ControllerCommand.ts",
```
Apollon77: we should also change name here ...

packages/protocol/src/certificate/CertificateAuthority.ts
```diff
@@ -275,13 +279,23 @@ export class CertificateAuthority {
...
-            (typeof certValues.nextCertificateId === "number" || typeof certValues.nextCertificateId === "bigint")
-        );
+            (typeof certValues.nextCertificateId === "number" || typeof certValues.nextCertificateId === "bigint");
+
+        if (!hasBasicFields) {
+            return false;
+        }
+
+        // rootKeyPair is optional when ICAC is present (3-tier PKI without RCAC private key)
+        const hasRootKeyPair =
```
Apollon77: ok, maybe better "rootKeyPairValid? Because "=== undefined" and "has" collides in my mind :-)

packages/protocol/src/certificate/CertificateAuthority.ts
```diff
@@ -370,9 +398,16 @@ export namespace CertificateAuthority {
...
+     * Configuration for CertificateAuthority with external certificates.
+     *
+     * When using ICAC (3-tier PKI), the rootKeyPair can be omitted since NOCs are signed
+     * by the ICAC, not the RCAC. This allows controllers to operate without access to
+     * the RCAC private key.
+     */
     export type Configuration = {
         rootCertId: bigint;
-        rootKeyPair: BinaryKeyPair;
+        rootKeyPair?: BinaryKeyPair; // Optional when using ICAC - only public key needed from rootCertBytes
```
Apollon77: maybe time now to split COnfiguration type into one "WithIcac" and one WithoutIcac"? Then the type here fir the "rootKeyPair" could be made exactly ... then it is only optional when it is "WithIcac". But thats onlyneeded here. For storage the optiomal is sufficient

packages/examples/src/multi-controller/SharedFabricController.ts
```diff
@@ -102,7 +103,7 @@ function getDefaultChipToolIpk(): Bytes {
 }
 
 async function main() {
-    const certDir = process.env.MATTER_CERTDIR ?? "./certificates";
+    const certDir = process.env.MATTER_CERTDIR ?? "/home/mantra/pooja/connectedhomeip/certificates";
```
Apollon77, line 106: I think you do not want to have this path in here, right? :-)

packages/examples/src/multi-controller/SharedFabricController.ts
```diff
@@ -111,8 +112,8 @@ async function main() {
     // Certificate paths - Note: RCAC private key is NOT required when using ICAC
     // Only need: RCAC cert (public), ICAC cert, ICAC private key
     const rcacCertPath = path.join(certDir, "rcac.chip");
-    const icacCertPath = path.join(certDir, "icac.chip");
-    const icacKeyPath = path.join(certDir, "icac_key.bin");
+    const icacCertPath = path.join(certDir, "icac2.chip");
```
Apollon77, line 115: Why this rename? i guess should be as before, right?

packages/protocol/src/certificate/CertificateAuthority.ts
```diff
@@ -290,12 +298,12 @@ export class CertificateAuthority {
         }
 
         // rootKeyPair is optional when ICAC is present (3-tier PKI without RCAC private key)
-        const hasRootKeyPair =
+        const rootKeyPairValid =
```
Apollon77, line 301: Please rename back because in fact it only tells that it has it ... it is not validation beside pure datatype :-)

packages/protocol/src/certificate/CertificateAuthority.ts
```diff
@@ -398,25 +406,45 @@ export namespace CertificateAuthority {
         intermediateCert?: boolean;
     };
 
+    /** Base configuration fields shared by both 2-tier and 3-tier PKI */
```
Apollon77, line 409: I think now there are conflicts between my changes here and yours, also see the other conflicts on file level. I removed the "intermediateCert boolean" already ... please rebase.

packages/examples/src/multi-controller/SharedFabricController.ts
bhargavnagotha9: yes,dont want it.

PR #2841, opened 2025-12-16
Adding OTA support for device and controller to matter.js

Apollon77:
This PR adds OTA support for device and controller to matter.js.

Can be reviewed by Commit, some summary information:
* We patch the "updateStateProgress" attribute of the OtaUpdateRequestor cluster to be Q to have automatic subscription suppression to 1 report per second because spec recommends limiting here
* The PersistedFileDesignator and Bdx Protocol layer now allows to register different storages for different transfer use cases (scopes). Currently these are the two BDX relevant cases "ota" and "log" (second one not yet implemented). With this storagecontexts can be marked for a purpose to use them as target or source for transfers. Also the "FileDesignator" for the transfer can so be different to the one we use for storage
* While the DclUpdateService is global the PR introduces a SoftwareUpdateManager as companion to the OtaSoftwareUpdateprovider cluster that manages logic for all peers of a node.
* It adds OTA support to PairedNode and the Controller implementation
* It adds OTA support to the matter shell
* It contains several fixes that came up while real-life testing OTA with real devices
* It also introduces a test that validates a successful OTA update process with our controller and device

Missing yet (will be added here or separately):
* More documentation and code documentation in the clusters
* Automatically register the Controller as OTA provider with all nodes it is paired too and update that daily
* some smaller TODOs

### DIFF
--- a/packages/protocol/src/dcl/DclOtaUpdateService.ts
+++ b/packages/protocol/src/dcl/DclOtaUpdateService.ts
@@ -42,7 +42,7 @@ export type OtaUpdateInfo = DeviceSoftwareVersionModelDclSchema;
 
 const OTA_DOWNLOAD_TIMEOUT = Minutes(5);
 
-const OTA_FILENAME_REGEX = /^([0-9a-f]+[./][0-9a-f]+[./](prod|test))$/i;
+const OTA_FILENAME_REGEX = /^[0-9a-f]+[./][0-9a-f]+[./](?:prod|test)$/i;
 
 /**
  * Service to query and manage OTA updates from the Distributed Compliance Ledger (DCL), but also allows to inject own


packages/general/src/crypto/StreamingStandardCrypto.ts
```diff
@@ -0,0 +1,32 @@
+import { Bytes } from "#util/index.js";
+import { Crypto, HashAlgorithm } from "./Crypto.js";
+import { StandardCrypto } from "./StandardCrypto.js";
+
+/**
+ * StandardCrypto implementation that supports streaming hash computation.
+ * Important: It collects chunks from async iterators into memory before computing the hash.
+ *            This is mainly used in testing because of that
```
lauckhart: I think it would be OK to put this in the base implementation. Could log a warning there

packages/node/src/behavior/system/software-update/SoftwareUpdateManager.ts
```diff
@@ -0,0 +1,741 @@
...
+
+    /** ReleaseNotesUrl for the update */
+    releaseNotesUrl?: string;
+
+    /** SpecificationVersion of the updated device */
+    specificationVersion?: number;
+}
+
+/**
+ * The Software Update Manager is the instance to bridge between the central OTA store and DCL service and mange the
```
lauckhart: Typo "mange"

packages/node/src/behavior/system/software-update/SoftwareUpdateManager.ts
```diff
@@ -0,0 +1,741 @@
...
+ * The following state allows configuring the behavior:
+ * * `startupConsents`: Use this to pre-initialize consent to allow nodes to update automatically. The value will not
+ *     be persisted, and change during runtime is only supported via `addUpdateConsent`.
+ * * `allowTestOtaImages`: When set to true, we also query the Test DCL additionally to the production DCL for update
+ *     and use this when a newer version is found. Default is false
+ * * `updateCheckInterval`: By default, we check the DCL for updates every 24h. This state value allows adjusting this.
+ * * `announceAsDefaultProvider`: By default, we announce ourselves as a default update provider to all nodes in the fabric.
+ *     Set to "false" if this is not wanted and updates are only pushed on availability
+ */
+export class SoftwareUpdateManager extends Behavior {
```
lauckhart, line 117: If this is a server component maybe "UpdateSchedulerServer" or something?  "Manager" is pretty generic

packages/node/src/behavior/system/software-update/SoftwareUpdateManager.ts
```diff
@@ -0,0 +1,741 @@
...
+interface UpdateConsent {
+    vendorId: VendorId;
+    productId: number;
+    targetSoftwareVersion: number;
+    peerAddress: PeerAddress;
+}
+
+interface UpdateQueueEntry extends UpdateConsent {
+    endpoint: Endpoint;
+    lastProgressUpdateTime?: number;
```
lauckhart: What are the units here? Timestamp?

packages/node/src/behavior/system/software-update/SoftwareUpdateManager.ts
```diff
@@ -0,0 +1,741 @@
...
+        this.internal.services?.close(DclOtaUpdateService);
+        this.internal.versionUpdateObservers.close();
+        await super[Symbol.asyncDispose]?.();
+    }
+}
+
+export namespace SoftwareUpdateManager {
+    export class State {
+        /** Use this to pre-initialize consent to allow nodes to update automatically. The content will not be persisted! */
+        startupConsents = new Array<UpdateConsent>();
```
lauckhart: These are tied too a peer and it would be nice if they would be persistent.  Could we have have a Peer-specific Behavior to store OTA state?

packages/protocol/src/action/client/ClientBdx.ts
```diff
@@ -0,0 +1,19 @@
...
+ * @license
+ * Copyright 2022-2025 Matter.js Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { RequestContext } from "#action/client/ClientInteraction.js";
+import { BdxMessenger } from "#bdx/BdxMessenger.js";
+import { Duration, WorkSlot } from "#general";
+
+export interface ClientBdxOptions {
```
lauckhart: I'm curious why this is "Options" and not some type of "Request"

packages/node/src/behavior/system/software-update/OtaAnnouncements.ts
```diff
@@ -0,0 +1,240 @@
...
+            const peerAddress = PeerAddress(peer.maybeStateOf(CommissioningClient)?.peerAddress);
+            if (peerAddress === undefined || this.#announcementQueue.some(p => PeerAddress.is(p, peerAddress))) {
+                continue;
+            }
+            this.#queuePeer(peerAddress);
+        }
+        this.#announcementTimer.start();
+    }
+
+    #queuePeer(peerAddress: PeerAddress) {
```
lauckhart, line 79: Why do we queue the address rather than the ClientNode?

packages/node/src/behavior/system/software-update/OtaAnnouncements.ts
```diff
@@ -0,0 +1,240 @@
...
+import { Node } from "#node/Node.js";
+import type { ServerNode } from "#node/ServerNode.js";
+import { Fabric, PeerAddress, Write } from "#protocol";
+import { EndpointNumber } from "#types";
+
+const logger = new Logger("OTAAnnouncements");
+
+export class OtaAnnouncements {
+    #announcementQueue = new Array<PeerAddress>();
+    #announcementTimer: Timer;
```
lauckhart, line 24: Is there any reason to do this with bare timers rather than a loop in an async function with (abortable) sleeps?  I find the logic a lot easier to follow that way

packages/node/src/behavior/system/software-update/SoftwareUpdateManager.ts
```diff
@@ -0,0 +1,741 @@
...
+    }
+
+    async #nodeOnline() {
+        if (this.internal.announcements !== undefined) {
+            await this.internal.announcements.close();
+            this.internal.announcements = undefined;
+        }
+
+        const fabricAuthority = this.env.get(FabricAuthority);
+        const ownFabric = fabricAuthority.fabrics[0];
```
lauckhart, line 145: We can only provide updates if we're the authority?  Is that a matter thing?

packages/node/src/behavior/system/software-update/SoftwareUpdateManager.ts
```diff
@@ -0,0 +1,741 @@
...
+            .filter(
+                ({
+                    vendorId: candidateVendorId,
+                    productId: candidateProductId,
+                    softwareVersion: candidateSoftwareVersion,
+                    minApplicableSoftwareVersion = 0,
+                    maxApplicableSoftwareVersion = softwareVersion,
+                    mode,
+                }) =>
+                    vendorId === candidateVendorId &&
```
lauckhart: Is vendor/product test redundant because we used find above to filter already?

packages/protocol/src/dcl/DclOtaUpdateService.ts
```diff
@@ -579,30 +644,24 @@ export class DclOtaUpdateService {
...
+    ) {
         if (this.#storage === undefined) {
             await this.construction;
         }
         const storage = this.#storage!;
 
         // Get all keys from storage
-        const keys = await storage.keys();
+        const keys = await storage.context.keys();
         logger.debug(`Scanning stored OTA files, found ${keys.length} total files`);
```
lauckhart: It seems like we'll say this a lot with lots of nodes...  Could we have updates organized hierarchically so we don't do o(n) search on disk for every lookup for a specific product?

packages/node/src/behaviors/ota-software-update-requestor/OtaSoftwareUpdateRequestorServer.ts
```diff
@@ -4,14 +4,1282 @@
...
+ * The default implementation provides anything needed to check for updates and transfer new updates, according to the
+ * Matter specification, but applying the update is too specific to your use case and the environment where the device
+ * is used.
+ *
+ * The following custom state attributes are available to configure the behavior of the cluster:
+ * * `updateQueryInterval`: Interval to check for updates, default: 24h as defined in Matter specification
+ * * `transferProtocolsSupported`: List of Transfer protocols that are announced as supported. By default, we only
+ *     export BDX; we also support HTTPS (but basically no one else)
+ * * `canConsent`: Can the node request consent from the user for the update itself? Default is set to true, and the
+ *     controller needs to take care
```
lauckhart: Lines between the bullets make them a little easier to parse. What does "take care" mean?

packages/node/src/behaviors/ota-software-update-provider/OtaSoftwareUpdateProviderServer.ts
```diff
@@ -4,14 +4,467 @@
...
+    Granted,
+    Denied,
+    Obtaining,
+    Unknown,
+}
+
 /**
  * This is the default server implementation of {@link OtaSoftwareUpdateProviderBehavior}.
+ *
+ * This cluster is usually used within the Client and allows Servers to handle OTA software updates.
```
lauckhart: Am confused by this comment. What does "within the client" mean?

packages/node/src/behaviors/ota-software-update-provider/OtaSoftwareUpdateProviderServer.ts
```diff
@@ -4,14 +4,467 @@
...
+ * For the state and detailed usage for Clients, please check the SoftwareUpdateManager behavior which provides
+ * configurability for this.
+ *
+ * For special use cases the cluster provides the following extension point methods:
+ * * `checkUpdateAvailable`: By default, the implementation uses the SoftwareUpdatemanager to check for available
```
lauckhart: If you make names links it's easier to jump to relevant code and make it s little clearer you're referring to a method

packages/node/src/behavior/system/software-update/SoftwareUpdateManager.ts
Apollon77: It basically just is a companion behavior to allow more functionality ... and in fact more lives in a client/controller, thats why I chose to not have server" on it

packages/protocol/src/action/client/ClientBdx.ts
Apollon77: I had it Request but bascially there are no real "request" parameter there beside specifying a timeout ... I can rename.

packages/node/src/behavior/system/software-update/OtaAnnouncements.ts
Apollon77: it is "delayed" what happens to it ... I wanted to avoid issues like "the node gets deleted in  between ... so I chose the "lazy connection" way

packages/general/src/crypto/StreamingStandardCrypto.ts
Apollon77: I added and only print a warning when data is more than 100kb ...

packages/node/src/behavior/system/software-update/SoftwareUpdateManager.ts
Apollon77: Adjusted

packages/node/src/behavior/system/software-update/SoftwareUpdateManager.ts
Apollon77: I thought about that a long time before and in my case there were some pitfalls when we store this. The question was if the expectation is that after a restart such consent would stay ... I decided for now to remove this startupConsent thing because the concept is that devs can implement own auto-consent logic. I moved it to "Internal"

packages/protocol/src/action/client/ClientBdx.ts
Apollon77: renamed

packages/node/src/behavior/system/software-update/OtaAnnouncements.ts
Apollon77: mainly because was convenient to reuse the timer for an initial delay started in constructor and then re-set for the others ... and even with a cancelable promise I need to track and handle it in several places. Was not that much of a benefit for me tbh

packages/node/src/behavior/system/software-update/SoftwareUpdateManager.ts
Apollon77: Hm ... no ... effectively we could use FabricManager and even iterate over ll fabrics... not sure if that would be a good idea, but yes then any bridge device could claim itself as" the default OTA in the fabric" ... would be a strange thing. I add a comment ... lets see if that use case ever comes up

packages/node/src/behavior/system/software-update/SoftwareUpdateManager.ts
Apollon77: yeahh we should trust our logic

packages/protocol/src/dcl/DclOtaUpdateService.ts
Apollon77: Valid.. ok i changed the structure, now checking for one known update is easy, just listing all or such (which should be rare cases) is a bit more effort now. But now automatically sub context are created and we store by vendorid.productid.(text|prod)

packages/node/src/behaviors/ota-software-update-requestor/OtaSoftwareUpdateRequestorServer.ts
Apollon77: fixed

packages/node/src/behaviors/ota-software-update-provider/OtaSoftwareUpdateProviderServer.ts
Apollon77: adjusted

packages/node/src/behaviors/ota-software-update-provider/OtaSoftwareUpdateProviderServer.ts
Apollon77: done.

packages/protocol/src/dcl/DclOtaUpdateService.ts
```diff
@@ -42,7 +42,7 @@
 
 const OTA_DOWNLOAD_TIMEOUT = Minutes(5);
 
-const OTA_FILENAME_REGEX = /^([0-9a-f]+)-([0-9a-f]+)-(prod|test)$/i;
+const OTA_FILENAME_REGEX = /^[0-9a-f]+[./][0-9a-f]+[./]prod|test$/i;
```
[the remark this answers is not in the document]
Apollon77: applied

packages/node/src/behavior/system/software-update/SoftwareUpdateManager.ts
lauckhart: I think consent is consent and would be annoying if user consented, power went out, then user needed to consent again...  So every implementer will want to persist it so could be something that we offer in future

packages/node/src/behavior/system/software-update/OtaAnnouncements.ts
lauckhart: K, it should always be safe to keep a reference to an endpoint around if you check deletion before using it, which I think we roughly do here

packages/node/src/behavior/system/software-update/SoftwareUpdateManager.ts
lauckhart: If we're not already we should be at least iterating over all of the fabrics we own (as controller can have more than own).  Here though I think we're just assuming we own the first fabric to determine if we own any

PR #2842, opened 2025-12-17
Add GlobalFabricId

lauckhart:
Adds a type for the "compressed" fabric ID and converts the fabric.operationalId to this.  Allows us to reference fabrics with a single global key without performing an index lookup and generally makes access more convenient.  Modifies `FabricManager` so that both FabricIndex and GlobalFabricId can be used as keys.


packages/protocol/src/fabric/Fabric.ts
```diff
@@ -557,7 +566,7 @@ export namespace Fabric {
         fabricId: FabricId;
         nodeId: NodeId;
         rootNodeId: NodeId;
-        operationalId: Bytes;
+        globalId: GlobalFabricId;
```
Apollon77: I guess this change breaks the storage of fabric s... or better the "loading them again". So we need to have some mapping somewhere any typing wise maybe the old key still there as "deprecated" to allow such a mapping. Or we need to check and recalculate this ID when it is empty on load ...

packages/types/src/datatype/NodeId.ts
```diff
@@ -32,11 +32,9 @@ export namespace NodeId {
      */
     export const UNSPECIFIED_NODE_ID = NodeId(0);
 
-    export const toHexString = (nodeId: NodeId) => {
-        const writer = new DataWriter();
-        writer.writeUInt64(nodeId);
-        return Bytes.toHex(writer.toByteArray()).toUpperCase();
-    };
+    export function strOf(nodeId: NodeId) {
+        return hex.fixed(nodeId, 16);
```
Apollon77, line 36: it could maybe be good to leave the "toUpperCase" here ... also logging wise maybe a good idea to have that ID which is "usually used uppercase" also in logs always uppercase (instead of "just another lowercase hex number" :-) ) Just for easier visiblity

packages/protocol/src/fabric/Fabric.ts
lauckhart: added backwards-compat for this field in particular...  It's not necessary if you use the new async `create` because it is computed, but if someone calls `new Fabric` it'll still be necessary

packages/types/src/datatype/NodeId.ts
lauckhart: We already log them mostly as `@<fabric-idx>:<node-id>` so this shouldn't really affect presentation in the logs...  So my thinking was that we leave case management to the places where it's required (e.g. DNS) and otherwise stick with usual conventions

Apollon77 (approved):

PR #2878, opened 2025-12-23
Better use on/off methods for scenes changes

Apollon77:
Just setting the state might be missed by some implementations. LevelControl and ColorCOntrol are also calling methods, so let's bring OnOff in sync

### DIFF
--- a/packages/node/src/behaviors/on-off/OnOffServer.ts
+++ b/packages/node/src/behaviors/on-off/OnOffServer.ts
@@ -250,8 +250,9 @@ export class OnOffBaseServer extends OnOffLogicBase {
         }
         if (onOff) {
             this.delayedPromises.add(this.on());
+        } else {
+            this.delayedPromises.add(this.off());
         }
-        this.delayedPromises.add(this.off());
     }
 
     #delayedOffTick() {
--- a/packages/node/test/behaviors/scenes-management/ScenesManagementServerTest.ts
+++ b/packages/node/test/behaviors/scenes-management/ScenesManagementServerTest.ts
@@ -4,7 +4,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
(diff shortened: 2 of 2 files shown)


packages/node/src/behaviors/on-off/OnOffServer.ts
```diff
@@ -216,7 +216,11 @@ export class OnOffBaseServer extends OnOffLogicBase {
         }
 
         if (transitionTime === 0) {
-            this.state.onOff = onOff;
+            if (onOff) {
+                this.on();
```
lauckhart: Need to do something with promise if `on` and/or `off` are async

packages/node/src/behaviors/on-off/OnOffServer.ts
```diff
@@ -216,15 +216,17 @@ export class OnOffBaseServer extends OnOffLogicBase {
...
+            if (onOff) {
+                return this.on();
+            }
+            return this.off();
         }
+        this.internal.applyScenePendingOnOff = onOff;
+        this.internal.applySceneDelayTimer = Time.getTimer(
+            "delayed scene apply",
+            Millis(transitionTime),
+            this.callback(this.#applyDelayedSceneOnOffValue),
```
lauckhart, line 233: The type for the time callback is actually broken in that the `any` return type prevents detection of the fact that timers do not handle promises

So there's still a problem here I think.  Need to store the promise somewhere so it can be awaited when the behavior is closed, and handle errors

Simplest might be a Multiplex in this.internal.  More ambitious could be an "async timer" utility class that has an internal Multiplex and an async `close` method...  Useful for similar cases in the future.  Aaand fixing Timer.Callback return type

Apollon77: updated with Multiplex

PR #2897, opened 2025-12-28
Allow to skip data version injection for read

Apollon77:
... via ClintNodeInteraction to allow better control for special cases.

### DIFF
--- a/packages/matter.js/src/MatterController.ts
+++ b/packages/matter.js/src/MatterController.ts
@@ -35,6 +35,7 @@ import {
     MatterError,
     MaybePromise,
     Minutes,
+    ObserverGroup,
     ServerAddress,
     ServerAddressUdp,
     StorageBackendMemory,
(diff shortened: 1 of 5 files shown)


packages/protocol/src/action/client/ClientRead.ts
```diff
@@ -7,4 +7,7 @@
 import { Read } from "#action/request/Read.js";
 import { ClientRequest } from "./ClientRequest.js";
 
-export interface ClientRead extends Read, ClientRequest {}
+export interface ClientRead extends Read, ClientRequest {
+    /** Set to true to skip the automatic data version injection */
+    skipDataVersionInjection?: boolean;
```
lauckhart: Maybe call this `readKnowmVersions`?

lauckhart (approved):

packages/protocol/src/action/client/ClientRead.ts
Apollon77: I changed to "includeKnownVersions" (because that we read is already clear.

lauckhart: The problem with that is it is referencing how the request is formulated but could be interpreted as how the response is formulated (include known version data in the read).  "Inject" had the same problem

I was suggesting inverting the value so `readKnownVersions = true` would return data from versions that are known.  Mechanically it would then *not* include versions in the request

PR #2991, opened 2026-01-13
Catch AbortedError when closing peers

Apollon77:


packages/matter.js/src/MatterController.ts
```diff
@@ -523,7 +524,12 @@ export class MatterController {
         const node = await this.node.peers.forAddress(peerAddress);
         const peer = this.node.env.get(PeerSet).for(peerAddress);
         await node.delete();
-        await peer.delete();
+        try {
+            await peer.delete();
+        } catch (error) {
+            // When there are open reconnections we could expect a peer closed abort error here, so ignore this
+            AbortedError.accept(error);
```
lauckhart: The peer should probably catch this error within delete() because it's intended for anyone waiting on a connecting to receive. The delete isn't what is actually aborted, this is just a side effect of waiting for the connecting attempt to end

Apollon77: Ok I move it there

PR #3006, opened 2026-01-14
Override Enum Conformance for FanControl

Apollon77:
We try to validate this conformance but we basically never can for an enum, so we fail and noone can use the value. make it Optional too.

grafik


support/models/src/local/FanControl.ts
```diff
@@ -0,0 +1,47 @@
...
+    tag: "cluster",
+    name: "FanControl",
+
+    children: [
+        // Override the conformance of "Medium" (according to spec it's "[Low]" which can never be validated for an enum)
+        {
+            tag: "datatype",
+            name: "FanModeEnum",
+            children: [
+                { tag: "field", name: "Off", conformance: "M", id: 0x0, description: "Fan is off" },
```
lauckhart: Why did you copy all of the fields here?  Should just have the one you're overriding

Apollon77: Valid, tried it but did not worked ... now it does ... and I also know my issue from first try ... updating

PR #3023, opened 2026-01-17
Enhance the number validation for null handling

Apollon77:


lauckhart (approved):
This is just additional tests?

Apollon77:
Yes. There was a discussion in matterbridge and unwanted to be sure and saw it missed the test so unaddressed just to be sure

PR #3024, opened 2026-01-17
Expose session names as strings

Apollon77:
returning a "Diagnostics" might not be a string directly.

Issue reported from @Luligu 
grafik


packages/node/src/behavior/system/sessions/SessionsBehavior.ts
```diff
@@ -30,7 +30,7 @@ export class SessionsBehavior extends Behavior {
 
     #convertToExposedSession(session: NodeSession): SessionsBehavior.Session {
         return {
-            name: session.via,
+            name: `${session.peerAddress.toString()}${Mark.SESSION}${hex.word(session.id)}`,
```
lauckhart: Maybe better to do as `${session.via}` to avoid duplication

packages/protocol/src/session/SessionManager.ts
```diff
@@ -702,7 +704,7 @@ export class SessionManager {
         return [...this.#sessions]
             .filter(session => session.isSecure && !session.isPase)
             .map(session => ({
-                name: session.via,
+                name: `${session.peerAddress.toString()}${Mark.SESSION}${hex.word(session.id)}`,
```
lauckhart: Same here

packages/node/src/behavior/system/sessions/SessionsBehavior.ts
Apollon77: good idea. adjusted

Luligu:
Thanks

PR #3025, opened 2026-01-17
Optimize storing legacy data

Apollon77:
It seems there can be race conditions where the peer is somethow still in, so we ignore deleted peers we just got informed and also check is commissioned

### DIFF
--- a/packages/matter.js/src/MatterController.ts
+++ b/packages/matter.js/src/MatterController.ts
@@ -641,7 +641,7 @@ export class MatterController {
 
     async close() {
         this.#migratedPeerObservers.close();
-        this.#legacyPeerStore?.close();
+        await this.#legacyPeerStore?.close();
         await this.#node?.close();
         this.#clients = undefined;
     }


packages/matter.js/src/MatterController.ts
```diff
@@ -767,7 +767,7 @@ export class MatterController {
         this.#migratedPeerObservers.on(server.peers.deleted, peer => {
             migratedPeers.delete(peer.id);
             logger.info(`Deleted commissioned node ${peer.id} from old format`);
-            peerStore.save().catch(error => logger.warn("Failed to persist legacy commissioned nodes", error));
+            peerStore.save(peer.id).catch(error => logger.warn("Failed to persist legacy commissioned nodes", error));
```
lauckhart: Unawaited promises are likely to create race conditions

Apollon77: Ok you are right ... added logic .. please have a look again

PR #3059, opened 2026-01-21
Enhance Write and Invoke handling according to Matter specs

Apollon77:
This PR refactors the write and Invoke handle to include:
* Allowing chunked write messages for client and server cases, use a iterator interface (basically data to write are prepared internally and then packed into messages as needed. "transaction" handling is tweaked to respect specs.
* Allowing Invoke responses to be split into multiple messages (and potentially do ourselves when client). Any error from server just leads to cancelling waiting for responses.
* When sending Invokes decide if they need to be split into multiple messages or can be sent together

We also enhance the Bridge test device and add testcase RR 1.1 which in the end with the number of endpoints we use will test multiple messages.

### DIFF
--- a/packages/node/src/node/server/InteractionServer.ts
+++ b/packages/node/src/node/server/InteractionServer.ts
@@ -320,8 +320,8 @@ export class InteractionServer implements ProtocolHandler, InteractionRecipient 
         messenger: InteractionServerMessenger,
         message: Message,
     ): Promise<void> {
-        let { suppressResponse, timedRequest, writeRequests, interactionModelRevision, moreChunkedMessages } =
-            writeRequest;
+        let { suppressResponse, writeRequests, moreChunkedMessages } = writeRequest;
+        const { timedRequest, interactionModelRevision } = writeRequest;
         const sessionType = message.packetHeader.sessionType;
 
         logger.info(() => [


packages/node/src/behavior/context/server/RemoteActorContext.ts
```diff
@@ -284,5 +285,10 @@ export namespace RemoteActorContext {
         timed?: boolean;
         fabricFiltered?: boolean;
         message?: Message;
+
+        /**
+         * Optional messenger for testing. If not provided, one will be created from the exchange.
+         */
+        messenger?: InteractionServerMessenger;
```
lauckhart: eww this is ugly for public API.  can we create a hook somewhere instead?  maybe in the `Interactable` to override exchange creation instead?  it's easily accessible for both client and server nodes

packages/node/src/node/server/OnlineServerInteraction.ts
```diff
@@ -35,19 +56,207 @@ export class OnlineServerInteraction implements Interactable<RemoteActorContext.
...
 
-    async write<T extends Write>(request: T, context: RemoteActorContext.Options): WriteResult<T> {
-        return RemoteActorContext(context).act(session => this.#interaction.write(request, session));
+    /**
+     * Process a write-request including chunking messages as needed.
+     * Yields write status results while also sending WriteResponses via the messenger.
+     */
+    async *write(request: Write, context: RemoteActorContext.Options): WriteResult {
+        const { exchange, messenger: existingMessenger } = context;
+        const messenger = existingMessenger ?? new InteractionServerMessenger(exchange);
```
lauckhart: This makes the Interactable abstraction very leaky.  Wire protocol concerns should be in InteractionServer and this class working at the semantic level

packages/node/src/node/server/OnlineServerInteraction.ts
```diff
@@ -35,19 +56,207 @@ export class OnlineServerInteraction implements Interactable<RemoteActorContext.
...
+
+        let { writeRequests, moreChunkedMessages, suppressResponse } = request;
+
+        // Create a writer that persists across all chunks to maintain list operation state
+        // (REPLACE_ALL followed by ADD needs to track previousProcessedAttributePath)
+        let writer: AttributeWriteResponse | undefined;
+
+        // Process chunks until moreChunkedMessages is false
+        while (true) {
```
lauckhart: Since the chunks execute pretty much entirely independently, is there any reason the Interactable needs to know about them?

Vs. having InteractionServer process them and make independent calls to write()...

Or vs. simply having interaction server collect the input chunks into a single `Write` (if matter allows responses after all input chunks are transmitted)?

packages/node/src/node/server/OnlineServerInteraction.ts
```diff
@@ -35,19 +56,207 @@ export class OnlineServerInteraction implements Interactable<RemoteActorContext.
...
+            }
+        }
     }
 
-    async *invoke(request: Invoke, context: RemoteActorContext.Options) {
+    /**
+     * Handle a complete invoke interaction including chunked responses.
+     * Yields invoke results while also sending InvokeResponses via the messenger.
+     */
+    async *invoke(request: Invoke, context: RemoteActorContext.Options): InvokeResult {
```
lauckhart, line 51: Comments for write() apply here as well

packages/protocol/src/action/client/ClientInteraction.ts
```diff
@@ -186,12 +188,13 @@ export class ClientInteraction<
...
+     * Write to node attributes.
      *
-     * Writes with the Matter protocol are generally not atomic, so this method only throws if the entire action fails.
-     * You must check each {@link WriteResult.AttributeStatus} to determine whether individual updates failed.
+     * Yields write status results as they are processed. For large write requests that exceed the maximum
+     * message size, the messenger automatically chunks the writes into multiple messages.
+     * When suppressResponse is true, yields nothing.
      */
-    async write<T extends ClientWrite>(request: T, session?: SessionT): WriteResult<T> {
+    async *write(request: ClientWrite, session?: SessionT): WriteResult {
```
lauckhart: I'm wondering if we should be streaming the write results...  It makes the interface considerably less convenient, and the performance benefit seems minimal given how few/small split writes will be irl...  So a very different profile from reads where this pattern does have clear benefits

packages/protocol/src/action/client/ClientInteraction.ts
```diff
@@ -358,6 +356,63 @@ export class ClientInteraction<
...
 
+    /**
+     * Split commands across multiple parallel invoke-exchanges.
+     * Results are streamed as they arrive from any batch, not buffered.
+     */
+    async *#invokeWithSplitting(
+        request: ClientInvoke,
+        maxPathsPerInvoke: number,
+        session?: SessionT,
+    ): DecodedInvokeResult {
```
lauckhart, line 368: Same comment here as for writes

packages/node/src/node/server/OnlineServerInteraction.ts
Apollon77: Right now the "Remote actor context "is maintained here - and for multi message writes we need to reuse that one to keep the context of the former "last processed attribzte. So it felt better to have this here, thats why also in OnlineServerInteraction and not deeper. And I remembered you wanted to get rid of InteractionServer and replace by this here, so felt more natural.

packages/node/src/behavior/context/server/RemoteActorContext.ts
Apollon77: if you have a better idea ... I did not had any real working that would not have been "meeehh" in other areas :-(

packages/node/src/node/server/OnlineServerInteraction.ts
Apollon77: Same reason: getting rid of InteractionServer was natural for me to put there. IN case of invoke we can easiely move that one layer up again ... easier than for write

packages/protocol/src/action/client/ClientInteraction.ts
Apollon77: Thats now exactly the things where "having same interface for Client and Server interactions becomes funny".

packages/node/src/node/server/OnlineServerInteraction.ts
lauckhart: By "getting rid of InteractionServer" I was envisioning a ProtocolHandler dedicated to translating Messages into Interactable calls...  So converting to/from wire protocol to JS semantics...  So the ideais it would still leave Messenger outside the domain of ServerInteraction, if that makes sense

packages/protocol/src/action/client/ClientInteraction.ts
lauckhart: Good idea to talk through but off the top of my head I don't see a conflict...  In the case of ServerInteraction you deal with chunking inside the ProtocolHandler, and in the case of ClientInteraction you do so inside the Interactable

In either case the decision as to whether to stream chunks asynchronously or assemble into a single synchronous Iterable seems like it's driven by the same memory/latency concerns...  I'm just wondering if it's a significant win to go the async route

packages/node/src/node/server/InteractionServer.ts
```diff
@@ -378,16 +386,120 @@ export class InteractionServer implements ProtocolHandler, InteractionRecipient
...
+                }
+
+                currentBatch = [];
+            };
+
+            for (const request of writeRequests) {
+                const { path } = request;
+                const listIndex = path.listIndex;
+
+                // Check if this is a list ADD that needs BUSY
```
lauckhart: If my understanding is correct and we're only doing this because test suite forces us to emulate CHIP bugs (e.g. constraints on behavior not covered by specification) then it would be good to add a comment to that effect.

(Though I was serious about tracking state in `ChipInteractionServerBugEmulator` to make it extra clear.)

Apollon77: I add.
in fact it is "forbidden to send add without a replace before it ... I think "that then busy is returned" is still not in spec

PR #3065, opened 2026-01-22
Add Invoke batching

Apollon77:
invokes are now batched (when device supports it and non-root-endopoint and not groups) within the same macro-task

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,10 +21,15 @@ The main work (all changes without a GitHub username in brackets in the below li
 - @matter/node
     - Feature: We now allocate node IDs as sequential numbers; the old behavior of randomized node behavior is available if you set `ControllerBehavior` state property `nodeIdAssignment` to `"random"`
     - Enhancement: Also export Matter events via ChangeNotificationService
+    - Enhancement: Added updateFailed event to the OTA behavior
+    - Enhancement: Allows to access the update queue 
     - Adjustment: Ignore invalid VendorIds or DeviceTypeIds when processing MDNS data
     - Fix: Added missing export for ColorControlClient
     - Fix: Prevents duplicate change events for Client behaviors when attributes are Quieter
     - Fix: Ensures to use correct intervals when resuming persisted subscriptions
+    - Fix: Ensures stalled or canceled OTA updates do not block additional tries
+    - Fix: Ensures that the OTA update queue is not blocked and handles stalled/errored entries correctly
+    - Fix: Improves OTA success and failure detections
 
 - @matter/nodejs
     - Fix: Also handle ENETUNREACH as a non-critical network error that triggers the retry logic and MDNS lookup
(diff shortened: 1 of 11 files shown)


packages/node/src/node/client/commands/CommandBatcher.ts
```diff
@@ -0,0 +1,238 @@
...
+
+/**
+ * Batches commands invoked within the same timer tick into a single invoke request.
+ *
+ * Commands invoked "near together" (within the same 0ms timer callback) are collected and
+ * sent as a batched-invoke with unique `commandRef` values for response correlation.
+ *
+ * Extends {@link CommandInvoker} to add batching capability on top of single command invocation.
+ */
+export class CommandBatcher extends CommandInvoker {
```
lauckhart: Would `CommandBatch` be an OK name?

Also, wondering if there's a reason to have a single instance for all batches vs. creating on-demand and integrating close/flush into a single process

packages/node/src/node/client/commands/ClientCommandMethod.ts
```diff
@@ -0,0 +1,35 @@
...
+ * are sent as a single batched invoke request for efficiency.
+ */
+export function ClientCommandMethod(name: string) {
+    // This is our usual hack to give a function a proper name in stack traces
+    const temp = {
+        // The actual implementation
+        async [name](this: ClusterBehavior, fields?: {}) {
+            const node = this.env.get(Node) as ClientNode;
+
+            return (node.interaction as ClientNodeInteraction).invoker.invoke({
```
lauckhart: What's the benefit of implementing this outside of `ClientNodeInteraction` vs. making it an internal implementation detail of `invoke`?

packages/node/src/node/client/commands/CommandInvoker.ts
```diff
@@ -0,0 +1,93 @@
...
+}
+
+/**
+ * Base class for command invocation.
+ * Handles single command invokes via the node's interaction interface.
+ *
+ * This class is used directly for groups (which don't support batching)
+ * and extended by {@link CommandBatcher} for nodes that support batched invokes.
+ */
+export class CommandInvoker {
```
lauckhart: Is there a more general way to implement this so that it only deals with an `Interactable` rather than a `Node`?  Then it wouldn't need to be in `@matter/node`

The benefit of sticking stuff in `@matter/protocol` is that it avoids tight coupling between components...  In order to keep packages well balanced I think we should consider `@matter/protocol` to include both low-level protocol implementation and associated helper logic.  Otherwise I worry `@matter/node` is going to just become another big intertwined `@project-chip/matter.js`.

Same goes for `CommandBatcher`

packages/node/src/node/client/ClientNodeInteraction.ts
```diff
@@ -27,18 +28,58 @@ import {
...
+    #invoker?: CommandInvoker;
 
     constructor(node: ClientNode) {
         this.#node = node;
     }
 
+    /**
+     * The node this interaction is associated with.
+     */
+    protected get node(): ClientNode {
```
lauckhart: I'm a little leary of exposing internal implementation details like this as if you need to ask an `Interactable` what it's node is then you're limiting the flexibility of your logic.

And we already have a way to convey "this logic requires a ClientNodeInteraction + ClientNode", which is just to take the `ClientNode` as an argument instead of `ClientNodeInteraction`

packages/protocol/src/action/client/ClientInteraction.ts
```diff
@@ -417,6 +457,125 @@ export class ClientInteraction<
...
+        const commandRef = this.#allocateCommandRef();
+        const { promise, resolver, rejecter } = createPromise<InvokeResult.DecodedData | undefined>();
+
+        this.#pendingCommands.set(commandRef, {
+            request: { ...cmd, commandRef } as Invoke.ConcreteCommandRequest<any>,
+            resolve: resolver,
+            reject: rejecter,
+        });
+
+        if (!this.#batchTimer?.isRunning) {
```
lauckhart: If a request comes in with a batch time that is shorter than remaining time on the running timer it seems like we should restart the timer with the new batch interval

Apollon77: valid

PR #3070, opened 2026-01-22
Add constrains to the IP datatypes

Apollon77:
... as per spec

### DIFF
--- a/packages/node/src/node/integration/ProtocolService.ts
+++ b/packages/node/src/node/integration/ProtocolService.ts
@@ -610,8 +610,9 @@ function invokeCommand(
             result = Promise.resolve(result)
                 .then(result => {
                     if (isObject(result)) {
-                        logger.debug(
-                            "Invoke result",
+                        logger.info(
+                            "Invoke",
+                            Mark.OUTBOUND,
                             Diagnostic.strong(`${path.toString()}.${command.name}`),
                             session.transaction!.via,
                             Diagnostic.dict(result),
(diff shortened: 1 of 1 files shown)


support/models/src/local/hwadr.ts
```diff
@@ -0,0 +1,13 @@
...
+ * Copyright 2022-2026 Matter.js Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+import { LocalMatter } from "../local.js";
+
+// Add length constraint to the datatype because the spec table misses this
+LocalMatter.children.push({
+    tag: "datatype",
+    name: "hwadr",
+    constraint: "6 to 8", // Formally 6 OR 8, but we do not support this, so min/max is the best we can do
```
lauckhart: If I'm not mistaken I think "6, 8" would in fact limit to 6 or 8 but not 7

Apollon77: Ok, then the Tlv bounds are gone but we said they are not that relevant anymore

support/models/src/local/hwadr.ts
```diff
@@ -0,0 +1,13 @@
...
+ * Copyright 2022-2026 Matter.js Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+import { LocalMatter } from "../local.js";
+
+// Add length constraint to the datatype because the spec table misses this
+LocalMatter.children.push({
+    tag: "datatype",
+    name: "hwadr",
+    constraint: "6, 8", // Formally 6 OR 8, but we do not support this, so min/max is the best we can do
```
lauckhart: Should remove comment here and in ipadr

PR #3075, opened 2026-01-22
DNS-SD client

lauckhart:
This is a replacement for current `MdnsClient` class as well as some other pieces of current networking stack.

Rough architecture:

  * `DnssdNames` provides centralized access to resolved DNS-SD names and manages:

    * The `MdnsSocket` used for communication

    * `DnssdName` is an observable representation of the known state of a single DNS-SD name

    * `DnssdSolicitor` implements active discovery (via query) of name records

  * `IpService` builds on former to provide a higher-level observable representation of a single DNS-SD service.  It manages:

    * `DnssdName` instances associated with the service and any hostnames identified by PTR records

    * `IpServiceResolution` manages the process of locating IP addresses using cached state and `DnssdSolicitor`

    * `IpServiceStatus` tracks connection attempts, records current state of the service, and initiates resolution as necessary

The public interface currently ties directly to our own DNS-SD implementation but I tried to structure so that we can add Bonjour, Avahi etc. implementations later.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,9 @@ The main work (all changes without a GitHub username in brackets in the below li
     - Enhancement: When forcing an MDNS update for an operational device, ensure we also get new IP addresses
     - Fix: Correctly handle multi-message write interactions (server and client) according to Matter specification
     - Fix: Correctly handle multi-message invoke responses (server and client) according to Matter specification
+    - Fix: Ensures to correctly stores and updates the "last known address" from the current session
+    - Fix: Reuse sessions that are pushed by devices themselves, even when we are in a reconnection cycle
+    - Fix: Handle new sessions that devices pushed during a reconnection cycle as successful reconnections in more places
 
 - @project-chip/matter.js
   - Adjustment: Skip the full read before the subscription in the first two reconnection tries
--- a/packages/matter.js/src/MatterController.ts
+++ b/packages/matter.js/src/MatterController.ts
@@ -34,7 +34,6 @@ import {
     isObject,
     Logger,
     MatterError,
-    MaybePromise,
     Minutes,
     ObserverGroup,
     ServerAddress,
@@ -72,7 +71,6 @@ import {
     PeerAddress,
     PeerAddressStore,
     PeerConnectionOptions,
-    PeerDataStore,
     PeerDescriptor,
     PeerSet,
     PhysicalDeviceProperties,
@@ -800,10 +798,6 @@ class CommissionedNodeStore extends PeerAddressStore {
         this.#controllerStore = controllerStore;
         this.#fabric = fabric;
         this.#peers = peers;
-    }
-
-    createNodeStore(_address: PeerAddress): MaybePromise<PeerDataStore | undefined> {
-        throw new ImplementationError("Not implemented");
     }
 
     async loadPeers() {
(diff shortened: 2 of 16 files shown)


packages/general/src/net/dns-sd/DnssdNames.ts
```diff
@@ -0,0 +1,208 @@
...
+    #handleMessage(message: MdnsSocket.Message) {
+        let goodbyesBefore: undefined | Timestamp;
+        for (let record of [...message.answers, ...message.additionalRecords]) {
+            if (this.#filter && !this.#filter(record)) {
+                continue;
+            }
+
+            const name = this.get(record.name);
+            if (record.ttl) {
+                if (record.ttl < this.#minTtl) {
```
Apollon77, line 79: I guess you need the same fix like me


```suggestion
                if (record.ttl  > 0 && record.ttl < this.#minTtl) {
```

packages/general/src/net/dns-sd/IpService.ts
```diff
@@ -0,0 +1,346 @@
...
+     * Known addresses.
+     */
+    get addresses() {
+        return this.#addresses;
+    }
+
+    /**
+     * Values from TXT records.
+     */
+    get kvs() {
```
Apollon77: kvs? keyvaluestore? when it has "txt" why not txt or idea to use it more generic?

packages/general/src/net/dns-sd/IpServiceResolution.ts
```diff
@@ -0,0 +1,134 @@
...
+import type { DnssdSolicitor } from "./DnssdSolicitor.js";
+import { IpService } from "./IpService.js";
+
+/**
+ * Discovers new IP addresses for an {@link IpService}.
+ *
+ * This primarily involves sending queries for SRV records using {@link DnssdSolicitor#discover}.  We also query for
+ * A and AAAA records for any SRV target hostnames for which we do not know IP addresses.
+ *
+ * Runs until aborted or we discover a new IP address (we ignore existing addresses).
```
Apollon77, line 24: this and the sentence before and not somehow interfering ... do we care about known IPs or not?

packages/general/src/net/dns-sd/DnssdNames.ts
lauckhart: this code only runs if `record.ttl > 0`

packages/general/src/net/dns-sd/IpServiceResolution.ts
lauckhart: The first bit refers to the queries we send, the second is when the function returns.  We always query for the PTR so the device should send answers for new IPs as associated records...  But we only explicitly query for known hosts with no IPs...  Just because that's something we explicitly know if we're somehow missing

packages/general/src/net/dns-sd/IpService.ts
lauckhart: Well it's not txt records but the k/v's extracted from them, could rename to just `values` or something...  I was thinking more of the semantics and less of the delivery mechanism

Not intended to serve any other purpose although they are writable

Apollon77: Yeah maybe values is more generic

packages/general/src/net/dns-sd/IpServiceResolution.ts
Apollon77: k

packages/general/src/net/dns-sd/IpService.ts
lauckhart: K I renamed to "parameters"

PR #3102, opened 2026-01-26
Only Subscribe on reconnection

Apollon77:
to optimize the chance to get a new subscription on strange connections. Aka "one big request vs two" (big because of the dataversion filters) ... and while in reconnection state it is ok. Also includes two logging adjustments

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@ The main work (all changes without a GitHub username in brackets in the below li
 
 - @project-chip/matter.js
   - Adjustment: Skip the full read before the subscription in the first two reconnection tries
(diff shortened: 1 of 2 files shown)


packages/general/src/net/ServerAddress.ts
```diff
@@ -78,7 +78,7 @@ export namespace ServerAddress {
         switch (address.type) {
             case "udp":
             case "tcp":
-                return `${address.type}://${address.ip}:${address.port}`;
+                return `${address.type}://[${address.ip}]:${address.port}`;
```
lauckhart: Maybe do this conditionally just for ipv6?

PR #3131, opened 2026-01-30
New logic for establishing client connections

lauckhart:
Most of this will be difficult to separate into separate commits so will likely need a "big bang" merge.  And is maybe best considered "unreviewable" 🙄.

In addition to inevitable bugs this commit does not address legacy codepaths and thus leaves a large amount of deprecated code in place.

Overview:

* Adds `PeerConnection` function that performs Peer (re-)connection.  This implements new connection logic we've discussed and isolates relevant state to a single closure.

* Other relevant bits include `Peer.connect()` and `PeerExchangeProvider`.

* Numerous changes to existing `@matter/protocol` components in support of above.  I tried to keep these changes minimal but the diffs end up pretty horrendous...  Mostly this deals with new options

* Adds additional information to `PeerDescriptor`, improves logic for reacting to descriptor changes, and now uses a partial descriptor to add/configure peers

* Adds optional access to `Interactable` and `NodeProtocol` via `Peer` class.  This makes generic access to Peer features (e.g. reading Basic Information attributes) from @node/protocol more straightforward

* Improves `ActionContext` and constituent interfaces to make them more useful in a client context.  For example you can now do e.g. `toggle(undefined, { abort })` to make an invoke abortable

* Adds `NetworkProfiles` to manage semaphores associated with network access, identified by a string name

* Adds configurability to MRP handling to support new connection logic (namely, maximum backoff time and maximum number of retries)

* Moves some logic generic logic from `@matter/node` to `@matter/protocol`

* Changes lifecycle management of some dependencies of `@matter/node` to align them with the `ServerNode` instead of `ServerNetworkRuntime`

* Various other updates to `@matter/node` involve movement to `@matter/protocol` code paths, but there are also various general improvements to issues I encountered along the way

* Adds `PeerTimingParameters` to control new connection logic

* Makes more operations abortable

* Improves `PeerAddress` interning and `ClientNode#peerAddress` management

* Considerable rework of network error handling to improve naming and consistency

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,10 @@ The main work (all changes without a GitHub username in brackets in the below li
     - Feature: (@adeepn) Added `DclBehavior` for centralized DCL configuration via environment variables (`MATTER_DCL_*`), config files, or programmatic setup
 
 - @matter/protocol
+    - Feature: We have rewritten the logic for establishing operational connections to other nodes.  The new implementation should be faster, more resilient and offers more knobs for tuning
+    - Breaking: Some of the lower-level APIs in @matter/protocol have changed.  This will be transparent to most users
+    - Feature: A new "network profile" feature allows you to tune parallelism and other interaction parameters based on categories including transport type and thread channel
+    - Feature: matter.js now responds immediately to IP changes advertised via MDNS
     - Feature: (@adeepn) `DclConfig` is now an interface with namespace defaults instead of a singleton; `DclClient` accepts `DclConfig` for configurable endpoints
     - Feature: (@adeepn) `DclCertificateService` and `DclOtaUpdateService` accept custom DCL endpoint configuration via options
     - Adjustment: No longer ignore too long incoming Matter messages but still log a warning


packages/node/src/behavior/system/commissioning/CommissioningClient.ts
```diff
@@ -341,30 +348,84 @@ export class CommissioningClient extends Behavior {
...
+        if (addr) {
+            this.#bindPeer(addr);
+            node.lifecycle.commissioned.emit(this.context);
+        } else if (oldAddr) {
+            this.#unbindPeer(oldAddr, true);
+            node.lifecycle.decommissioned.emit(this.context);
+        }
+    }
+
+    #catsChanged(cats?: CaseAuthenticatedTag[]) {
```
Apollon77, line 369: lets add a TODO here because no clue how we should handle that in general ... there is logic missing because such a case should invalidate the current session, or?

packages/node/src/behavior/system/network/NetworkClient.ts
```diff
@@ -39,36 +39,42 @@ export class NetworkClient extends NetworkBehavior {
...
-                        RemoteDescriptor.fromLongForm(this.#node.state.commissioning),
-                    );
+                    peerSet.addKnownPeer({
+                        address: peerAddress,
+                        operationalAddress,
+                        discoveryData: RemoteDescriptor.fromLongForm(this.#node.state.commissioning),
+                    });
                 }
             }
-            if (!this.#node.lifecycle.isCommissioned) {
```
Apollon77, line 51: is that handled?

packages/node/src/behavior/system/subscriptions/SubscriptionsServer.ts
```diff
@@ -218,26 +217,20 @@ export class SubscriptionsBehavior extends Behavior {
...
-                        `Failed to connect to ${peerAddress}`,
-                        error instanceof MatterError ? error.message : error,
-                    );
-                    continue;
-                }
+
+            const peer = peers.addKnownPeer({ address: peerAddress, operationalAddress });
+            let session;
+            try {
+                session = await peer.connect({ connectionTimeout: REESTABLISH_SUBSCRIPTIONS_TIMEOUT });
```
Apollon77, line 224: does that return "ok" when we already have a session?

packages/protocol/src/peer/PeerConnection.ts
```diff
@@ -0,0 +1,463 @@
...
+ * - The connection to the fallback address aborts if the fallback address is not discovered
+ *
+ * - Attempts continue until the address expires or connects successfully
+ *
+ * - We configure MRP to run indefinitely for each attempt with a configurable max delay between messages
+ *
+ * - Starting a new attempt does not cancel previously running attempts; we thus rely on the MRP retransmission window
+ *   to ensure we are sending a reasonable number of packets
+ *
+ * - Once a session is established, any outstanding attempts abort and the function returns
```
Apollon77, line 74: maybe already when we get a Sigma2 or Sigma2Resume response? I know session might still fail bt then reasons are different and the IP "is good" where we got that

packages/protocol/src/peer/PeerConnection.ts
```diff
@@ -0,0 +1,463 @@
...
+                `Connection error (retry in ${Duration.format(context.timing.delayAfterNetworkError)}):`,
+                Diagnostic.errorMessage(e),
+            );
+            delay = context.timing.delayAfterNetworkError;
+        } else if (e instanceof ChannelStatusResponseError) {
+            if (
+                e.protocolStatusCode === SecureChannelStatusCode.NoSharedTrustRoots &&
+                (await context.sessions.deleteResumptionRecord(peer.address))
+            ) {
+                logger.error(
```
Apollon77: i guess warn is ok, or? kind of "I retry and stuff can solve it I would not log error" ... but nit

packages/protocol/src/peer/PeerExchangeProvider.ts
```diff
@@ -0,0 +1,78 @@
...
+    readonly channelType = ChannelType.UDP;
+
+    override async initiateExchange(options?: NewExchangeOptions): Promise<MessageExchange> {
+        const abort = options?.abort;
+
+        // Connections grab their own network slot so connect before obtaining our own
+        const session = await this.#peer.connect(options);
+
+        const network = this.#context.networks.select(this.#peer, options?.network);
+        const slot = await network.semaphore.obtainSlot(abort);
```
Apollon77: can we get some kinda of network info somehow into the log also at least on level of exchange creation or logging ... would potentially be easier for debugging

packages/protocol/src/session/case/CaseClient.ts
```diff
@@ -217,49 +238,51 @@ export class CaseClient {
...
 
             // Create a secure session. Configured CATs take precedence over resumption record ones
             const sessionCaseAuthenticatedTags = caseAuthenticatedTags ?? resumptionRecord?.caseAuthenticatedTags;
             const secureSessionSalt = Bytes.concat(
                 operationalIdentityProtectionKey,
-                await crypto.computeHash([sigma1Bytes, sigma2Bytes, sigma3Bytes]),
+                await abort.attempt(crypto.computeHash([sigma1Bytes, sigma2Bytes, sigma3Bytes])),
+            );
+            secureSession = await abort.attempt(
+                this.#sessions.createSecureSession({
```
Apollon77, line 301: that is just adding stuff for us? Why the "attempt"?

packages/node/src/behavior/system/commissioning/CommissioningClient.ts
lauckhart: My intention wasn't to make it truly dynamic, just to make it so that the changes are picked up if you configure after instantiating the node.  It's effectively the same as the old behavior, which is to say that changing it only affects new sessions.  It's just that we need to sync the CATs from the ClientNode to the Peer now

packages/node/src/behavior/system/network/NetworkClient.ts
lauckhart: I'm not sure why it had the "!isComissioned" test before but if we have basicInformation we always read capabilityMinima from that

packages/node/src/behavior/system/subscriptions/SubscriptionsServer.ts
lauckhart: If a session already exists it returns the newest one

packages/protocol/src/peer/PeerConnection.ts
lauckhart: well we can't abort altogether because that would abort the attempt that received the sigma2...  I'm not sure if it's worth it to abort other attempts when we get sigma2, would add some complexity and is kind of an edge case, do you think the optimization is worth it?

packages/protocol/src/peer/PeerConnection.ts
lauckhart: OK, I checked and it was info before, any are fine by me

packages/protocol/src/session/case/CaseClient.ts
lauckhart: so that we stop if there's an abort during the crypto operation...  I generally try to detect abort during every async operation no matter how minor

packages/protocol/src/peer/PeerConnection.ts
Apollon77: not sure if too much edge case ... if a device has multiple ips and we try then (with the 5s delay) then it still might be that first takes longer but both IPs are reachable .. so we would start two sessions in parallel and one win ... and one is basically stale, right?

lauckhart: Yes, that's how it'd work

So as optimization it would require following:

1. For device with multiple IPs,
2. Where device is slow to perform case computations or network is slow,
3. Transmission or response computation of extra sigma3 could reduce qos in some way (assuming it's not too expensive on our side), and
4. Faster completion of case negotiation cannot result in a preferable session (e.g faster or less congested network medium or preferable device configuration)

Last two are kind of a balance and I can see that we might conclude this criteria is satisfied, but maybe worth some experimental feedback before committing?

packages/protocol/src/peer/PeerExchangeProvider.ts
lauckhart: You mean specifically relating to semaphore slots held or something more general?

For semaphore specifically dedicated messages seem noisy but maybe we could log "slow acquisition" similar to how we handle transactions

packages/protocol/src/peer/PeerConnection.ts
Apollon77: yeah lets see the feedback

packages/protocol/src/peer/PeerExchangeProvider.ts
Apollon77: Mainly thinking about the network config "name/type" or relevant limits or such

lauckhart: Oh, maybe include relevant details in the operational log message then?

PR #3165, opened 2026-02-04
Extract DCL URLs to configurable DclConfig

adeepn:
Background:
The Distributed Compliance Ledger (DCL) operates as a multi-node, decentralized network. In real-world deployments, direct access to a specific public DCL node cannot always be assumed: individual nodes may be temporarily unavailable, unreachable due to network policies, or intentionally isolated.

At the same time, the DCL architecture explicitly allows clients to retrieve the same data from any reachable node in the network. This enables vendors and integrators to deploy their own observer or light proxy nodes — for example, to serve regional traffic, improve latency and reliability, or support Matter device operation within restricted or fully offline environments (corporate networks, factories, labs, or test infrastructures).

Hardcoding DCL and related service endpoints makes such deployment models difficult or impossible. Allowing the DCL node and associated service endpoints to be configured at runtime via environment variables provides the necessary flexibility to support alternative network topologies, private observer nodes, and custom proxy setups without modifying application code.

PR:
Move hardcoded DCL and GitHub URLs to centralized DclConfig module with environment variable support for runtime configuration.

Environment variables:
- MATTER_DCL_PRODUCTION_URL - Production DCL API endpoint
- MATTER_DCL_TEST_URL - Test DCL API endpoint
- MATTER_DCL_GITHUB_OWNER - GitHub owner for dev certificates
- MATTER_DCL_GITHUB_REPO - GitHub repository name
- MATTER_DCL_GITHUB_BRANCH - GitHub branch
- MATTER_DCL_GITHUB_PATH - Path to certificates in repository

### DIFF
--- a/packages/protocol/src/dcl/DclClient.ts
+++ b/packages/protocol/src/dcl/DclClient.ts
@@ -37,8 +37,8 @@ export class MatterDclResponseError extends MatterDclError {
 export class DclClient {
     #baseUrl: string;
 
-    constructor(private readonly production: boolean = true) {
-        this.#baseUrl = this.production ? DclConfig.dcl.productionUrl : DclConfig.dcl.testUrl;
+    constructor(config: DclConfig = DclConfig.production) {
+        this.#baseUrl = config.url;
     }
 
     async #fetchPaginatedJson<ItemT>(
--- a/packages/protocol/src/dcl/DclConfig.ts
+++ b/packages/protocol/src/dcl/DclConfig.ts
@@ -1,56 +1,49 @@
 /**
  * @license
  * Copyright 2026 Matter.js Authors
- * Copyright 2026 Viacheslav Bocharov <adeep@baodeep.com>
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import { Environment } from "#general";
+/**
+ * Configuration for a DCL (Distributed Compliance Ledger) API endpoint.
+ *
+ * Swagger docs: https://zigbee-alliance.github.io/distributed-compliance-ledger/#/
+ */
+export interface DclConfig {
+    url: string;
+}
+
+export namespace DclConfig {
+    /** Default production DCL endpoint (CSA official) */
+    export const production: DclConfig = {
+        url: "https://on.dcl.csa-iot.org",
+    };
+
+    /** Default test/development DCL endpoint */
+    export const test: DclConfig = {
+        url: "https://on.test-net.dcl.csa-iot.org",
+    };
+}
 
 /**
- * Centralized URL configuration for DCL-related services.
+ * Configuration for GitHub certificate repository.
  *
- * Uses matter.js Environment.vars configuration system which supports:
- * - Environment variables (MATTER_* prefix)
- * - Config file (config.json)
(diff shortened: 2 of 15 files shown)


packages/protocol/src/dcl/DclClient.ts
```diff
@@ -36,12 +33,12 @@ export class MatterDclResponseError extends MatterDclError {
...
 }
 
-/** A client clas to use "fetch" to get REST DAta from DCL (Decentraland) */
+/** A client class to use "fetch" to get REST data from DCL (Distributed Compliance Ledger) */
 export class DclClient {
     #baseUrl: string;
 
     constructor(private readonly production: boolean = true) {
-        this.#baseUrl = this.production ? DCL_PRODUCTION_URL : DCL_TEST_URL;
+        this.#baseUrl = this.production ? DclConfig.dcl.productionUrl : DclConfig.dcl.testUrl;
```
lauckhart: Given that you're shooting for full configurability instead of just production/test as an option I think it would make sense here to have the input be a `DclConfig` interface that could default to e.g. `DclConfig.production` which would be the default production values

packages/protocol/src/dcl/DclConfig.ts
```diff
@@ -0,0 +1,56 @@
...
+ *
+ * Configuration variables (with their environment variable equivalents):
+ * - dcl.production.url (MATTER_DCL_PRODUCTION_URL) - Production DCL API endpoint
+ * - dcl.test.url (MATTER_DCL_TEST_URL) - Test DCL API endpoint
+ * - dcl.github.owner (MATTER_DCL_GITHUB_OWNER) - GitHub owner for development certificates
+ * - dcl.github.repo (MATTER_DCL_GITHUB_REPO) - GitHub repository name
+ * - dcl.github.branch (MATTER_DCL_GITHUB_BRANCH) - GitHub branch
+ * - dcl.github.path (MATTER_DCL_GITHUB_PATH) - Path to certificates directory in repository
+ */
+export const DclConfig = {
```
lauckhart: We generally try to avoid singletons although vs. hard-coded this seems like an improvement

Ideally this would be an interface, with a corresponding namespace that has e.g. "production" and "test" values constants that provide defaults

packages/protocol/src/dcl/DclConfig.ts
```diff
@@ -0,0 +1,56 @@
...
+    /** DCL API endpoints */
+    dcl: {
+        /** Production DCL URL (CSA official) */
+        get productionUrl(): string {
+            return Environment.default.vars.get("dcl.production.url", "https://on.dcl.csa-iot.org");
```
lauckhart: Rather than Environment.default it would be good to use the Node's environment.

If you're game, the ideal way of doing this is to have e.g. `DclBehavior` that defines these values in its state.  Then matter.js would automatically set the state values from the node's environment variables, but a dev could also set manually when building the node, etc.

If the behavior feels too ambitious you could instead create a `DclConfig` factory function that takes the `Environment` and defaults as a parameter and returns the `DclConfig` instance

Or if you don't want to tackle any of that you could just leave a TODO in this file

Apollon77 (approved):

PR #3176, opened 2026-02-07
Spec parser adjustment and 1.4.2 Reparse

Apollon77:
TBD


packages/model/src/standard/elements/unit-localization.element.ts
```diff
@@ -27,7 +27,7 @@ export const UnitLocalization = Cluster(
 
     Attribute(
         {
-            name: "SupportedTemperatureUnits", id: 0x1, type: "list", access: "R V", conformance: "P, TEMP",
+            name: "SupportedTemperatureUnits", id: 0x1, type: "list", access: "R V", conformance: "TEMP",
```
lauckhart, line 30: What happened to the "P"?

packages/model/src/standard/elements/water-heater-mode.element.ts
```diff
@@ -25,7 +25,13 @@ export const WaterHeaterMode = Cluster(
     Attribute({ name: "CurrentMode", id: 0x1 }),
     Attribute({ name: "StartUpMode", id: 0x2, conformance: "X" }),
     Attribute({ name: "OnMode", id: 0x3, conformance: "X" }),
-    Datatype({ name: "ModeOptionStruct", type: "ModeOptionStruct" }),
+
+    Datatype(
+        { name: "ModeOptionStruct", type: "struct" },
```
lauckhart: Just noting this should revert

packages/model/src/standard/resources/access-control.resource.ts
```diff
@@ -471,23 +471,17 @@ Resource.add(
                             "\n" +
                             "    ◦ Before expansion:" +
                             "\n" +
-                            "https://company.domain.example/matter/arl?vid=FFF1&pid=1234&MTcu=_" +
-                            "\n" +
-                            "  ◦ After expansion: " +
-                            "    https://company.domain.example/matter/arl?vid=FFF1&pid=1234&MTcu=https%3A%2F%2Fc " +
```
lauckhart, line 477: It looks like the URLs disappeared

packages/model/src/standard/resources/ecosystem-information.resource.ts
```diff
@@ -127,10 +127,12 @@ Resource.add({
...
                     details: "This field shall indicate the timestamp of when the UniqueLocationIDs was last modified." +
                         "\n" +
-                        "NOTE If multiple server instances update the UniqueLocationIDs field at the same time, it is " +
-                        "possible one of the updates will be missed. This is considered an acceptable limitation to reduce " +
-                        "the complexity of the design. Since this is meant to be provided from user input, it is unlikely " +
-                        "these signals would be happening at one time."
+                        "> [!NOTE]" +
+                        "\n" +
+                        "> If multiple server instances update the UniqueLocationIDs field at the same time, it is possible " +
+                        "  one of the updates will be missed. This is considered an acceptable limitation to reduce the " +
```
lauckhart: There's extra spaces at the beginning of the lines

packages/model/src/standard/resources/operational-state.resource.ts
```diff
@@ -199,9 +199,7 @@ Resource.add({
                 "  • The device shall respond with an OperationalCommandResponse command with an ErrorStateID of " +
                 "    NoError." +
                 "\n" +
-                "The following table defines the compatibility of this cluster’s states with the Pause command." +
-                "\n" +
-                "### Table 3. Pause Compatibility"
+                "The following table defines the compatibility of this cluster’s states with the Pause command."
```
lauckhart: The table caption disappeared which makes sense, maybe add a deletion to take this line out (pattern "The following table.*:$" or something)

packages/model/src/standard/resources/thermostat-cluster.resource.ts
```diff
@@ -540,7 +549,7 @@ Resource.add(
                     "\n" +
                     "⇒ OccupiedHeatingSetpoint - Calculated Local Temperature ≥? EmergencyHeatDelta" +
                     "\n" +
-                    "⇒ 16°C - 10°C ≥? 2°C" +
+                    "  • ⇒ 16°C - 10°C ≥? 2°C" +
```
lauckhart: Extraneous bullet?

support/codegen/src/mom/spec/add-documentation.ts
```diff
@@ -183,26 +188,65 @@ export function addDocumentation(target: { details?: string }, definition: HtmlR
             }
         }
 
+        // Convert numeric superscripts before text extraction (safe for prose, not for constraint columns)
```
lauckhart, line 176: Do we properly handle 2^62 in constraints now?  I had to hack as 262 before

support/codegen/src/mom/spec/add-documentation.ts
```diff
@@ -183,26 +188,65 @@ export function addDocumentation(target: { details?: string }, definition: HtmlR
...
+        convertSuperscripts(p);
+
         // Extract text
         let text = Str(p);
 
+        // For Asciidoctor list items, add numbered/bulleted marker.  The FormattedText serializer handles
+        // indentation based on marker type, so we use depth-appropriate markers matching its Bullets array
```
lauckhart, line 183: We could also convert to markdown style bullets...  I just did it this way because the pdf extraction did not capture indentation

support/codegen/src/mom/spec/html-translators.ts
```diff
@@ -48,17 +57,40 @@ export const Str = (el: HTMLElement) => {
...
 
             // Convert "foo- bar" to "foo-bar"
             .replace(/([a-z]-) ([a-z])/g, "$1$2")
     );
 };
 
+/**
+ * Convert numeric superscripts to ^N notation in an element (e.g. 10<sup>6</sup> → 10^6, m<sup>3</sup> → m^3).
+ * Skips ordinals (st, nd, rd, th) and footnote markers.  Only use on prose/description elements, not on constraint
+ * or type cells where ^ would break parsing.
```
lauckhart, line 76: Should add ^ to constraint syntax...  I didn't add because I couldn't extract previously

support/codegen/src/mom/spec/scan-document.ts
```diff
@@ -175,11 +195,55 @@ export function* scanDocument(docRef: HtmlReference) {
...
+                        currentRef.prose.push(element);
+                    }
+                    break;
+
                 case "TABLE":
                     if (!currentRef) {
                         break;
                     }
 
+                    // Extract content from Asciidoctor admonition blocks (notes, warnings, etc.)
```
lauckhart: Rather than deformating to match Adobe might be better to modify the html parser to support admonition directly. As is it ends up adding two unnecessary hacks (HTML -> plaintext, then plaintext -> markdown)

support/codegen/src/mom/spec/spec-file.ts
```diff
@@ -105,11 +105,29 @@ export class SpecFile {
             throw new Error(`No HTML files in ${path}`);
         }
 
+        // Sort to ensure stable processing order: cluster, core, device, namespace.  This matters because
+        // global elements from core must appear before device types in the output
```
lauckhart, line 109: Hmm why is that?  Maybe we could fix it?

support/codegen/src/util/words.ts
```diff
@@ -27,5 +27,14 @@ Words.add("led");
 Words.add("url");
 Words.add("wifi");
 
+// Networking/technical terms used in Matter specification
```
lauckhart: Is it possible we could kill the word list entirely?  IIRC it was only here to work around accidental word splits in pdf output

packages/model/src/standard/elements/unit-localization.element.ts
Apollon77: This was a fix "post 1.4.2 official publish and was updated.

Apollon77: This change is intended

support/codegen/src/mom/spec/spec-file.ts
Apollon77: The names in the "new" directory are different and so the order of files he adds and so the order of processing differs a bit which moves e.g. the global elements around in the results. This is to ensure a defined order of processing that matches the old filenaming.

packages/model/src/standard/elements/water-heater-mode.element.ts
Apollon77: Done

support/codegen/src/mom/spec/add-documentation.ts
Apollon77: adjusted in html-translators

packages/model/src/standard/resources/access-control.resource.ts
Apollon77: fixed

packages/model/src/standard/resources/ecosystem-information.resource.ts
Apollon77: Thats intention because of the "> "

packages/model/src/standard/resources/thermostat-cluster.resource.ts
Apollon77: fixed

packages/model/src/standard/resources/operational-state.resource.ts
Apollon77: done

support/codegen/src/mom/spec/add-documentation.ts
Apollon77: done

support/codegen/src/mom/spec/html-translators.ts
Apollon77: done

support/codegen/src/util/words.ts
Apollon77: done

support/codegen/src/mom/spec/scan-document.ts
Apollon77: done

packages/model/src/standard/resources/ecosystem-information.resource.ts
lauckhart: These are single-line strings wrapped for source code readability (note there's " " at the end rather than "\n").  The string will end up like "...it is possible<3 spaces>one of the updates will be..."

The wrapping based on the ">" happens when wrapping these lines for comments.

Apollon77: addressed

PR #3179, opened 2026-02-08
Model Adjustments for Spec 1.5

Apollon77:
Contains handling of
* more mathematical formulas in constraints
  * max (2^62) - 1 
  * 0 to segmentDuration / 2
  * -1 * panMax to panMax
* referencing Cluster.Fields in conformance (SolicitOffer.VideoStreamID)
*  TriggerType == Motion 
*  RequiresEncodedPixels == True (case will be converted when scraping to "true")
*  Precedence: == higher than |  to allow "TriggerType == Command | TriggerType == Motion"
* Add new Model element for ConditionRequirement


packages/model/src/aspects/Conformance.ts
```diff
@@ -646,7 +667,7 @@ function ParsedAst(conformance: Conformance, definition: string): Conformance.As
 
 namespace Parser {
     // Highest precedence first
-    export const BinaryOperatorPrecedence = [[">", "<", ">=", "<="], ["&"], ["|", "^"], ["==", "!="]];
+    export const BinaryOperatorPrecedence = [["==", "!=", ">", "<", ">=", "<="], ["&"], ["|", "^"]];
```
lauckhart: Is this correct?  Usually comparisons have higher precedence than equality

packages/model/src/aspects/Constraint.ts
```diff
@@ -591,6 +611,8 @@ namespace Parser {
             while (
                 tokens.token?.type === "+" ||
                 tokens.token?.type === "-" ||
+                tokens.token?.type === "*" ||
+                tokens.token?.type === "/" ||
                 tokens.token?.type === "." ||
                 tokens.token?.type === "^"
             ) {
```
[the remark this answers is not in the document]
lauckhart: I agree this may be necessary now

packages/model/src/logic/definition-validation/RequirementValidator.ts
```diff
@@ -52,6 +52,15 @@ ModelValidator.validators[RequirementElement.Tag] = class RequirementValidator e
             }
         }
 
+        // For condition requirements, temporarily clear the type before base validation since their type
+        // is a cross-reference (e.g. "RootNode.AclExtensionCond") not a resolvable model type
+        const savedType = this.model.type;
```
lauckhart: Qualified model types are resolvable (implemented in ModelTraversal#findQualifiedType) and should still be validated

packages/model/src/logic/definition-validation/ConditionValidator.ts
```diff
@@ -0,0 +1,18 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { ConditionElement } from "../../elements/index.js";
+import { ConditionModel } from "../../models/index.js";
+import { ModelValidator } from "./ModelValidator.js";
+
+ModelValidator.validators[ConditionElement.Tag] = class ConditionValidator extends ModelValidator<ConditionModel> {
+    override validate() {
+        // TODO we need to validate as needed. Likely here we can not do that much because we
```
lauckhart: This comment appears to be confusing runtime validation (validating data) vs. model validation (validating the metadata).  I think it can just be deleted as this performs the metadata validation that's possible

packages/model/src/logic/definition-validation/ValueValidator.ts
```diff
@@ -34,6 +34,20 @@ export class ValueValidator<T extends ValueModel> extends ModelValidator<T> {
...
+                const [parentName, fieldName] = name.split(".", 2);
+                const cluster = this.model.owner(ClusterModel);
+                const camelParent = camelize(parentName, true);
+                const parent = cluster?.children.find(
+                    c => c.name === camelParent || c.name === parentName || camelParent.endsWith(c.name),
+                );
+                if (parent) {
+                    return parent.member(camelize(fieldName, true));
+                }
+            }
```
lauckhart: Should this fall through if the parent isn't found?

packages/model/src/logic/definition-validation/ValueValidator.ts
```diff
@@ -34,6 +34,20 @@ export class ValueValidator<T extends ValueModel> extends ModelValidator<T> {
                 return cluster?.features.find(f => f.name === name);
             }
 
+            // Dot-field references like "SolicitOffer.VideoStreamID" — resolve the command then the field.
+            // The command name may include a cluster prefix (e.g. "WebRTCProvideOffer" for "ProvideOffer")
+            if (name.includes(".")) {
+                const [parentName, fieldName] = name.split(".", 2);
```
lauckhart: Qualified names can have multiple levels of nesting, so I don't think splitting in 2 is the right thing to do.  But also, `member` should handle qualified types.  So I think this block is incorrect but also hopefully unnecessary.

packages/model/test/aspects/ConformanceTest.ts
```diff
@@ -27,10 +30,21 @@ const TEST_DEFINITIONS = [
     "[LF & PA_LF & ABS]",
     "!USR & (PIN | RID | FGP)",
     "OperationalStateID >= 128 & OperationalStateID <= 191",
+
+    // Dot-field references (field presence in another command)
```
lauckhart: I'm not sure why it's referring to "command" here...  Conformance is a general concept that doesn't apply just to commands

packages/model/test/aspects/ConformanceTest.ts
lauckhart, line 1: I think I saw a condition used as a conformance expression, is that handled anywhere?  Or future TODO?

lauckhart: I ask because the model will fail validation if there are names that do not resolve

packages/model/src/aspects/Conformance.ts
```diff
@@ -670,8 +670,8 @@ function ParsedAst(conformance: Conformance, definition: string): Conformance.As
 }
 
 namespace Parser {
-    // Highest precedence first
-    export const BinaryOperatorPrecedence = [["==", "!=", ">", "<", ">=", "<="], ["&"], ["|", "^"]];
+    // Highest precedence first: comparisons bind tighter than equality
```
lauckhart: Don't think the comment needs to describe the new part specifically?

support/models/src/v1.5/spec.ts
lauckhart, line 1: Is this added by mistake?

packages/model/src/logic/ModelTraversal.ts
```diff
@@ -456,6 +456,20 @@ export class ModelTraversal {
      */
     findMember(scope: Model | undefined, key: Children.Selector, allowedTags: ElementTag[]): Model | undefined {
         return this.operation(() => {
+            // Qualified names like "SolicitOffer.VideoStreamID" — walk intermediate segments
+            // without tag restriction, apply allowedTags only on the final segment
```
lauckhart: Interesting, this is a maybe a literal interpretation of my review comment?  I meant more that member lookup for `base` already uses qualified lookup for the `type` field.  But having a standalone "resolve" method that does qualified resolution would make sense...  Though I probably wouldn't call it "findMember" then.

Moving the logic that finds the base here, to findMember, isn't a terrible idea, but it seems like one way or another the logic should merge so we only implement once.  And for `ValueValidator` to use it it'd need to preload the "ignore" list with the node in question so it doesn't resolve to itself.

Alternately `ValueValidator` could maybe use `base` to validate the location referenced in `type`.

packages/model/src/logic/ModelTraversal.ts
```diff
@@ -456,6 +456,20 @@ export class ModelTraversal {
...
+            if (typeof key === "string" && key.includes(".")) {
+                const parts = key.split(".");
+                for (let i = 0; i < parts.length - 1 && scope; i++) {
+                    // Try exact match first, then suffix match for cluster-prefixed names
+                    // (e.g. "WebRTCProvideOffer" matching command "ProvideOffer")
```
lauckhart: What is going on here?  I don't think something called "oaisjeflsijdflkdsjfProvideOffer" should match something called "ProvideOffer" (or that something called "8" should match something called "uint8" or "int8" depending on which order they are in).

I'm not sure of the context of where this was created but if the spec is calling something "WebRTCProviderOffer" and referencing it using "ProvideOffer" then it sounds like maybe either the latter is an error or we should remove the prefix in the former and consider it an override.

support/models/src/v1.5/spec.ts
Apollon77: ahh yes was me ... i remove again

packages/model/src/logic/ModelTraversal.ts
Apollon77:
```
endsWith suffix matching removed:                                                                                                                               
  - The spec used WebRTCProvideOffer.VideoStreamID in conformance but the command is named ProvideOffer
  - Instead of loose endsWith matching in findMember (which would match any string ending in ProvideOffer), added an explicit conformance repair in aspect-repairs.ts that    
  maps WebRTCProvideOffer → ProvideOffer                                                      
  - Fixed the two conformance values in spec.ts directly (what generate-spec would produce)
  - Simplified findMember to use only exact child matching via selectAll
```

packages/model/src/logic/ModelTraversal.ts
Apollon77:
```
findMember architecture:
  - With the endsWith hack removed, findMember's qualified name handling is now just a simple walk of intermediate segments with exact matching, then the normal
  inheritance-based member search on the final segment
  - This is a different concern from findQualifiedType (which searches up through scopes to find a type path root) — findMember starts at a known scope and walks down through
   named children
```

packages/model/src/aspects/Conformance.ts
Apollon77:
```
Fixed. Changed "Highest precedence first: comparisons bind tighter than equality" to just "Highest precedence first". 
```

packages/model/test/aspects/ConformanceTest.ts
Apollon77:
```
 (conditions as conformance expressions) — This is already tracked as a TODO in RequirementValidator.ts:55-58. Condition names from device types (like LI from      
  Lighting) aren't validated yet because it requires resolving conditions from the device type hierarchy, case-insensitive matching, and cluster feature resolution. Currently
   validation is skipped for requirement conformance references so it won't fail, just won't catch errors. You could reply to lauckhart pointing to that TODO.
```

packages/model/src/logic/ModelTraversal.ts
lauckhart: OK, are we discussing use of qualified names in a context other than the `type` field of a model?  A concrete example of how this is used would be helpful

I can see that "member" search may have different requirements from a "type" search, and that having shared implementation may be unnecessary or undesirable from a complexity standpoint

The remaining problem here though is that I can't think of a place where we would want to resolve the non-leaf segments of a qualified name without considering names that are present due to inheritance. The search for non-leaf segments implemented here is (as you describe) only considering  children that are explicitly present directly in `children`

packages/model/src/logic/ModelTraversal.ts
Apollon77: This handles conformance expressions from the WebRTC clusters in spec 1.5 that use qualified names to cross-reference fields in sibling commands. There are 4 cases:

- `SolicitOffer.VideoStreamID` (WebRtcTransportRequestor SolicitOfferResponse.VideoStreamId)
- `SolicitOffer.AudioStreamID` (WebRtcTransportRequestor SolicitOfferResponse.AudioStreamId)
- `ProvideOffer.VideoStreamID` (WebRtcTransportProvider ProvideOfferResponse.VideoStreamId)
- `ProvideOffer.AudioStreamID` (WebRtcTransportProvider ProvideOfferResponse.AudioStreamId)

So `SolicitOffer` and `ProvideOffer` are actual command names within the same cluster, and `VideoStreamID`/`AudioStreamID` are fields within those commands. It's not a suffix/substring match situation — `selectAll` does exact name matching here, just without tag restriction on intermediate segments.

That said, this does duplicate logic that `findQualifiedType` already handles. Would you prefer we:

1. Reuse/adapt `findQualifiedType` instead of reimplementing the walk in `findMember`?
2. Handle it purely in the `ValueValidator` conformance resolver without changing `findMember` at all (e.g. using `base`-style resolution)?
3. Something else?

Apollon77: Correction to my previous reply: I was looking at the working tree which already has the `endsWith` fallback removed. The commit you reviewed still has:

```typescript
scope =
    scope.children.selectAll(parts[i])[0] ??
    scope.children.find(c => parts[i].endsWith(c.name));
```

Your concern about the suffix matching is valid — that's already been removed locally. The remaining question is whether the qualified name walk itself (splitting on "." and resolving segments) belongs in `findMember` or should reuse existing infrastructure like `findQualifiedType`.

packages/model/src/logic/ModelTraversal.ts
Apollon77: Yes, this is about conformance expressions, not the `type` field. Concrete example from the WebRTC clusters in spec 1.5:

In **WebRtcTransportRequestor**, the `SolicitOfferResponse` command has fields with conformance referencing fields in a sibling command:

```
Command: SolicitOfferResponse
  Field: VideoStreamId, conformance: "SolicitOffer.VideoStreamID"
  Field: AudioStreamId, conformance: "SolicitOffer.AudioStreamID"
```

Similarly in **WebRtcTransportProvider**:

```
Command: ProvideOfferResponse
  Field: VideoStreamId, conformance: "ProvideOffer.VideoStreamID"
  Field: AudioStreamId, conformance: "ProvideOffer.AudioStreamID"
```

The conformance validator resolves these references by calling `model.member("SolicitOffer.VideoStreamID")` from the cluster scope, which needs to find the `SolicitOffer` command, then the `VideoStreamID` field within it.

Regarding your point about inheritance for non-leaf segments: you're right that the current implementation only looks at direct `children` for intermediate segments. In the WebRTC case this works because the commands are direct children of the cluster. But in the general case, if a command were inherited from a base cluster, this wouldn't find it.

Should the intermediate segment walk also use `findBase` to search inherited scope, similar to how the final segment already does via the `while (scope) { ... scope = this.findBase(scope); }` loop?

lauckhart: Yes

However, the semantics of validation in this case are rather specialized in that they mean "this field is mandatory in the response if the named field is present in the request".  I'm unsure if the feature added here is applicable for this use case.

For **model (metadata) validation**, the name resolution function needs to understand "for response commands, the request command should be in scope when resolving names".  Commands other than the request and response should *not* be in scope because that is nonsensical.

For **runtime (data) validation**, the name resolution function then needs to resolve the name of the request to the input request object. And the compiled validation function **or** the name resolver function need to understand compound names.

Apollon77: Ok, taken out as discussed, see final comment

Apollon77:
The 1.5 spec scraping (generate-spec)shows these errors:

```
2026-02-15 19:32:46.826 INFO   ValidateModel        ⎸ Validating matter model
2026-02-15 19:32:47.226 ERROR  ValidateModel        *** Validation error summary ***
2026-02-15 19:32:47.226 ERROR  ValidateModel        enum16 with no children code: CHILDLESS_ENUM xref: cluster§1.9.6.2 src: ModeSelect.state.standardNamespace
2026-02-15 19:32:47.226 ERROR  ValidateModel        Value "0" is not a object code: INVALID_VALUE xref: cluster§9.12.5.16.5 src: CommodityTariff.TariffComponentStruct.peakPeriod
2026-02-15 19:32:47.226 ERROR  ValidateModel        Value "0" is not a object code: INVALID_VALUE xref: cluster§9.12.5.16.6 src: CommodityTariff.TariffComponentStruct.powerThreshold
2026-02-15 19:32:47.226 ERROR  ValidateModel        Type TLSCAID does not resolve code: TYPE_UNKNOWN xref: cluster§11.4.5.3.4 src: ICEServerStruct.caid
2026-02-15 19:32:47.226 ERROR  ValidateModel        Type VideoStreamID does not resolve code: TYPE_UNKNOWN xref: cluster§11.4.5.5.5 src: WebRTCSessionStruct.videoStreamId
2026-02-15 19:32:47.226 ERROR  ValidateModel        Type AudioStreamID does not resolve code: TYPE_UNKNOWN xref: cluster§11.4.5.5.6 src: WebRTCSessionStruct.audioStreamId
2026-02-15 19:32:47.226 ERROR  ValidateModel        Conformance name reference "SolicitOffer.VideoStreamID" does not resolve code: UNRESOLVED_CONFORMANCE_NAME xref: cluster§11.5.7.2.3 src: WebRtcTransportProvider.solicitOfferResponse.videoStreamId
2026-02-15 19:32:47.226 ERROR  ValidateModel        Conformance name reference "SolicitOffer.AudioStreamID" does not resolve code: UNRESOLVED_CONFORMANCE_NAME xref: cluster§11.5.7.2.4 src: WebRtcTransportProvider.solicitOfferResponse.audioStreamId
2026-02-15 19:32:47.226 ERROR  ValidateModel        Conformance name reference "ProvideOffer.VideoStreamID" does not resolve code: UNRESOLVED_CONFORMANCE_NAME xref: cluster§11.5.7.4.2 src: WebRtcTransportProvider.provideOfferResponse.videoStreamId
2026-02-15 19:32:47.226 ERROR  ValidateModel        Conformance name reference "ProvideOffer.AudioStreamID" does not resolve code: UNRESOLVED_CONFORMANCE_NAME xref: cluster§11.5.7.4.3 src: WebRtcTransportProvider.provideOfferResponse.audioStreamId
2026-02-15 19:32:47.226 ERROR  ValidateModel        Error counts by code:
2026-02-15 19:32:47.226 ERROR  ValidateModel        ⎸ CHILDLESS_ENUM: 1
2026-02-15 19:32:47.226 ERROR  ValidateModel        ⎸ INVALID_VALUE: 2
2026-02-15 19:32:47.226 ERROR  ValidateModel        ⎸ TYPE_UNKNOWN: 3
2026-02-15 19:32:47.226 ERROR  ValidateModel        ⎸ UNRESOLVED_CONFORMANCE_NAME: 4
2026-02-15 19:32:47.226 ERROR  ValidateModel        *** Total 10 validation errors ***
2026-02-15 19:32:47.226 ERROR  ValidateModel        *** Total 10 invalid elements (0.13%) ***
2026-02-15 19:32:47.226 DEBUG  ValidateModel        *** Total 7925 elements ***
```

The generated model itself does not show any errors!

PR #3206, opened 2026-02-13
OTA optimizations

Apollon77:
This PR contains several optimizations, fixes and cleanups in the OTA process based on issues shown in production.

It also adds a new UpdateFailed event to inform about this and to be more transparent

addressing https://github.com/matter-js/matterjs-server/issues/210
addressing https://github.com/ioBroker/ioBroker.matter/issues/694

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,9 @@ The main work (all changes without a GitHub username in brackets in the below li
     - Enhancement: Added some jitter to the subscription max ceiling to spread out subscription responses from devices
     - Fix: Initializes the Message Reception state counter correctly as defined by the Matter specification
     - Fix: Ensures that BDX sessions inform upper layers correctly in all canceled cases
+    - Fix: Fixes Session mapping for PASE/CASE messages
+    - Fix: When we have sent out the success for a session resumption ensure the session is registered even if we do not get the ack
+    - Fix: Ensures processing of received messages that lead to earlier Standalone Acks for previous messages 
 
 - @matter/types
     - Enhancement: Re-Parsed the Matter 1.4.2 specification to improve captured details. No functional changes
--- a/packages/protocol/src/protocol/ExchangeManager.ts
+++ b/packages/protocol/src/protocol/ExchangeManager.ts
@@ -178,7 +178,11 @@ export class ExchangeManager {
         if (packet.header.sessionType === SessionType.Unicast) {
             if (packet.header.sessionId === UNICAST_UNSECURE_SESSION_ID) {
                 if (this.#isClosing) return;
-                const initiatorNodeId = packet.header.sourceNodeId ?? NodeId.UNSPECIFIED_NODE_ID;
+
+                // Responses include our ephemeral initiator nodeId as destNodeId and no sourceNodeId
+                // Initiating requests include their sourceNodeId but no destNodeId
+                const initiatorNodeId =
+                    packet.header.destNodeId ?? packet.header.sourceNodeId ?? NodeId.UNSPECIFIED_NODE_ID;
                 session =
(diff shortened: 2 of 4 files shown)


packages/matter.js/src/device/PairedNode.ts
```diff
@@ -399,7 +399,13 @@ export class PairedNode {
                 }`,
             );
             if (this.#connectionState === NodeStates.Connected) {
-                this.#scheduleReconnect();
+                if (this.#nodeShutdownReason === NodeShutDownReason.ForUpdate) {
+                    this.#nodeShutdownDetected = false;
+                    this.#nodeShutdownReason = undefined;
+                    this.#scheduleReconnect(Millis(RECONNECT_DELAY_AFTER_SHUTDOWN * 4));
```
lauckhart: What condition leads to the "x4" version of the delay?  Maybe a soars constant would be helpful?

Might also be helpful to create a separate function so the portions of this function that are copied in updateTimeoutHandle could be shared?

packages/node/src/behavior/system/software-update/SoftwareUpdateManager.ts
```diff
@@ -113,6 +114,22 @@ export interface SoftwareUpdateInfo {
...
+    vendorId: VendorId;
+    productId: number;
+    targetSoftwareVersion: number;
+
+    /**
+     * - `"queued"` — waiting for its turn; no announcement has been sent yet.
+     * - `"in-progress"` — the OTA provider has been announced to the node and we are waiting for it to complete.
+     * - `"stalled"` — no progress update received for 15 minutes. The entry will be automatically reset and retried.
+     */
+    status: "queued" | "in-progress" | "stalled";
```
lauckhart, line 128: Maybe this interface is "PendingUpdate"?  Seems odd that a "queued update" may not be queued :)

packages/node/src/behaviors/ota-software-update-provider/OtaSoftwareUpdateProviderServer.ts
```diff
@@ -405,20 +443,56 @@ export class OtaSoftwareUpdateProviderServer extends OtaSoftwareUpdateProviderBe
...
+        const now = Time.nowMs;
         if (updateToken !== undefined) {
             const key = `${requestorNodeId}-${fabricIndex}-${Bytes.toHex(updateToken)}`;
-            return this.internal.inProgressDetails.get(key);
+            const details = this.internal.inProgressDetails.get(key);
+            if (details !== undefined && details.timestamp + Minutes(15) < now) {
+                logger.info(
+                    `Removing stale in-progress OTA entry for Requestor`,
+                    peerAddress,
+                    `(age: ${Math.round((now - details.timestamp) / 60_000)}min)`,
```
lauckhart: Could use Duration.format here

Also this is another copied code block

packages/nodejs-shell/src/MatterNode.ts
```diff
@@ -179,6 +179,9 @@ export class MatterNode {
         this.#observers.on(updateManagerEvents.updateDone, peer => {
             logger.info(`Update done for peer `, peer);
         });
+        this.#observers.on(updateManagerEvents.updateFailed, peer => {
+            logger.info(`Update failed for peer `, peer);
```
lauckhart: Space suffix is unnecessary

packages/matter.js/src/device/PairedNode.ts
Apollon77: Condition is that applying an update usually takes longer and also the devce should contact us when done and restarted with a "versionApplied" message, so this is basically to ensure more time then the standard 30s

PR #3210, opened 2026-02-15
Do not include controller in ServerNode by default

lauckhart:


Apollon77 (approved):

lauckhart:
> @lauckhart see linter issue, and please add a line in changelog. And do we need to update the "new" Controller example?

OK, I removed `ControllerBehavior` from `MatterController` API too then if we're going to mention in changelog.  Controller example does not require changes.

PR #3249, opened 2026-02-19
Refactor BleScanner

Apollon77:


packages/protocol/src/common/AbstractBleScanner.ts
```diff
@@ -0,0 +1,348 @@
...
+    SD: number; // Additional Field for Short discriminator
+};
+
+export type DiscoveredBleDevice<TPeripheral> = {
+    deviceData: CommissionableDeviceData;
+    peripheral: TPeripheral;
+    hasAdditionalAdvertisementData: boolean;
+};
+
+export abstract class AbstractBleScanner<TPeripheral> implements Scanner {
```
lauckhart: I think just "BleScanner" might be a better name, aligning with our convention elsewhere

packages/protocol/src/common/AbstractBleScanner.ts
```diff
@@ -0,0 +1,348 @@
...
+    readonly #discoveredMatterDevices = new Map<string, DiscoveredBleDevice<TPeripheral>>();
+
+    constructor(client: BleScannerClient<TPeripheral>) {
+        this.#client = client;
+        this.#client.setDiscoveryCallback((peripheral, manufacturerData) =>
+            this.#handleDiscoveredDevice(peripheral, manufacturerData),
+        );
+    }
+
+    protected abstract getPeripheralAddress(peripheral: TPeripheral): string;
```
lauckhart: Given that a.) this is the only bit that's abstract, and b.) it's just a function to retrieve a property of another object... Could we instead just give TPeripheral a proper type (e.g. TPeripheral extends BlePeripheral) and convert BleScanner into a concrete class without derivatives?

Apollon77: TPeripheral is the generic which points to whatever underlaying object the BLE implementation uses ... Noble has a "Peripheral" class and the react native stuff something complete differerent. So we can not streamline this because this is coming from the concrete implementation. Thats the generics of that class. Or what you mean?

Apollon77: Adjusted

PR #3278, opened 2026-02-23
SSL support for remote protocols

lauckhart:
* Adds PEM support to `Key`

* Adds PKCS8 export to `Key`

* Modifies `X509` to support field combinations appropriate for SSL certs

* Fixes bug in PEM decoding

* Adds `https` and `wss` protocols and key/cert parameters to `HttpService`

* Adds key/cert state values to `RemoteServer`

* Adds self-signed cert generation to `RemoteServer` when the user doesn't provide

### DIFF
--- a/packages/general/src/crypto/X509.ts
+++ b/packages/general/src/crypto/X509.ts
@@ -50,10 +50,8 @@ export namespace X509 {
     export function certificateToDer(cert: Certificate | UnsignedCertificate) {
         const { serialNumber, signatureAlgorithm, issuer, validity, subject, publicKey, extensions } = cert;
 
-        const {
-            basicConstraints: { isCa, pathLen },
-        } = extensions;
-        if (!isCa && pathLen !== undefined) {
+        const { basicConstraints } = extensions;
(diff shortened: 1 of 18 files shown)


packages/general/src/crypto/X509.ts
```diff
@@ -50,10 +50,8 @@ export namespace X509 {
     export function certificateToDer(cert: Certificate | UnsignedCertificate) {
         const { serialNumber, signatureAlgorithm, issuer, validity, subject, publicKey, extensions } = cert;
 
-        const {
-            basicConstraints: { isCa, pathLen },
-        } = extensions;
-        if (!isCa && pathLen !== undefined) {
+        const { basicConstraints } = extensions;
+        if (basicConstraints && !basicConstraints?.isCa && basicConstraints?.pathLen !== undefined) {
```
Apollon77: why the "?" when you checked for existence of basicConstraints before?

lauckhart: dunno, fixed

PR #3292, opened 2026-02-26
Handle errors during BLE write

Apollon77:
addresses https://github.com/matter-js/matterjs-server/issues/327

When the BLE peripheral disconnects while noble's _withDisconnectHandler has a characteristicC1ForWrite.writeAsync() in flight, noble rejects that promise with Error:
  Disconnected 19. This rejection was unhandled in two places:

   1. btpSendAckTimeoutTriggered() — called by a timer whose callback promise is never caught by StandardTimer. If writeBleCallback threw here, the rejection propagated all the way up as an 
  unhandled rejection, crashing the process.
   2. processSendQueue() — if writeBleCallback threw mid-send, sendInProgress was never reset to false, permanently blocking future sends.

  Fixes applied to packages/protocol/src/ble/BtpSessionHandler.ts:

   - btpSendAckTimeoutTriggered(): Added if (!this.isActive) return guard at the top, and wrapped writeBleCallback in try/catch that logs the error and returns. The disconnect event 
  independently closes the session.
   - processSendQueue(): Wrapped writeBleCallback in try/catch that logs, resets sendInProgress, and returns early. The disconnect event handles session cleanup.


packages/protocol/src/ble/BtpSessionHandler.ts
```diff
@@ -449,7 +457,13 @@ export class BtpSessionHandler {
             };
             this.prevAckedSequenceNumber = this.prevIncomingSequenceNumber;
             const packet = BtpCodec.encodeBtpPacket(btpPacket);
-            await this.writeBleCallback(packet);
+            try {
+                await this.writeBleCallback(packet);
+            } catch (error) {
```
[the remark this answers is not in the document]
Apollon77: @copilot open a new pull request to apply changes based on [this feedback](https://github.com/matter-js/matter.js/pull/3292#discussion_r2861437669)

packages/protocol/src/ble/BtpSessionHandler.ts
```diff
@@ -386,7 +385,20 @@ export class BtpSessionHandler {
             const packet = BtpCodec.encodeBtpPacket(btpPacket);
             logger.debug(`Sending BTP packet raw: ${Bytes.toHex(packet)}`);
 
-            await this.writeBleCallback(packet);
+            try {
+                await this.writeBleCallback(packet);
+            } catch (error) {
+                // BLE connection was lost mid-send; the disconnect event will close the session.
+                // Clear the queue to avoid malformed state from partially-consumed DataReaders.
+                logger.debug("BTP packet send failed, BLE connection likely already closed", error);
```
lauckhart: Seems a little odd to ignore these errors with only a debug note.  You don't think this could make it hard to diagnose issues in the future?

packages/protocol/src/ble/BtpSessionHandler.ts
```diff
@@ -447,9 +460,15 @@ export class BtpSessionHandler {
                     sequenceNumber: this.getNextSequenceNumber(),
                 },
             };
-            this.prevAckedSequenceNumber = this.prevIncomingSequenceNumber;
             const packet = BtpCodec.encodeBtpPacket(btpPacket);
-            await this.writeBleCallback(packet);
+            try {
+                await this.writeBleCallback(packet);
+            } catch (error) {
```
lauckhart, line 469: Same here as previous comment

lauckhart (approved):

packages/protocol/src/ble/BtpSessionHandler.ts
Apollon77: Good idea, adjusted

PR #3299, opened 2026-02-28
Add write-ahead log storage backend

lauckhart:
This adds WAL-backed storage.  Commit messages are typical verbose Claude; see those for implementation details.

## High level goals:

1. Create a replacement for the current "file" storage that reduces I/O and disk space consumed during normal operation, and

2. Support atomic commit of matter.js transactions at the storage level

## Design

I made a few design decisions here that we can consider:

1. We lean into our current usage pattern, which is to load storage at startup and then operate from memory.  The new storage loads the full storage tree on read but flushes it from memory on write.  This means that individual reads would be expensive.  The advantage is that we do not need to maintain a redundant in-memory structure.

2. There's a single journal with periodic snapshots into a single blob.  We could potentially split this out by node or something, but given our current usage pattern there wouldn't be much benefit to this.

3. I added support for cleaning up old WAL entries, but I turned it off by default.  So the journal will currently grow unbounded.

4. We use gzip to compress the snapshot and old logs (only the head segment is uncompressed).  This trades CPU for space and I/O, but the files should have a high compression ratio so I think it's worth it.

## Supporting changes

1. General filesystem abstraction.  This lets us implement file-based functionality without tying it specifically to Node's APIs.

2. Renamed "Storage" to "StorageDriver".  The term "storage" was really overused and collides with the built-in web Storage object

3. Refactored storage driver management.  Added a general-purpose registration system for drivers to `StorageService` and removed the node.js-specific `StorageFactory`

4. Modified how CHIP testing initializes storage so we can test using different storage backends

5. Mapped matter.js transactions to SQLite transactions too for the SQLite driver

6. Added Gzip component

### DIFF
--- a/packages/general/src/storage/StorageService.ts
+++ b/packages/general/src/storage/StorageService.ts
@@ -19,7 +19,6 @@ import { StorageMigration } from "./StorageMigration.js";
 const logger = Logger.get("StorageService");
 
 const DRIVER_JSON = "driver.json";
-
 /**
  * Service adapter for the Matter.js storage API.
  */
@@ -100,8 +99,7 @@ export class StorageService {
         }
 
         const fs = this.#environment.get(Filesystem);
-        const parentDir = fs.directory(namespace);
-        const dir = parentDir;
+        let dir = fs.directory(namespace);
 
         // Detect existing driver
         let descriptor = await this.#readDescriptor(dir);
@@ -126,7 +124,7 @@ export class StorageService {
         // Migration: if we detected an existing driver that differs from the target, migrate
         if (detectedKind !== undefined && detectedKind !== targetKind) {
             await this.#migrate(fs, namespace, dir, detectedKind, targetKind);
-            // After migration, re-read descriptor
+            dir = fs.directory(namespace);
             descriptor = await this.#readDescriptor(dir);
         }
 
@@ -208,13 +206,7 @@ export class StorageService {
         await file.write(JSON.stringify(descriptor, undefined, 2));
     }
 
-    async #migrate(
-        fs: Filesystem,
-        namespace: string,
-        dir: Directory,
-        fromKind: string,
-        toKind: string,
-    ) {
+    async #migrate(fs: Filesystem, namespace: string, sourceDir: Directory, fromKind: string, toKind: string) {
         const fromImpl = this.#drivers.get(fromKind);
         const toImpl = this.#drivers.get(toKind);
         if (!fromImpl || !toImpl) {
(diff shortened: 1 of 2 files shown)


packages/general/src/storage/StorageService.ts
```diff
@@ -80,9 +183,85 @@ export class StorageService {
...
             Diagnostic.dict({
-                location: location ?? "(unknown)",
-                available: !!this.#factory,
+                location: this.#location ?? "(unknown)",
+                available: this.#factory !== undefined || this.#drivers.size > 0,
             }),
         ];
     }
+
+    async #readDescriptor(dir: import("../fs/Directory.js").Directory): Promise<StorageDriver.Descriptor | undefined> {
```
Apollon77: what's that? dynamic import?

Apollon77: ... and more such places down the lines

packages/general/src/storage/StorageService.ts
```diff
@@ -80,9 +183,85 @@ export class StorageService {
...
+
+            await sourceStorage.close();
+            await targetStorage.close();
+
+            if (result.success) {
+                logger.info(
+                    `Migration complete: ${result.migratedCount} items migrated, ${result.skippedCount} skipped`,
+                );
+            } else {
+                logger.warn(
```
Apollon77: do we need to cleanup anything in that case?

packages/nodejs/src/storage/fs/StorageBackendDisk.ts
```diff
@@ -33,7 +33,13 @@ interface ContextIndex {
     keys?: Set<string>;
 }
 
-export class StorageBackendDisk extends Storage {
+export class StorageBackendDisk extends StorageDriver {
+    static readonly id = "file";
+
+    static create(dir: import("@matter/general").Directory) {
```
Apollon77: also here dynamic imports?

packages/general/src/storage/StorageManager.ts
```diff
@@ -5,17 +5,25 @@
...
+    }
+
+    get driverId() {
+        return this._driver.id;
+    }
 
     initialize(): MaybePromise<void> {
-        if (!this.storage.initialized) {
-            const init = this.storage.initialize();
+        if (!this._driver.initialized) {
```
Apollon77: why _driver?

support/chip-testing/test/test.config.ts
```diff
@@ -20,6 +22,15 @@ Object.assign(globalThis, { chip });
 
 await chip.initialize();
 
+// Default to memory driver for tests; overridable via MATTER_STORAGE_DRIVER env var.  We use Boot.init so the
+// setting survives Boot.reboot() which the test runner calls before each file
+Boot.init(() => {
```
Apollon77, line 27: is this also used when strting the TestIntstance file like AllClusters directly?

Apollon77 (approved):

support/chip-testing/test/test.config.ts
lauckhart: No, this only applies to things run by matter-test

packages/general/src/storage/StorageManager.ts
lauckhart: typo, fixed

packages/general/src/storage/StorageService.ts
lauckhart: OK resolved in additional (already reviewed) commit

PR #3327, opened 2026-03-03
Rework passcode generation and tests

Apollon77:
to be more spec compliant and to also ensure all works with "shorter than 8 number passcodes"


packages/protocol/src/session/pase/PaseClient.ts
```diff
@@ -27,12 +38,19 @@ export class PaseClient {
...
+                passcode >= 1 &&
+                passcode <= 99_999_998 &&
+                !CommissioningOptions.FORBIDDEN_PASSCODES.includes(passcode)
+            ) {
+                return passcode;
+            }
         }
-        return passcode;
+
+        throw new InternalError("Unable to generate valid passcode in 100 tries; entropy source is broken");
```
lauckhart: Should use constant for # of asteroids for uniformity

PR #3330, opened 2026-03-03
Refactor and enhance Commissioning APIs

Apollon77:
We now try devices one after the other always one address and so give all devices a better chance to be tried within the timeout. We also ensure to continue (but remove the device) on PASE errors (e,g wrong passcode).

Also some code was streamlined between the legacy and node driven path. Also node driven commissionings can now handle multiple targets by chosing the relevant node depending on the commissioning result.

This change can also be a basis to easiely implement Multi-QR code commissioning in a later step

This PR also adds support for Aborting PASE processes comparable to CASE.

### DIFF
--- a/packages/nodejs-ble/src/NobleBleChannel.ts
+++ b/packages/nodejs-ble/src/NobleBleChannel.ts
@@ -535,7 +535,8 @@ export class NobleBleChannel extends BleChannel<Bytes> {
                             () => logger.debug(`Peripheral ${peripheralAddress}: Disconnected from peripheral`),
                             error => logger.debug(`Peripheral ${peripheralAddress}: Error while disconnecting`, error),
                         );
-                    });
+                    })
+                    .catch(() => {});
             },
 
             // callback to forward decoded and de-assembled Matter messages to ExchangeManager


packages/node/src/behavior/system/controller/ControllerBehavior.ts
```diff
@@ -160,6 +168,99 @@ export class ControllerBehavior extends Behavior {
...
+            const config = this.fabricAuthorityConfig;
+            fabric = await fabricAuthority.defaultFabric(config);
+        }
+        if (!fabricAuthority.hasControlOf(fabric)) {
+            throw new ImplementationError(
+                `Cannot commission into fabric ${fabric.fabricIndex} because we do not control this fabric`,
+            );
+        }
+
+        const peerAddress = await this.env.get(ControllerCommissioner).commissionWithDiscovery({
```
lauckhart: We shouldn't be using these old methods...  if we want to try commissioning multiple nodes sequentially let's just iterate over discovered nodes as they come in, not discover/commission via an entirely different code path.  Looks like that would make most of the code in this function unnecessary as well as it's converting to/from legacy api

packages/node/src/behavior/system/controller/ControllerBehavior.ts
```diff
@@ -160,6 +168,99 @@ export class ControllerBehavior extends Behavior {
         }
     }
 
+    /**
+     * Commission a node by discovery criteria from the controller context.
+     *
+     * This method is intended for ambiguous discovery criteria (for example same discriminator on multiple devices).
+     * It discovers candidates and then commissions one candidate at a time using the node-level commissioning path.
```
lauckhart: This comment seems wrong. Doesn't it *attempt* to commission sequentially and stop when the first one succeeds?

And I think it *doesn't* use the node-level commissioning path. It moves everything to the old api then duplicates what the new api does

packages/protocol/src/peer/CommissioningConnection.ts
```diff
@@ -0,0 +1,94 @@
...
+    NoResponseTimeoutError,
+    ServerAddress,
+    UnexpectedDataError,
+} from "@matter/general";
+import { CommissioningConnectionPool } from "./CommissioningConnectionPool.js";
+import { TransientPeerCommunicationError } from "./PeerCommunicationError.js";
+
+const logger = Logger.get("CommissioningConnection");
+
+export class CommissioningConnection {
```
lauckhart: This appears to be a function disguised as a class (single function as instance API) maybe just turn into a factory function?

packages/protocol/src/peer/CommissioningConnection.ts
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    #isRetryableNetworkError(error: Error) {
+        return causedBy(error, NoResponseTimeoutError, TransientPeerCommunicationError);
+    }
+}
+
+export namespace CommissioningConnection {
+    export interface Options {
+        devices: CommissionableDevice[];
```
lauckhart, line 164: Some comments would help here. There is both an array of devices and a function that returns a promise for an array of devices. Does this represent two input sources to feed the commission loop?  Maybe we could just have a single AsyncIterable as input?

packages/node/src/behavior/system/controller/ControllerBehavior.ts
Apollon77: Hm, ok then we need to somehow track the nodes and their addresses that exist initially and also fire off a new  discovery that might add multiple more nodes or addresses and then iterate over them AND have a option for the "new commissioning flow" to limit the tries for a node to just one address because the idea is to give all discovered nodes a chance and then still also sort them out when a PASE error is returned thats not a timeout. Like that?

lauckhart: The new discovery logic is streaming. So "discovery" is a process rather than an event and you react as discovery occurs. CommissioningDiscovery makes it appear to be "one time" but that's only because its output is (currently) a single node. The other discovery classes are streaming

lauckhart: And yes, we would need to add some options to commissioning to support coordination when dealing with multiple nodes.

I'm wondering though about the strategy of round robin by IP.  Might it make more sense to just initiate commissioning in parallel as discovered until the point where we successfully initiate commissioning on one node?  Then abort everything else?  Then timeouts, latency etc. couldn't slow down commissioning for subsequent nodes

Apollon77: Yes I think parallel until we got a PASE and then cancelling the others is the right approach I guess. We just need to be careful and likely need a special enhancement ... we should try to complete PASE instead of aborting it and then close the session because else we could block devices for new PASE tries for 60s. So we can "race" with the abort for the sendPbkdfParamRequest as soon as we got this we should finish and not abort inside pase process ... and directly close again ... or "Abort" means we send an InvalidParam at any step in between (if not yet "finshed").... WDYT?

packages/node/src/behavior/system/controller/ControllerBehavior.ts
```diff
@@ -33,7 +36,7 @@
     getFabricQname,
 } from "@matter/protocol";
 import { CaseAuthenticatedTag, FabricId, FabricIndex, NodeId } from "@matter/types";
-import type { CommissioningClient } from "../commissioning/CommissioningClient.js";
+import { CommissioningClient } from "../commissioning/CommissioningClient.js";
```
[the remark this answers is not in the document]
Apollon77: will be fixed

lauckhart (approved):

PR #3337, opened 2026-03-04
Do not re-announce commissionable too early

### LINKED ISSUE #3336: ServerNode publishes ephemeral commissionable mDNS advertisement after removeFabric, before factory reset completes
## Summary

When a  handles  (the last fabric), it goes through two distinct phases before reaching a stable uncommissioned state. During the transition, **multiple ephemeral commissionable mDNS service instances** are published and then immediately unpublished. A commissioner that discovers one of these transient advertisements and begins PASE commissioning will hit a `peer-unresponsive` / `Operation aborted` error because the device undergoes a full factory reset mid-handshake, destroying all sessions.

## Observed Sequence (device-side logs)

```
# removeFabric received – device starts tearing down:
11:09:50.881  Unpublishing kind: operational service: mdns:9C763ED3A41A095C-0000000000000001._matter._tcp.local
11:09:50.902  removeFabric statusCode: 0 fabricIndex: 1

# Phase 1 – first ephemeral commissionable advertisement (pre-reset):
11:09:52.409  Publishing kind: commissionable service: mdns:C35F47C9B2EF2964._matterc._udp.local
11:09:52.409  Broadcast kind: commissionable service: mdns:C35F47C9B2EF2964 number: 1 next: 1.16s
11:09:53.683  Broadcast kind: commissionable service: mdns:C35F47C9B2EF2964 number: 2 next: 2.4s
11:09:56.197  Broadcast kind: commissionable service: mdns:C35F47C9B2EF2964 number: 3 next: 4.47s

# Phase 2 – node going offline for factory reset:
11:09:58.084  Node: node0 going offline
11:09:58.084  AdministratorCommissioning: Ending commissioning
11:09:58.085  Unpublishing kind: commissionable service: mdns:7330FC95C0788E59   (was open commissioning window)
11:09:58.086  Unpublishing kind: commissionable service: mdns:C35F47C9B2EF2964   ← Phase 1 advertisement retracted after only ~5.7 s

# Another ephemeral advertisement published then immediately retracted:
11:09:58.084  Publishing kind: commissionable service: mdns:3B3879BF735416F5._matterc._udp.local
11:09:58.187  Unpublishing kind: commissionable service: mdns:3B3879BF735416F5   ← lived only 103 ms

# Factory reset:
11:09:58.298  DeviceCommissioner: No longer commissioning
11:09:58.298  Node: node0 is offline
11:09:58.298  Node: node0 resetting to factory defaults

# Stable state – final commissionable advertisement:
11:09:58.352  Publishing kind: commissionable service: mdns:3C3801E6F7D13685._matterc._udp.local
11:09:58.352  Commissioning: node0 is uncommissioned passcode: 20202021 discriminator: 3840
```

### Timeline summary

| Δt from removeFabric | Event |
|---|---|
| +0 ms | `removeFabric` completes |
| +1,507 ms | **Phase 1 advertisement** `C35F47C9B2EF2964` published (broadcasts 3 times) |
| +7,677 ms | Phase 1 advertisement **retracted**; factory reset begins |
| +7,775 ms | Another service `3B3879BF735416F5` published then retracted after **103 ms** |
| +7,850 ms | Factory reset completes; **stable** advertisement `3C3801E6F7D13685` published |

## Impact

Any commissioner (e.g. Home Assistant via Python Matter Server) that:
1. Performs mDNS discovery filtered by discriminator immediately after decommissioning a node, **and**
2. Calls `commissionOnNetwork` within ~1–8 seconds

…will discover `C35F47C9B2EF2964`, start PASE, then fail with:

```
Commission failed: Operation aborted (peer-unresponsive)
```

because the device factory-resets mid-handshake, invalidating all sessions.

## Workaround

Wait at least **10–12 seconds** after decommission before attempting `commissionOnNetwork` to ensure the device has completed its factory reset cycle and the stable advertisement is the only one visible.

## Expected Behavior

The commissionable mDNS advertisement should **not** be published until the device has completed its factory reset and reached a stable uncommissioned state. Specifically:
- No commissionable advertisement should be emitted during the teardown/transition period.
- Only the final, stable advertisement (after `node0 is uncommissioned`) should be broadcast.

Alternatively, if the ephemeral advertisement is intentional (e.g. for parallel discovery), the node should **not** destroy sessions mid-commissioning; instead it should let any in-progress PASE handshake complete or explicitly send a commissioning failure status before resetting.

## Environment

- matter.js version: checked against current `main`
- Node.js: 22.x
- Platform: macOS (arm64)
- Reproduced in: `@matter/main` `ServerNode` used as a simulated light device in integration tests

## Additional Notes

The root cause appears to be in how `removeFabric` teardown interacts with the `AdministratorCommissioning` / commissioning window handling and the `DeviceCommissioner` reset path. The pre-reset state briefly transitions through commissioning mode before the factory reset takes over, resulting in the ephemeral advertisement.

The log line `AdministratorCommissioning: Ending commissioning` at 11:09:58.084 (same millisecond as `Node: node0 going offline`) suggests the commissioning window cleanup and node shutdown happen concurrently and each independently triggers a new advertisement.

Apollon77:
fixes #3336

Because of the fact that we mark the fabric as deleted before we close the session this logic triggered a commissionable announcement but we were also preparing for factory reset, so we announced, then resetted and announced again

### DIFF
--- a/packages/node/test/node/FailsafeCommissioningTest.ts
+++ b/packages/node/test/node/FailsafeCommissioningTest.ts
@@ -19,7 +19,8 @@
  *    - No factory reset (state.commissioned was never true; handleFabricChange sees false→false, no state change).
  */
 
-import { Lifecycle, Crypto, MockCrypto, Seconds } from "@matter/general";
+import { ServerNode } from "#node/ServerNode.js";
+import { Crypto, Lifecycle, MockCrypto, Seconds } from "@matter/general";
 import {
     Advertiser,
     ControllerCommissioningFlow,
@@ -28,7 +29,6 @@ import {
     FabricManager,
     ServiceDescription,
 } from "@matter/protocol";
-import { ServerNode } from "#node/ServerNode.js";
 import { MockSite } from "./mock-site.js";
 
 describe("Failsafe commissioning re-announcement", () => {


packages/protocol/src/protocol/DeviceAdvertiser.ts
```diff
@@ -120,6 +120,14 @@ export class DeviceAdvertiser {
                 return;
             }
 
+            // Only PASE sessions (commissioning sessions) should resume commissioning advertisement.
+            // CASE sessions on fabrics being deleted (decommissioning teardown) must NOT trigger
+            // commissionable re-advertisement.  The factory reset following decommissioning of the
+            // last fabric will advertise at the appropriate time.
+            if (!session.isPase) {
```
lauckhart, line 127: I think I agree with copilot in this case, would be good to add a test.  You mentioned integration tests failed but there aren't any tests that cover this (including chip) or we would have caught it already

Also, how does this work if commissioning transitions to case but then doesn't call commissioningComplete?  Does that do a rollback without a factory reset?

lauckhart (approved):

packages/protocol/src/protocol/DeviceAdvertiser.ts
Apollon77: test added

packages/protocol/test/protocol/DeviceAdvertiserTest.ts
```diff
@@ -0,0 +1,131 @@
...
+import { Observable } from "@matter/general";
+import { FabricIndex, NodeId, VendorId } from "@matter/types";
+
+/**
+ * Minimal mock advertiser that records calls to advertise().
+ */
+class MockAdvertiser extends Advertiser {
+    readonly advertiseCalls: Array<{ description: ServiceDescription; event: Advertiser.BroadcastEvent }> = [];
+
+    protected getAdvertisement(_description: ServiceDescription) {
```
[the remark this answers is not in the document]
Apollon77: As you see all good and runs, i think this does not count for abstract classes because they are basically defined and not overwritten

PR #3339, opened 2026-03-04
Add some more BasicInfo warnings

Apollon77:
... to use less defaults :-)


packages/node/src/behaviors/basic-information/BasicInformationServer.ts
```diff
@@ -92,7 +92,15 @@ export class BasicInformationServer extends Base {
...
-            logger.warn("uniqueId and serialNumber shall not be the same.");
+            logger.warn("uniqueId and serialNumber shall not be the same");
+        }
+
+        if (this.state.productLabel !== undefined && this.state.productLabel.includes(this.state.vendorName)) {
+            logger.warn("productLabel should not include vendorName");
+        }
+
+        if (this.state.manufacturingDate === "20200101") {
+            logger.warn("manufacturingDate should not be 20200101");
```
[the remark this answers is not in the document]
lauckhart: Matter devices could not have been manufactured on this date, and this constant has been informally elevated to "special" by the CHIP SDK.  So the value is OK but maybe we could assign to a constant such as "CHIP_DUMMY_MANUFACTURE_DATE"

packages/node/src/behaviors/bridged-device-basic-information/BridgedDeviceBasicInformationServer.ts
```diff
@@ -36,6 +36,18 @@ export class BridgedDeviceBasicInformationServer extends BridgedDeviceBasicInfor
         if (serialNumber !== undefined && uniqueId === this.state.serialNumber) {
             logger.warn("uniqueId and serialNumber shall not be the same.");
```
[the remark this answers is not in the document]
lauckhart: In the extremely unlikely event we generate a unique ID that matches the serial number we shouldn't penalize the user...  If we want to do anything we should generate unique IDs until it doesn't match the serial number.  But I'm not sure it's worth the code for something that Claude describes as "Effectively impossible. That's roughly 1 in 10⁵⁵ — far less likely than anything that has ever happened."

packages/node/src/behaviors/bridged-device-basic-information/BridgedDeviceBasicInformationServer.ts
```diff
@@ -36,6 +36,18 @@ export class BridgedDeviceBasicInformationServer extends BridgedDeviceBasicInfor
...
         }
+
+        if (
```
lauckhart: Does it make sense to print these warnings for bridged devices, where they could come from the device itself?

If it does, maybe would be better implemented as a static utility on BasicInformation so implementation can be shared.

packages/node/src/behaviors/bridged-device-basic-information/BridgedDeviceBasicInformationServer.ts
```diff
@@ -36,6 +36,18 @@ export class BridgedDeviceBasicInformationServer extends BridgedDeviceBasicInfor
...
+            this.state.productLabel !== undefined &&
+            this.state.vendorName !== undefined &&
+            this.state.productLabel.includes(this.state.vendorName)
+        ) {
+            logger.warn("productLabel should not include vendorName");
+        }
+
+        if (this.state.manufacturingDate === "20200101") {
```
lauckhart: A bridged non-matter device could in fact have been manufactured on 2020-01-01, right?

Apollon77: Yes but right mnow the cert would fail and I guess noone knows the manufactiring data of bridged devices ... lets see if we get feedback, but I will clarify

PR #3345, opened 2026-03-05
fix(react-native): restore DER key-container path for ECDSA sign/verify

Apollon77:
## Summary
- add React Native `signEcdsa` override that uses DER `pkcs8` key containers with `createSign` when available, with fallback to `StandardCrypto`
- add React Native `verifyEcdsa` override that uses DER `spki` key containers with `createVerify` when available, with fallback to `StandardCrypto`
- add StandardCrypto regression tests for PKCS8/SPKI key path and DER signature round-trip verification

## Context
Commit `ffff1aa525069e1bc81202dd39ac68ab18b4a4f4` moved DER *signature* conversion into `EcdsaSignature` and removed per-method `dsaEncoding` flags. DER signature support still exists.

The React Native compatibility issue here is separate: quick-crypto ECDSA key ingestion can be unreliable through subtle/JWK import, while Node-style sign/verify works with DER key containers. This PR restores that RN compatibility path without changing shared signature semantics.

## Details
- DER signature handling remains centralized in `EcdsaSignature` (IEEE-P1363 internal representation)
- RN override only switches key container handling for ECDSA when `pkcs8`/`spki` are present
- fallback to `StandardCrypto` preserves behavior for plain JWK inputs

## Tests
- `npm run build -w packages/react-native`
- `npx matter-test --spec test/crypto/StandardCryptoTest.ts` (from `packages/general`)
- new tests:
  - `signs & verifies with PKCS8/SPKI keys`
  - `verifies DER-encoded ECDSA signatures`

### DIFF
--- a/packages/nodejs-ble/src/NobleBleChannel.ts
+++ b/packages/nodejs-ble/src/NobleBleChannel.ts
@@ -484,12 +484,19 @@ export class NobleBleChannel extends BleChannel<Bytes> {
         logger.debug(
             `Peripheral ${peripheralAddress}: Sending BTP handshake request: ${Diagnostic.json(btpHandshakeRequest)}`,
         );
-        await characteristicC1ForWrite.writeAsync(Buffer.from(Bytes.of(btpHandshakeRequest)), false);
 
-        characteristicC2ForSubscribe.once("data", handshakeHandler);
+        try {
+            await characteristicC1ForWrite.writeAsync(Buffer.from(Bytes.of(btpHandshakeRequest)), false);
 
-        logger.debug(`Peripheral ${peripheralAddress}: Subscribing to C2 characteristic`);
-        await characteristicC2ForSubscribe.subscribeAsync();
+            characteristicC2ForSubscribe.once("data", handshakeHandler);
+
+            logger.debug(`Peripheral ${peripheralAddress}: Subscribing to C2 characteristic`);
+            await characteristicC2ForSubscribe.subscribeAsync();
+        } catch (error) {
+            btpHandshakeTimeout.stop();
+            characteristicC2ForSubscribe.removeListener("data", handshakeHandler);
+            throw error;
+        }
 
         const handshakeResponse = await handshakeResponseReceivedPromise;
 


packages/react-native/src/crypto/ReactNativeCrypto.ts
```diff
@@ -27,6 +37,54 @@ export class ReactNativeCrypto extends StandardCrypto {
     static override provider() {
         return new ReactNativeCrypto(crypto as unknown as WebCrypto);
     }
+
+    override async signEcdsa(privateKey: JsonWebKey, data: Bytes | Bytes[]) {
+        const pkcs8 = (privateKey as JsonWebKey & { pkcs8?: Bytes }).pkcs8;
```
lauckhart, line 42: If this requires `Key` as input it should test the type and create a `Key` if it isn't already a `Key`.  Then use that interface.

If there isn't already one, create `Key.of()` or something to do the JWK -> `Key` upgrade

packages/react-native/src/crypto/ReactNativeCrypto.ts
```diff
@@ -27,6 +37,54 @@ export class ReactNativeCrypto extends StandardCrypto {
...
+                    format: "der",
+                    type: "pkcs8",
+                    dsaEncoding: "ieee-p1363",
+                }),
+            ),
+        );
+    }
+
+    override async verifyEcdsa(publicKey: JsonWebKey, data: Bytes, signature: EcdsaSignature) {
+        const spki = (publicKey as JsonWebKey & { spki?: Bytes }).spki;
```
lauckhart, line 67: This logic should be shared with above...  Key.of would provide that

packages/react-native/src/crypto/ReactNativeCrypto.ts
```diff
@@ -27,6 +37,54 @@ export class ReactNativeCrypto extends StandardCrypto {
     static override provider() {
         return new ReactNativeCrypto(crypto as unknown as WebCrypto);
     }
+
+    override async signEcdsa(privateKey: JsonWebKey, data: Bytes | Bytes[]) {
```
lauckhart, line 41: Rather than duplicating the node.js implementation, we could make that implementation use SPKI and invoke from here.

Regardless, wherever we use SPKI instead of JWK let's leave a comment as to why.

packages/react-native/src/crypto/ReactNativeCrypto.ts
```diff
@@ -27,6 +37,54 @@ export class ReactNativeCrypto extends StandardCrypto {
...
+        const pkcs8 = (privateKey as JsonWebKey & { pkcs8?: Bytes }).pkcs8;
+        if (pkcs8 === undefined) {
+            return super.signEcdsa(privateKey, data);
```
lauckhart, line 44: If I understand the bug correctly this is doomed to fail...  But ensuring there's a `Key` would fix that

PR #3357, opened 2026-03-07
Ensures that the read/subscribe request is always updated

Apollon77:
... and when the read fails


packages/protocol/src/action/client/subscription/SustainedSubscription.ts
```diff
@@ -127,10 +127,14 @@ export class SustainedSubscription extends ClientSubscription {
 
             // Subscribe
             for (const retry of this.#retries) {
+                if (needToRefreshRequest && refreshRequest !== undefined) {
```
lauckhart: Seems like this may have moved a bit too high, probably should be in the try block

Apollon77: Moved,  thought about this a bit and it should not throw ..but yeah better check it

PR #3359, opened 2026-03-07
fix(protocol): decouple final report processing from status ack wait

Apollon77:
## Problem

Final attribute updates were delayed by 500–600ms+ because updates were blocked until the ACK to the final status response was received.

Root cause: an `await sendStatus()` in the final report path that was added to fix a dangling promise.

## Solution

Replace the blocking `await` with proper promise tracking in `InteractionMessenger`. The final status send is tracked and awaited only during `close()`, not on the hot path of report processing. This decouples state updates from protocol ACK timing while preserving reliability.

### `InteractionMessenger`

- Add `#endMessagePromise` to track fire-and-forget final status sends
- Add `trackEndMessage()` helper to set the promise
- Remove blocking `await sendStatus()` from final report path
- In `close()`, await the tracked promise before closing the exchange
- Add regression test: final report iteration completes before status ACK while close still waits for it

### `Peers.#onShutdown`

- Add `subscriptionActive` guard (matching existing pattern in `#onLeave`) to ignore stale shutdown events from device restart that arrive before subscription is fully established
- Fixes race condition exposed by faster report completion: pre-existing shutdown events were closing active sessions mid-subscription


packages/protocol/src/interaction/InteractionMessenger.ts
```diff
@@ -164,6 +177,11 @@ class InteractionMessenger {
     }
 
     async close() {
+        this.#pendingSendsClosed = true;
+        if (this.#pendingSends.size) {
+            // allSettled intentionally never throws because pending sends here are best-effort by design.
+            await MatterAggregateError.allSettled(this.#pendingSends);
```
lauckhart: Slightly confusing message because MatterAggregateError.allSettled is designed specifically so it does throw...  So message is that this shouldn't throw because the sends shouldn't?

packages/protocol/src/interaction/InteractionMessenger.ts
```diff
@@ -164,6 +177,11 @@ class InteractionMessenger {
...
+        }
         await this.exchange.close();
```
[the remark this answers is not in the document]
Apollon77: See comment the tracked promises have an empty catch body, so they will never throw

packages/protocol/src/interaction/InteractionMessenger.ts
```diff
@@ -117,6 +120,16 @@ class InteractionMessenger {
         return this.exchange.send(messageType, payload, options);
     }
 
+    protected trackPendingSend(promise: Promise<unknown>) {
+        if (this.#pendingSendsClosed) {
+            return;
+        }
+        this.#pendingSends.add(promise);
+        promise.finally(() => this.#pendingSends.delete(promise)).catch(() => {
+            // Errors are handled by the creator of the promise.
```
lauckhart: Regardless of intent if it does fail for some reason (e.g. future refactoring) then this will silently swallow. I don't think we should ever do that

packages/protocol/src/interaction/InteractionMessenger.ts
```diff
@@ -117,6 +120,16 @@ class InteractionMessenger {
...
+        });
```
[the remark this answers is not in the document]
Apollon77: The logging is handled by the place that adds it

packages/protocol/src/interaction/InteractionMessenger.ts
```diff
@@ -117,6 +120,16 @@ class InteractionMessenger {
         return this.exchange.send(messageType, payload, options);
     }
 
+    protected trackPendingSend(promise: Promise<unknown>) {
+        if (this.#pendingSendsClosed) {
```
lauckhart: Ack so close to properly tracking the promise but then we leave some dangling. Given that this function promises to track I think it should do so even if they trigger during the close cycle. It seems like the pending-send set should be continually awaited until there is some notion of "done", at which point it is an error to attempt further sends

lauckhart (approved):

packages/protocol/src/interaction/InteractionMessenger.ts
```diff
@@ -164,6 +177,11 @@ class InteractionMessenger {
     }
 
     async close() {
```
lauckhart, line 176: So awaits this?  If the goal is to move along without waiting for final delivery of optional messages it may make sense to place the returned "close" promise out-of-band too.  Otherwise this seems like it's only half a fix..  Data reports will go through quickly but the interaction itself won't return until all closes occur

Apollon77: from what we see the close is after the processing so that should be fine

packages/protocol/src/interaction/InteractionMessenger.ts
```diff
@@ -164,6 +172,18 @@ class InteractionMessenger {
...
+        let drainingIterations = 0;
+        while (this.#pendingSends.size) {
```
lauckhart: Is there not some way where a pending send could be added after a macrotask in which there were none?  That plus the arbitrary 10 iterations makes this seem a bit fragile. Would be good to get some type of "wait until nothing else can be sent" here

PR #3394, opened 2026-03-12
Optimize OTA and Session handling

Apollon77:
This PR addresses some optimizations seen by recent IKEA OTA updates:
* Lower OTA busy responses to the minimum 2mins (5 before)
* When a new queryImage comes in while a process is still in progress try to determine if we can restart the BDX transfer directly (new session was added but old transfer on old session) or if user need to wait
* Fix issues in the OTA state engine for special cases
* When a startUp event is received clean up "old" sessions (not the current one)
* When checking for recent session use the one where the peer "Most recently" responded and not where we tried to send something (should have been since the beginning)

### DIFF
--- a/packages/node/src/behavior/system/software-update/SoftwareUpdateManager.ts
+++ b/packages/node/src/behavior/system/software-update/SoftwareUpdateManager.ts
@@ -32,6 +32,7 @@ import {
     Timer,
     Timestamp,
 } from "@matter/general";
+import type { Session } from "@matter/protocol";
 import {
     BdxProtocol,
     DclOtaUpdateService,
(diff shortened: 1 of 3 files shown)


packages/node/src/node/client/Peers.ts
```diff
@@ -461,6 +462,37 @@ export class Peers extends EndpointContainer<ClientNode> {
             await this.owner.env.get(SessionManager).handlePeerShutdown(peerAddress);
         }
     }
+
+    async #onStartUp(node: ClientNode) {
```
lauckhart, line 502: Would be useful to note what the point of this function is...  E.g. "Process startup events to detect those events that indicate actual node reboot in real-time.  This allows us to terminate sessions that would otherwise linger until timed out and cause interaction delays."

PR #3399, opened 2026-03-12
Redesign MRP kick: abort exchange and restart fresh

Apollon77:
## Summary

Replaces the MRP kick mechanism with a cleaner architectural approach:

- **Old behavior**: Kick short-circuited the MRP retransmission timer on the current exchange, which was fragile due to a race condition where kicks could arrive while a retransmission send was in-flight (timer stopped, `isRunning === false`).
- **New behavior**: Kick aborts the current CASE handshake exchange via a `localAbort` signal and lets `connect()` loop immediately to a fresh exchange with a clean MRP backoff (~2.25s first retry). DNS-SD address discovery or an explicit `peer.kick()` now takes effect within seconds instead of waiting out a potentially many-minute backoff.

### Key changes

- **`MessageExchange`**: Removed `exchange.kick()` method and the `#kick`/`#kickPending` fields entirely — kick logic now lives at the `PeerConnection` layer
- **`PeerConnection`**: Added `KickOrigin` type (`"discover"` | `"connect"`), `lastRestartAt` shared state, and `localAbort`-based restart logic in `attemptOnce()`
- **`PeerTimingParameters`**: Added `mrpKickRestartIntervalDiscover` (30 min) and `mrpKickRestartIntervalConnect` (10 min) rate-limit thresholds
- **`Peer`**: Updated `QuietObservable<[KickOrigin]>` typing and `ConnectionProcess` interface to thread origin through

### Rate limiting (two layers)

1. **`QuietObservable`** (3s debounce, `skipSuppressedEmits: true`): First-line defense against rapid-fire mDNS churn
2. **Per-origin `lastRestartAt` threshold**: Shared across all concurrent address attempts so at most one restart fires per kick

### Design notes

- `localAbort` wraps `addressAbort` (not `overallAbort`) — firing it aborts only the current `pair()` call, not the entire address retry loop
- `lastRestartAt` is scoped to the `PeerConnection` call, shared across concurrent address attempts to prevent burst restarts
- If a kick fires just as `pair()` succeeds, `overallAbort()` has already fired, making `localAbort()` harmless

## Test plan

- [x] 21 new unit tests in `PeerConnectionKickTest.ts` covering:
  - PeerTimingParameters defaults and overrides
  - KickOrigin threading through QuietObservable
  - Rate-limiting logic (thresholds, shared state, cross-origin)
  - localAbort restart pattern (abort propagation, loop restart, double-abort safety)
  - Observable type contract (QuietObservable assignability, use() disposal)
- [x] Full protocol test suite passes (607/607, ESM + CJS)
- [x] Type-check clean, linter clean, formatter clean


packages/protocol/src/peer/PeerTimingParameters.ts
```diff
@@ -60,6 +60,23 @@ export interface PeerTimingParameters {
...
     minimumTimeBetweenMrpKicks: Duration;
+
+    /**
+     * Minimum time between kick-triggered exchange restarts initiated by DNS-SD address discovery.
+     *
+     * A "discover" kick restarts the CASE handshake exchange from scratch with a fresh MRP backoff,
+     * allowing the new address to be used immediately. This threshold prevents mDNS churn from
+     * restarting the handshake too frequently.
+     */
+    mrpKickRestartIntervalDiscover: Duration;
```
lauckhart: This is getting pretty esoteric :)

Trying to think how we could make these more comprehensible, maybe something like:

```ts
retryBackoffResetWindow: {
    connect?: Duration,
    addressChange?: Duration,
}
```

...but I acknowledge this is no more likely to be understood :)

Apollon77: I decided for 

```
kickThrottleInterval: Duration           // was minimumTimeBetweenMrpKicks (3s)
  kickRestartCooldown: {
      addressChange: Duration,             // 30 min
      connect: Duration,                   // 10 min
  }
```

PR #3400, opened 2026-03-12
Allow invocation of methods over remote API even if they aren't Matter commands

lauckhart:
`@method` decorator for marking methods as remotable over privileged APIs (e.g. WebSocket) without exposing them to the Matter protocol layer.

### DIFF
--- a/packages/model/src/decoration/decorators/command.ts
+++ b/packages/model/src/decoration/decorators/command.ts
@@ -4,6 +4,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+import { CommandElement } from "#elements/CommandElement.js";
 import { CommandModel } from "#models/CommandModel.js";
 import { Decorator } from "@matter/general";
 import { element } from "./element.js";
@@ -23,5 +24,5 @@ export function command(...modifiers: element.Modifier<Decorator.MethodCollector
  * not a valid MEI, these commands are invisible to the Matter protocol layer.
  */
 export function method(...modifiers: element.Modifier<Decorator.MethodCollector>[]) {
-    return element(CommandModel, -1, ...modifiers);
+    return element(CommandModel, CommandElement.NO_ID, ...modifiers);
 }
--- a/packages/model/src/elements/CommandElement.ts
+++ b/packages/model/src/elements/CommandElement.ts
@@ -32,6 +32,11 @@ export namespace CommandElement {
     export type Properties = BaseElement.Properties<CommandElement>;
 
     /**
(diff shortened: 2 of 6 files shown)


packages/model/src/decoration/decorators/command.ts
```diff
@@ -14,3 +14,14 @@ import { element } from "./element.js";
...
+
+/**
+ * Decorates a method as a non-Matter command intended for remote invocation over privileged, non-Matter APIs (e.g.
+ * WebSocket) rather than the Matter protocol.
+ *
+ * Works like {@link command} but automatically assigns {@link CommandId.NONE} (-1) as the command ID.  Since -1 is
+ * not a valid MEI, these commands are invisible to the Matter protocol layer.
+ */
+export function method(...modifiers: element.Modifier<Decorator.MethodCollector>[]) {
+    return element(CommandModel, -1, ...modifiers);
```
Apollon77: should use the new constant here, or?

packages/model/test/decoration/semantics/FieldSemanticsTest.ts
```diff
@@ -165,6 +165,29 @@ describe("FieldSemantics", () => {
         expect(rsp.isResponse).true;
     });
 
+    it("adds a method with implicit ID -1", () => {
```
Apollon77, line 168: Could we add a test to verify that something like that added to a cluster behavior is not adding "-1" to the accepted/generatedCommands list?

packages/model/src/decoration/decorators/command.ts
lauckhart: @matter/model doesn't have access to CommandId. I added a separate constant

packages/model/test/decoration/semantics/FieldSemanticsTest.ts
lauckhart: I hadn't implemented ClusterBehavior support.  Added in latest commit

packages/model/test/decoration/semantics/FieldSemanticsTest.ts
```diff
@@ -165,6 +165,29 @@ describe("FieldSemantics", () => {
...
+        class Foo {
+            @method(uint8)
+            @response(2, uint32)
```
Apollon77, line 171: so the response would be added still, right? Because id = 2?

Apollon77: PS: might be ok because test

lauckhart: yes

PR #3405, opened 2026-03-13
Replace MdnsClient with CommissionableMdnsScanner using DnssdNames infrastructure

Apollon77:
## Summary

- Add dynamic `addFilter`/`removeFilter` to `DnssdNames` for runtime filter management
- Add `ServiceDiscovery` abstract base class for generic DNS-SD service discovery
- Implement `CommissionableMdnsScanner` using `DnssdNames` infrastructure instead of the legacy `MdnsClient`
- Fix cache invalidation: subscribe per-entry `DnssdName` observers to evict devices on TTL expiry or goodbye
- Fix `findCommissionableDevices` returning early on first device found (per Scanner interface contract)
- Use `Time.sleep`/`Promise.race` for timeout logic (MockTime compatible)
- Delete legacy `MdnsClient.ts` (1406 lines) and simplify `ActiveDiscoveries`
- Wire up `CommissionableMdnsScanner` in `ControllerBehavior` and `ServerNetworkRuntime`
- 


packages/general/src/net/dns-sd/DnssdNames.ts
```diff
@@ -166,6 +197,21 @@ export class DnssdNames {
         await this.#solicitor.close();
     }
 
+    /**
+     * Register an additional ingress filter. Records accepted by ANY registered filter are processed.
+     * If no filters are registered, all records are accepted.
+     */
+    addFilter(filter: (record: DnsRecord) => boolean) {
```
lauckhart: Adding a getter for this.#filters would be more flexible and make this implementation smaller (so `names.filters.add` instead of `names.addFilter`).

packages/general/src/net/dns-sd/ServiceDiscovery.ts
```diff
@@ -0,0 +1,198 @@
...
+import { ObserverGroup } from "#util/Observable.js";
+import { MaybePromise } from "#util/Promises.js";
+import type { DnssdName } from "./DnssdName.js";
+import type { DnssdNames } from "./DnssdNames.js";
+import { IpService } from "./IpService.js";
+
+/**
+ * A service instance discovered via DNS-SD.
+ */
+export interface DiscoveredService {
```
lauckhart: This class seems to be duplicating properties of the ipService.  Is there a reason to not just return the service/name directly?  Could add getters to strip off ".local" from the qname but for convenience but doesn't seem like there's a reason to duplicate the qname here

packages/general/src/net/dns-sd/ServiceDiscovery.ts
```diff
@@ -0,0 +1,198 @@
...
+    /** Full instance qname, e.g. "MyBorderRouter._meshcop._udp.local" */
+    instanceName: string;
+
+    /** Service type without ".local", e.g. "_meshcop._udp" */
+    serviceType: string;
+
+    /** Resolved IP addresses, port, and TXT parameters for this instance */
+    ipService: IpService;
```
lauckhart: IpService is specifically for maintaining the state of a service we're communicating with.  That's orthogonal to discovery, which is how the service was located.  Could this just return DnssdNames instead?  Then the caller can create an IpService if they need that.

packages/general/src/net/dns-sd/ServiceDiscovery.ts
```diff
@@ -0,0 +1,198 @@
...
+export abstract class ServiceDiscovery<T> extends CancelablePromise<T> {
+    readonly #names: DnssdNames;
+    readonly #serviceType: string;
+    readonly #filter: (record: DnsRecord) => boolean;
+    readonly #observers = new ObserverGroup();
+    readonly #ipServices = new Map<string, IpService>();
+    #isStopped = false;
+    #resolve!: (value: T) => void;
+    #reject!: (cause?: unknown) => void;
+    #timeout?: Timer;
```
lauckhart: I'm hoping to move away from the "I offer an explicit timeout parameter"...  If you are abortable, you don't need to worry about the *reason* for the abort or implement anything yourself to support a specific abort trigger.

packages/general/src/net/dns-sd/ServiceDiscovery.ts
```diff
@@ -0,0 +1,198 @@
...
+ *     protected onDiscovered(service: DiscoveredService) {
+ *         this.#found.push(parseMyService(service));
+ *     }
+ *
+ *     protected onComplete() { return this.#found; }
+ * }
+ * const results = await new MyDiscovery(mdnsService.names);
+ * ```
+ */
+export abstract class ServiceDiscovery<T> extends CancelablePromise<T> {
```
lauckhart: If I understand correctly, is this essentially wildcard name discovery?  Could the API simply be `DiscoverWildcard(suffix: string, abort: AbortSignal): AsyncGenerator<DnssdName>`?

This function would be simpler, more composable and everything it does would be handled by native JS features:

* stop -> controller.abort()
* timeout -> timed abort
* onDiscovered -> yield
* onComplete -> return
* cleanup -> finally

packages/node/src/behavior/system/controller/discovery/ParallelPaseDiscovery.ts
```diff
@@ -90,6 +90,10 @@ export abstract class ParallelPaseDiscovery<W> extends Discovery<W> {
                 this.#pending.delete(attempt);
             });
 
+        // Suppress unhandled-rejection for losing candidates; actual errors are consumed
+        // in onComplete() via allSettled().
+        attempt.catch(() => {});
```
lauckhart: Hides errors

packages/protocol/src/common/Scanner.ts
```diff
@@ -95,6 +96,26 @@ export function DiscoveryData(kvs: Map<string, string>) {
     return dd;
 }
 
+/**
+ * Format DiscoveryData for diagnostic output.
+ */
+export function discoveryDataDiagnostics(data: DiscoveryData & { addresses?: ServerAddress[] }, kind?: string) {
```
lauckhart: This looks like a factory, might capitalize first letter

packages/protocol/src/mdns/CommissionableMdnsScanner.ts
```diff
@@ -0,0 +1,361 @@
...
+
+        this.#names.removeFilter(this.#filter);
+        this.#observers.close();
+        for (const { ipService } of this.#cache.values()) {
+            await ipService.close();
+        }
+        this.#cache.clear();
+    }
+
+    async findCommissionableDevices(
```
lauckhart: Are there still consumers of this method?  We might delete if not

packages/protocol/src/mdns/MdnsService.ts
```diff
@@ -56,7 +54,20 @@ export class MdnsService {
             });
 
             this.#server = new MdnsServer(this.#socket, this.#construction);
-            this.#client = new MdnsClient(this.#socket, this.#construction);
+
+            this.#names = new DnssdNames({
```
lauckhart, line 70: Why'd we change this from being lazy?  Doesn't seem like a big deal but just curious

packages/general/src/net/dns-sd/DnssdNames.ts
Apollon77: Done — exposed `filters` as a getter returning the `Set` directly. Callers now use `names.filters.add(f)` / `names.filters.delete(f)`.

packages/general/src/net/dns-sd/ServiceDiscovery.ts
Apollon77: Agreed — replaced the entire `ServiceDiscovery` class and `DiscoveredService` interface with a simple `discoverNames()` function that yields `DnssdName` directly. No more IpService bundling; callers create it if they need it.

packages/general/src/net/dns-sd/ServiceDiscovery.ts
Apollon77: Yes, done — `discoverNames()` now just yields `DnssdName`s. No IpService involved.

packages/general/src/net/dns-sd/ServiceDiscovery.ts
Apollon77: Done — timeout parameter removed. The new API takes only an `AbortSignal`. Callers use `AbortSignal.timeout()` or their own controller for timed abort.

packages/general/src/net/dns-sd/ServiceDiscovery.ts
Apollon77: Done — replaced the class with `discoverNames(names, suffix, signal)` returning an async iterable of `DnssdName`.

Ended up using a custom `NameDiscovery` interface (manual `AsyncIterableIterator` impl) instead of `async function*` because of a TS 5.9 type inference quirk: since `DnssdName` extends `BasicObservable` (making it `AsyncIterable<DnssdName.Changes>`), TypeScript incorrectly infers the generator yield type as `Changes` instead of `DnssdName`. The custom interface works around this while maintaining the same `for await` consumer ergonomics.

packages/node/src/behavior/system/controller/discovery/ParallelPaseDiscovery.ts
Apollon77: Fixed — removed the `.catch(() => {})`. The winner's error is now captured into `#winnerError` instead of being re-thrown from the `.catch` handler, so the promise never rejects and unhandled-rejection suppression isn't needed. The error is rethrown in `onComplete()` after awaiting the winner attempt.

The `allSettled` cleanup `.catch` now logs unexpected errors instead of swallowing silently.

packages/protocol/src/common/Scanner.ts
Apollon77: Done — renamed to `DiscoveryDataDiagnostics`.

packages/protocol/src/mdns/CommissionableMdnsScanner.ts
Apollon77: No production callers — only test code. Removed from `CommissionableMdnsScanner` and from the `Scanner` interface. Tests rewritten to use `findCommissionableDevicesContinuously` directly.

packages/protocol/src/mdns/MdnsService.ts
Apollon77: Reverted back to lazy creation. Scanner construction was moved from `ControllerBehavior.initialize()` to `#nodeOnline()`, so `names` no longer needs to be ready at MdnsService construction time.

packages/general/src/net/dns-sd/DnssdNames.ts
```diff
@@ -148,6 +171,17 @@ export class DnssdNames {
         return this.#names.get(name.toLowerCase());
     }
 
+    /**
+     * Iterate all currently discovered {@link DnssdName}s.
+     */
+    *discoveredNames() {
```
lauckhart: I think this should be a getter, making it a function because it's generator-backed is kind of leaking the implementation into the API, could do it as an IIFE though if you want it to be lazy

lauckhart (approved):

packages/general/src/net/dns-sd/DnssdNames.ts
Apollon77: Done — converted to a getter returning `Iterable<DnssdName>` via IIFE, so the generator is an implementation detail.

PR #3409, opened 2026-03-14
memory reduction

lauckhart:
Adds some additional heap analysis tooling + a test; implements a series of optimizations to address low-hanging fruit.  The focus was on controller use case immediately after commissioning but should offer improvement across the board.

Top-line stats (heap only):

* 20 devices commissioned: 66MB -> 57.3 MB (13.2% reduction)
* 100 devices commissioned: 125.1MB -> 89.8 MB (28.2% reduction)


lauckhart:
> Some test timeouts with 600_000 ?

Yeah those aren't really tests, they capture leak analysis or heap dumps.  I implemented as tests because the infrastructure is convenient and theoretically we could put limits in place to make them pass/fail.  But right now they just run simulations and record heap dumps and/or memlab analysis.  And can be slow if e.g. we simulate a controller with hundreds of nodes.

PR #3423, opened 2026-03-18
Fix unhandled rejection crash during parallel PASE commissioning

Apollon77:
## Summary

Fixes a crash (unhandled promise rejection killing the Node.js process) during BLE and multi-admin commissioning when multiple devices match the same short discriminator.

**Root cause:** `ParallelPaseDiscovery.registerAttempt()` wraps each commissioning attempt in a promise with `.finally(() => pending.delete(attempt))`. When one device wins PASE and the abort signal cancels the loser, the loser's promise rejects and `.finally()` eagerly removes it from `#pending` — before `onComplete()` can drain it via `allSettled()`. The rejected promise has no handler attached, so Node.js detects it as an unhandled rejection and crashes.

**Fix:** Two `void attempt.catch(() => {})` calls in `ParallelPaseDiscovery`:
- On `attempt` — marks losing attempts as "handled" for Node.js's rejection tracking. Winner errors still propagate through `#winnerPromise` (which is a separate `.then()` chain, awaited in `onComplete()`).
- On `#winnerPromise` — defensive, prevents unhandled rejection if the winner's commissioning rejects before `onComplete()` awaits it.

**Crash scenario (from matterjs-server#394):**
1. Discovery finds 2+ devices matching short discriminator 9 (4-bit, so collisions are common)
2. `CommissioningDiscovery.onDiscovered()` calls `registerAttempt()` for each → parallel `commission()` calls
3. Device B wins PASE via BLE → `winOnPase()` fires abort signal
4. Device A's `commission()` gets aborted → its promise rejects
5. `.finally()` removes Device A from `#pending` before `onComplete()` runs
6. Nobody catches Device A's rejection → **unhandled rejection → process crash**

## Changes

- `packages/node/src/behavior/system/controller/discovery/ParallelPaseDiscovery.ts` — the fix (2 lines)
- `packages/protocol/test/peer/CommissioningConnectionTest.ts` — new tests for external abort semantics
- `packages/node/test/behavior/system/controller/discovery/ParallelPaseRaceTest.ts` — standalone race pattern test that replicates the exact `registerAttempt`/`onComplete` logic and verifies:
  - **WITHOUT fix**: losing attempts produce unhandled rejections (the bug)
  - **WITH fix**: zero unhandled rejections
  - Winner errors still propagate through `onComplete()`
  - Multiple losers are all handled cleanly

Fixes https://github.com/matter-js/matterjs-server/issues/394


packages/node/src/behavior/system/controller/discovery/ParallelPaseDiscovery.ts
```diff
@@ -64,13 +64,20 @@ export abstract class ParallelPaseDiscovery<W> extends Discovery<W> {
             this.#winnerPromise = attempt.then(result => {
                 this.#winner = extractWinner(result);
             });
+            // Prevent unhandled rejection if the winner rejects before onComplete awaits
+            void this.#winnerPromise.catch(() => {});
```
lauckhart: This means that the error would be swallowed if onComplete isn't invoked.

What does "complete" mean when onComplete is invoked?  It seems like it's just wrapping #winnerPromise in another try/catch.  Could this just be collapsed into a single promise that is caught if there's an error and resolves if not?  Then we wouldn't need any intermediate "make this promise handled and I promise I really do handle it" shenanigans.

packages/node/src/behavior/system/controller/discovery/ParallelPaseDiscovery.ts
```diff
@@ -64,13 +64,20 @@ export abstract class ParallelPaseDiscovery<W> extends Discovery<W> {
...
         };
 
         attempt = Promise.resolve(factory(winOnPase)).finally(() => {
             this.#pending.delete(attempt);
         });
 
+        // Prevent unhandled rejection for losing attempts that settle (and get removed from
+        // #pending by .finally()) before onComplete drains #pending via allSettled.  Winner
+        // errors still propagate through #winnerPromise which is awaited in onComplete.
+        void attempt.catch(() => {});
```
lauckhart: I really think we should avoid burying errors.  E.g. never catch `error: any` and ignore.  In this case if an attempt throws ideally we have something along the lines "oh there's an error?  well if I'm aborted and it's this or this error, ignore, otherwise rethrow".

So we might bury expected errors in specific circumstances...  But nothing this broad

packages/node/src/behavior/system/controller/discovery/ParallelPaseDiscovery.ts
```diff
@@ -53,23 +54,35 @@ export abstract class ParallelPaseDiscovery<W> extends Discovery<W> {
...
+        attempt = Promise.resolve(factory(winOnPase))
+            .catch(error => {
+                if (isWinner) {
+                    // Winner's error is meaningful — must propagate to onComplete
+                    throw error;
+                }
+                // Loser: error is an expected side effect of the parallel race (cancellation,
+                // timeout, abort).  Resolve to prevent an unhandled rejection — the race
+                // semantics mean only the winner's result matters.
+                return undefined;
```
lauckhart, line 87: How do we know it's expected?  Could we have an "attempt aborted" exception, then log everything else?

Apollon77: Updated — the `.catch()` now distinguishes three cases by checking `isWinner` and `this.#abort.signal.aborted`:

**Winner** → rethrow (error propagates to `onComplete`)

**Loser after abort fired** (someone else won) — expected, resolve silently:
1. `AbortedError` — abort signal cancelled `CommissioningConnection`
2. `CommissioningError("Commissioning cancelled: another device was already successfully connected")` — loser's PASE succeeded but `continueCommissioningAfterPase` returned false
3. `PeerCommunicationError("Could not connect to device")` — PASE timed out before abort took effect

**Loser before abort fired** (no winner yet) — logged at info, then resolved:
4. `UnexpectedDataError` (wrong passcode) — different device, wrong credentials
5. Any unexpected error (bugs, network stack issues)

Cases 1–3 are silent (abort already fired = expected cancellation). Cases 4–5 get `logger.info("Parallel attempt failed:", error)` so they're visible in logs but don't crash the process and we see also full stack for the beginning to see relevant cases.

PR #3424, opened 2026-03-19
Matter 1.5 Update

Apollon77:
This PR updates the model to Matter 1.5.0, adds needed spec-fix overrides and regenerates all classes. For changes see changelog.

TODO is that we need to add code logic to validate the "Multi endpoint structure requirements" for some device types that was introduced by 1.5.0

The changes are added as single commits


packages/model/src/standard/resources/occupancy-sensor-type-bitmap.resource.ts
```diff
@@ -0,0 +1,23 @@
...
+Resource.add({
+    tag: "datatype", name: "OccupancySensorTypeBitmap", xref: "cluster§2.7.5.2",
+    details: "> [!NOTE]" +
+        "\n" +
+        "> This enum is as defined in ClusterRevision 4 and its definition shall NOT be extended; the feature " +
+        "flags provide the sensor modality (or modalities) for later cluster revisions. See Backward " +
+        "Compatibility section.",
+    children: [
+        { tag: "field", name: "Pir", description: "Indicates a passive infrared sensor." },
+        { tag: "field", name: "Ultrasonic", description: "Indicates a ultrasonic sensor." },
```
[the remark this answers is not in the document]
Apollon77: Generated code!

packages/model/src/standard/resources/occupancy-bitmap.resource.ts
```diff
@@ -0,0 +1,20 @@
...
+
+import { Resource } from "#models/Resource.js";
+
+Resource.add({
+    tag: "datatype", name: "OccupancyBitmap", xref: "cluster§2.7.5.1",
+
+    children: [{
+        tag: "field", name: "Occupied", description: "Indicates the sensed occupancy state",
+        xref: "cluster§2.7.5.1.1",
+        details: "If this bit is set, it shall indicate the occupied state else if the bit if not set, it shall " +
```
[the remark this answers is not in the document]
Apollon77: Generated code!

packages/model/src/standard/elements/ota-software-update-provider.element.ts
```diff
@@ -40,7 +40,7 @@ export const OtaSoftwareUpdateProvider = Cluster(
     Command(
         { name: "QueryImageResponse", id: 0x1, conformance: "M", direction: "response" },
         Field({ name: "Status", id: 0x0, type: "StatusEnum", conformance: "M" }),
-        Field({ name: "DelayedActionTime", id: 0x1, type: "uint32", conformance: "Status == Busy" }),
+        Field({ name: "DelayedActionTime", id: 0x1, type: "uint32", conformance: "Status == Busy, O", default: 0 }),
```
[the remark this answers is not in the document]
Apollon77: Generated code!

packages/model/src/standard/elements/iceserver-struct.element.ts
```diff
@@ -0,0 +1,23 @@
...
+
+/*** THIS FILE IS GENERATED, DO NOT EDIT ***/
+
+import { MatterDefinition } from "../MatterDefinition.js";
+import { DatatypeElement as Datatype, FieldElement as Field } from "../../elements/index.js";
+
+export const IceServerStruct = Datatype(
+    { name: "ICEServerStruct", type: "struct" },
+    Field(
+        { name: "UrLs", id: 0x0, type: "list", conformance: "M", constraint: "max 10[max 2000]" },
```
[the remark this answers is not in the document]
Apollon77: Generated code!

packages/model/src/standard/resources/doorbell.resource.ts
```diff
@@ -0,0 +1,19 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/*** THIS FILE IS GENERATED, DO NOT EDIT ***/
+
+import { Resource } from "#models/Resource.js";
+
+Resource.add({
+    tag: "deviceType", name: "Doorbell", xref: "device§16.9",
+    details: "A Doorbell device is a switch which when pressed usually causes a Section 16.7, “Chime” to activate.",
```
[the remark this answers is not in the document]
Apollon77: Generated code!

packages/model/src/logic/cluster-variance/InferredComponents.ts
```diff
@@ -153,6 +153,23 @@ const VarianceMatchers: VarianceMatcher[] = [
         },
     },
 
+    // [FOO<& BAR>*].[a-z][0-9-]*[+-]? (optional when features enabled; choice conformance suffix per spec
```
lauckhart: claude please be less verbose; parenthetical example would do

packages/model/src/logic/cluster-variance/InferredComponents.ts
```diff
@@ -153,6 +153,23 @@ const VarianceMatchers: VarianceMatcher[] = [
...
+    // section 7.3.14: C.an[+|-] where 'a' is the choice set letter, n is optional count, +/- optional range)
+    {
+        pattern: pattern("[", CONJUNCT_FEATURES, "]", "\\.[a-z][0-9-]*[+\\-]?"),
+        processor(add, match) {
+            add(true, { allOf: splitConjunction(match[0]) });
+        },
+    },
+
+    // [FOO & !BAR]Why we ned b*
```
lauckhart: claude please be less...  scrambly? ;)

support/codegen/src/mom/spec/repairs/aspect-repairs.ts
```diff
@@ -77,6 +77,10 @@ export function repairConformanceRule(conformance?: string) {
     // Null is specified with inconsistent case ("MfgCode != NULL" in cluster 1.4.2, lowercase elsewhere)
     conformance = conformance.replace(/= null/gi, "= null");
 
+    // Spec prefixes command name with cluster abbreviation in conformance references (e.g. "WebRTCProvideOffer"
```
lauckhart: please mention version

support/codegen/src/mom/spec/html-translators.ts
```diff
@@ -130,8 +130,15 @@ export const ConstraintStr = (el: HTMLElement) => {
     return parts.join(" ");
 };
 
-/** String with no space at all */
-export const NoSpace = (el: HTMLElement) => Str(el).replace(/\s/g, "");
+/** String with no space at all.  Cleans up verbose Asciidoctor cross-reference link text */
+export const NoSpace = (el: HTMLElement) => {
```
lauckhart: NoSpace is a pretty specific name, if we don't need a true `NoSpace` let's rename to something more appropriate, otherwise let's create a separate function

support/codegen/src/mom/spec/load-devices.ts
```diff
@@ -54,12 +54,35 @@ function augmentDevice(device: DeviceReference, content: HtmlReference) {
             device.composingTypes = content;
             break;
 
-        case "cluster requirements on composing device types":
+        case "cluster requirements on component device types":
+        case "cluster requirements on composing device types": // older spec terminology
```
lauckhart: "Pre 1.5 spec terminology"?

support/codegen/src/mom/spec/translate-cluster.ts
```diff
@@ -45,6 +45,17 @@ export function* translateCluster(definition: ClusterReference) {
 
     const metadata = translateMetadata(definition, children);
     if (!metadata) {
+        // No cluster IDs but may have shared datatypes (e.g. WebRTC Transport common section).
+        // Emit datatypes as standalone globals
+        if (definition.datatypes?.length) {
```
lauckhart: Types defined in a cluster section should remain scoped to that cluster regardless of whether it has an ID.  Derived clusters will inherit.  If non-derived clusters reference them it is a spec bug and automatic scoping of name should kick in.  Regardless, they are not globals

support/codegen/src/mom/spec/translate-datatype.ts
```diff
@@ -217,16 +217,29 @@ export function translateValueChildren(
...
                 conformance: Optional(ConformanceCode),
                 description: Optional(Alias(StrWithSuperscripts, "summary", "notes")),
                 meaning: Optional(Str),
-            });
+            };
 
+            let records = translateTable("value", definition, schema);
             records = records.filter(r => r.name !== "Reserved");
 
+            // If the first table produced no valid records (e.g. range-based enum like ClosureErrorEnum),
```
lauckhart: Please make this less likely to accidentally exhibit unintended behavior (e.g. instead of current "one table produces nothing so look at second table" could change to "skip tables with only value ranges")

support/codegen/src/mom/spec/translate-device.ts
```diff
@@ -154,29 +154,79 @@ function addConditions(device: DeviceTypeElement, deviceRef: DeviceReference) {
...
+        condition: Identifier,
+        conformance: Optional(ConformanceCode),
+    });
+
+    if (!records.length) {
+        return;
+    }
+
+    // Build a lookup from condition name to detail section for documentation extraction
+    const detailLookup = new Map<string, HtmlReference>();
```
lauckhart: This doesn't appear to be doing anything different from normal detail section mapping, is there a reason not to use the implementation shared by every other element?

support/codegen/src/mom/spec/translate-global.ts
```diff
@@ -91,7 +91,7 @@ function* translateDatatypes(ref: GlobalReference): Generator<DatatypeElement> {
                     detail.name = "semtag";
                     break;
 
-                case "LocationDescriptorStruct":
```
lauckhart, line 94: Please try to preserve support for old documents

support/codegen/src/mom/spec/translate-namespace.ts
```diff
@@ -24,9 +24,6 @@ export function* translateNamespace(nsRef: HtmlReference) {
         throw new Error("Multiple namespace IDs detected");
     }
 
-    const props = ids[0];
-    props.name = props.name.replace(/^Common/, "");
```
lauckhart, line 28: Why was this removed?

support/codegen/src/util/finalize-model.ts
```diff
@@ -116,7 +118,7 @@ function childrenIdentity(model: ValueModel) {
  * exactly one other cluster.  When detected, replace the datatype name with a qualified name referencing the other
  * cluster.
  */
-function patchIllegalCrossClusterReferences(cluster: ClusterModel, scopedDatatypes: ScopedDatatypes) {
+function patchIllegalCrossClusterReferences(cluster: Model, scopedDatatypes: ScopedDatatypes) {
```
lauckhart: This may be an artifact of incorrectly promoting datatypes to global which might cause global datatypes to reference datatypes in clusters.  So either:

1. Revert this change if it is not needed once datatypes are properly scoped, or
2. Rename function and first argument if this is in fact needed, and would be nice to indicate what non-cluster conditions are that it is required

packages/model/src/standard/elements/battery-storage.element.ts
```diff
@@ -13,66 +13,100 @@ export const BatteryStorageDt = DeviceType(
...
         ),
 
+        Requirement(
+            { name: "ElectricalEnergyMeasurement", id: 0x91, conformance: "M", element: "serverCluster" },
+            Requirement({ name: "EXPORTEDENERGY", conformance: "M", element: "feature" })
+        ),
         Requirement(
             { name: "Descriptor", id: 0x1d, element: "serverCluster" },
-            Requirement({ name: "TagList", conformance: "M", element: "feature" })
+            Requirement({ name: "TAGLIST", conformance: "M", element: "feature" })
```
lauckhart, line 36: What happened to case here?  Did they really rename TagList to TAGLIST?  Seems to have happened in a number of places e.g. TAGLIST, ALTERNATINGCURRENT, DIRECTCURRENT

lauckhart (commented):
Took a pass through the non-generated changes.  Everything I noted is small stuff except for what looks like "improperly promoting datatype to global".  Only checked a couple of generated files, noted strange change of case in requirements

support/codegen/src/mom/spec/translate-datatype.ts
Apollon77: Changed the approach: instead of falling through when a table produces no records, we now explicitly detect and skip range-only tables by checking if the first row's value column matches a range pattern (e.g. "0x00 to 0x7F"). This avoids accidentally consuming the wrong table.

support/codegen/src/mom/spec/translate-namespace.ts
Apollon77: Pre-1.5 the spec used names without the "Common" prefix, so we stripped it. In 1.5 the spec intentionally renamed these namespaces to include the prefix (e.g. "Common Area"), so the stripping was removed. Old specs that don't have the prefix won't be affected since there's nothing to strip.

packages/model/src/standard/elements/battery-storage.element.ts
Apollon77: Feature names in device type requirements are uppercased to match the feature code (e.g. `TAGLIST`, `ALTERNATINGCURRENT`, `DIRECTCURRENT`). This is consistent with how `addClusters` has always handled features in regular device requirements (see on-off-light: `LIGHTING`, `ONOFF`). The old composing code just didn't apply the same uppercasing — the new `addComposing` code fixes that inconsistency. The cluster definition itself uses the PascalCase `title` (`TagList`) while the ALL_CAPS version is the feature code used in conformance matching.

support/codegen/src/mom/spec/translate-device.ts
Apollon77: Fixed — switched to using `translateTable`'s built-in `installPreciseDetails` by renaming the `condition` field to `name` in the schema and using tag `"condition"` so the standard "${name} ${tag}" matching finds the detail sections (e.g. "ManagedAclAllowed Condition"). No more custom detail mapping.

PR #3431, opened 2026-03-19
De-TLV part 1

lauckhart:
This is a major refactoring of the typing and runtime TLV configuration for @matter/node.  It saves memory and simplifies type hints.

Current memory savings are ~5MB (10% of heap) for devices and simple controllers.

These changes should be largely compatible with the current public API except for edge cases.  The primary exception is that we don't export clusters from the main barrel export of @project-chip/matter.js.  That was necessary to prevent loading of all clusters into memory unconditionally.

Depending on the level of breaking changes we want to accept to semi-internal structures (namely dropping the old cluster definitions) we can save additional memory and reduce code complexity significantly.

This is a "big ball of mud" change.  I broke up into logical commits to make it easier to tell what's going on, but only the final commit actually compiles.

Depends on #3425, #3426, #3427, #3428, #3429 and #3430.  Those are all relatively small though; this isn't really decomposable.

### DIFF
--- a/examples/device-onoff-advanced/src/cluster/MyFancyOwnFunctionality.ts
+++ b/examples/device-onoff-advanced/src/cluster/MyFancyOwnFunctionality.ts
@@ -162,20 +162,23 @@ export type MyFancyCommandRequest = TypeFromSchema<typeof MyFancyOwnFunctionalit
 export type MyFancyCommandResponse = TypeFromSchema<typeof MyFancyOwnFunctionality.TlvMyFancyCommandResponse>;
 export type MyFancyEvent = TypeFromSchema<typeof MyFancyOwnFunctionality.TlvMyFancyEvent>;
 
-export namespace MyFancyOwnFunctionalityInterface {
+export interface MyFancyOwnFunctionalityNamespace {
+    Commands: {
+        Base: MyFancyOwnFunctionalityNamespace.Base;
+        Components: [{ flags: {}; methods: MyFancyOwnFunctionalityNamespace.Base }];
+    };
+}
+
+export namespace MyFancyOwnFunctionalityNamespace {
     export interface Base {
         myFancyCommand(request: MyFancyCommandRequest): MaybePromise<MyFancyCommandResponse>;
     }
 }
 
-export type MyFancyOwnFunctionalityInterface = {
-    components: [{ flags: {}; methods: MyFancyOwnFunctionalityInterface.Base }];
-};
-
 /**
  * Behavior definition for the cluster command handler implementation
  */
-export const MyFancyOwnFunctionalityBehavior = ClusterBehavior.withInterface<MyFancyOwnFunctionalityInterface>().for(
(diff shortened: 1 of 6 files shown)


packages/node/src/behavior/cluster/ValidatedElements.ts
```diff
@@ -206,17 +200,18 @@ export class ValidatedElements {
...
-        for (const name in commands) {
-            const command = commands[name];
-            if (!command) {
-                this.error(`cluster.commands.${name}`, "Undefined element in cluster definition", true);
+        for (const member of this.#schema.conformant.commands) {
+            if (member.isResponse) {
                 continue;
             }
 
+            const name = camelize(member.name);
```
Apollon77: propertyName? :-)

packages/node/src/behavior/cluster/ValidatedElements.ts
```diff
@@ -271,15 +268,13 @@ export class ValidatedElements {
...
         }
 
-        for (const name in expected) {
-            const event = expected[name];
-            if (!event) {
-                this.error(`cluster.events.${name}`, "Undefined element in cluster definition", true);
-                continue;
-            }
+        for (const member of this.#schema.conformant.events) {
+            const name = camelize(member.name);
```
Apollon77: propertyName

packages/node/src/behavior/internal/ServerBehaviorBacking.ts
```diff
@@ -101,29 +101,50 @@ export class ServerBehaviorBacking extends BehaviorBacking {
...
+        // Update attribute list from validated elements (includes global attributes via scope)
+        globals.attributeList = ([...validation.attributeIds.values()] as AttributeId[]).sort((a, b) => a - b);
 
-        // Validate the feature map
+        // Update accepted & generated command lists from the schema (ClusterModel)
         const schema = Schema(behavior.type) as ClusterModel;
+        const commandModels = new Map<string, CommandModel>();
+        for (const cmd of schema.commands) {
+            if (cmd.isRequest) {
+                commandModels.set(camelize(cmd.name), cmd);
```
Apollon77: propertyName? ... and some more cases below

packages/types/src/cluster/ClusterTypeOfModel.ts
```diff
@@ -0,0 +1,146 @@
...
+    // Attributes — include all (conformant + conditional + deprecated-with-type), skip globals and disallowed.
+    // Deprecated attributes without a type (e.g. WindowCovering VelocityLift) are excluded since they have no
+    // schema and the codegen also omits them.
+    const attributes = {} as Record<string, ClusterType.Attribute>;
+    for (const attr of model.attributes) {
+        if (GLOBAL_IDS.has(attr.id) || attr.isDisallowed || attr.effectiveMetatype === undefined) {
+            continue;
+        }
+
+        const key = camelize(attr.name);
```
Apollon77: propoertyName? ... also below in this file more cases

PR #3437, opened 2026-03-21
Fix BDX session leak when processTransfer fails with non-BdxError

Apollon77:
## Summary

- Fixes a session leak in `BdxSession.processTransfer()` where non-BdxError exceptions (e.g. `[aborted]` from peer-unresponsive timeout) cause `BdxError.accept()` to rethrow before `close()` is called
- The orphaned session stays permanently in `BdxProtocol.#activeBdxSessions`, blocking all future OTA announcements for that peer with "BDX transfer still active, skipping announcement" until server restart
- Uses `BdxError.is()` to branch instead of `BdxError.accept()`, ensuring `close()` is always called regardless of error type

Addresses: matter-js/matterjs-server#407

## Test plan

- [ ] Verify existing BDX tests still pass
- [ ] Confirm that a BDX transfer aborted by peer-unresponsive timeout properly cleans up the session from `#activeBdxSessions`
- [ ] Verify OTA retry announcements proceed after a failed transfer instead of being blocked indefinitely

### DIFF
--- a/packages/node/test/node/ClientTuningTest.ts
+++ b/packages/node/test/node/ClientTuningTest.ts
@@ -56,7 +56,7 @@ describe("ClientTuningTest", () => {
         // Unspecified fields should retain their defaults
         expect(timing.delayAfterPeerError).equals(PeerTimingParameters.defaults.delayAfterPeerError);
         expect(timing.delayAfterUnhandledError).equals(PeerTimingParameters.defaults.delayAfterUnhandledError);
-        expect(timing.minimumTimeBetweenMrpKicks).equals(PeerTimingParameters.defaults.minimumTimeBetweenMrpKicks);
+        expect(timing.kickThrottleInterval).equals(PeerTimingParameters.defaults.kickThrottleInterval);
     });
 
     it("custom profiles propagate to NetworkProfiles", async () => {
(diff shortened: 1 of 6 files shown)


packages/protocol/src/bdx/BdxSession.ts
```diff
@@ -137,15 +137,18 @@ export class BdxSession {
...
-            BdxError.accept(error);
-            await this.#messenger.sendError(error.code);
+            if (BdxError.is(error)) {
+                await this.#messenger.sendError(error.code);
 
-            logger.warn(`BDX session failed with error:`, error);
+                logger.warn(`BDX session failed with error:`, error);
 
-            await this.close(error);
+                await this.close(error);
```
lauckhart: Maybe just do the close outside the if/else entirely?

packages/protocol/src/bdx/BdxSession.ts
```diff
@@ -137,15 +137,18 @@ export class BdxSession {
 
             await this.close();
         } catch (error) {
-            BdxError.accept(error);
-            await this.#messenger.sendError(error.code);
+            if (BdxError.is(error)) {
+                await this.#messenger.sendError(error.code);
```
lauckhart: If this throws do we still close?  Or log the original error?

PR #3438, opened 2026-03-21
Command response conformance validation with cross-command field references -> 1.5

Apollon77:
## Summary

- Adds conformance validation for command responses through the existing `RootSupervisor` pipeline, the same way attributes are validated
- Resolves cross-command field references in conformance (e.g. `SolicitOffer.VideoStreamID` on response fields) — a new spec 1.5 pattern where response field presence depends on request fields
- Extends `ValidationLocation` with `requestData` to pass the decoded request payload when validating responses
- Adds `requestModel`/`responseModel` to `CommandTypeProtocol` so the validation pipeline has access to command schemas

### Details

**Model validation** (`ValueValidator.ts`): Dotted conformance names like `SolicitOffer.VideoStreamID` are resolved by finding the command in the owning cluster and looking up the field. Eliminates 4 `UNRESOLVED_CONFORMANCE_NAME` validation errors.

**Runtime conformance** (`conformance-compiler.ts`): Cross-command references resolve from `ValidationLocation.requestData` when available, falling back to optional when no request context is present.

**Response validation** (`ProtocolService.ts`): After a command handler returns, the response is validated through `RootSupervisor.get(responseModel).validate()` with the request data as context. Currently warn-only to avoid breaking existing implementations that may not fully comply.

**Request validation**: Stubbed as commented-out TODO — needs review of existing implementations before activation.

## Test plan

- [x] Model test verifies cross-command conformance references resolve without errors
- [x] Model test verifies `SolicitOfferResponse.videoStreamId` conformance parses as `SolicitOffer.VideoStreamID`
- [x] Full test suite passes (all packages, ESM/CJS/Web)
- [ ] Verify OTA response conformance warnings are expected (conditional fields)


packages/model/src/logic/definition-validation/ValueValidator.ts
```diff
@@ -27,6 +27,17 @@ export class ValueValidator<T extends ValueModel> extends ModelValidator<T> {
 
         this.#validateAspect("conformance");
         this.model.conformance.validateReferences(this, name => {
+            // Cross-command field reference (e.g. "SolicitOffer.VideoStreamID")
```
lauckhart: Qualified name resolution should be a function of model traversal, and ideally share implementation with the qualified resolution algorithm used for types

packages/model/src/logic/definition-validation/ValueValidator.ts
```diff
@@ -27,6 +27,17 @@ export class ValueValidator<T extends ValueModel> extends ModelValidator<T> {
...
+            if (name.includes(".")) {
+                const [commandName, fieldName] = name.split(".", 2);
```
lauckhart: Qualified name resolution should allow for arbitrary depth

Conformance is general-purpose.  There shouldn't be an assumption about the type of a name, qualified or not

If there is element-specific specialization that should go into the validator for that element.  E.g. in this case, CommandValidator might allow for conformance on requests to resolve request fields

packages/node/src/behavior/state/validation/conformance-compiler.ts
```diff
@@ -71,6 +71,28 @@ export function astToFunction(schema: ValueModel, supervisor: RootSupervisor): V
     // Name resolution scope may change as we visit the AST; the base version creates a name reference if the the name
     // is visible in scope.  Other nodes may temporarily override this though
     let createNameReference = (name: string): DynamicNode => {
+        // Cross-command field references (e.g. "SolicitOffer.VideoStreamID") resolve from the
+        // request payload passed via ValidationLocation.requestData
+        if (name.includes(".")) {
+            const [, fieldName] = name.split(".", 2);
```
lauckhart: This has the same problems as definition validation, but is worse in that it ignores the first segment of the identifier so has different, fairly arbitrary semantics.  Definition validation and the conformance compiler should be using the same name resolution implementation

The conformance compiler should be resolving the qualified name rather than ignoring it

packages/node/src/behavior/state/validation/conformance-compiler.ts
```diff
@@ -71,6 +71,28 @@ export function astToFunction(schema: ValueModel, supervisor: RootSupervisor): V
...
+            const key = camelize(fieldName);
+            return {
+                code: Code.Evaluate,
+
+                evaluate: (_value, options) => {
+                    // If requestData is available, resolve the field from it
+                    if (options?.requestData !== undefined) {
```
lauckhart: "requestData" is specific to a subset of a certain element type and doesn't belong here

Name resolution should be handled by the name resolver.  I don't understand the full context but it doesn't feel like there's a need to break encapsulation like this

packages/node/src/behavior/state/validation/conformance-compiler.ts
```diff
@@ -71,6 +71,28 @@ export function astToFunction(schema: ValueModel, supervisor: RootSupervisor): V
...
+                        return {
+                            code: Code.Value,
+                            value: (options.requestData as Val.Struct)[key],
+                        };
+                    }
+                    // Without request context, treat as optional (safe default)
+                    return { code: Code.Optional };
```
lauckhart: The fallback for unresolved qualified names should be identical to that of unqualified names

packages/protocol/src/action/protocols.ts
```diff
@@ -234,6 +234,16 @@ export interface CommandTypeProtocol extends AddressableElementProtocol<CommandI
      * Access control information for the command.
      */
     limits: AccessControl.Limits;
+
+    /**
+     * The model for the request command, used for conformance validation.
+     */
+    requestModel?: Schema;
```
lauckhart: This is a low-level interface for dealing with wire concerns.  Conformance validation is an application-level concept...  So I don't think this belongs here

lauckhart (changes requested):
I tried to provide enough information to help Claude architect this properly but I'm not sure if it'll be enough...  If you want I can walk him through implementation on my side

PR #3444, opened 2026-03-22
Add session IP address monitoring with probe-based validation

Apollon77:
## Summary

- When mDNS passively detects address changes while a session is active, the Peer debounces for 10s, checks if the session's current IP is still in discovered addresses, and probes with an empty read to verify reachability
- Adds `addressOverride` plumbing through the entire send path (per-exchange, not per-session) so probes can target specific addresses without redirecting other traffic
- Adds `probe()` method to `ClientInteraction` with `requireExistingSession` — verifies session liveness without triggering reconnection
- Adds `probeAddress` network sub-profile for independent probe throttling
- Refactors `SustainedSubscription` to use `probe()` instead of inline read
- Refactors `MessageChannel.send()` to use an options object (`MessageChannelSendOptions`)

### DIFF
--- a/packages/general/src/storage/StorageService.ts
+++ b/packages/general/src/storage/StorageService.ts
@@ -111,7 +111,9 @@ export class StorageService {
         const cached = this.#openDrivers.get(cacheKey);
         if (cached) {
             cached.refs++;
-            return new StorageManager(new StorageDriverHandle(cached.driver, () => this.#release(cacheKey)));
+            const manager = new StorageManager(new StorageDriverHandle(cached.driver, () => this.#release(cacheKey)));
+            await manager.initialize();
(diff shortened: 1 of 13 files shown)


packages/protocol/src/peer/Peer.ts
```diff
@@ -66,6 +68,8 @@ export class Peer {
     #observers = new ObserverGroup();
     #exchangeProvider?: ExchangeProvider;
     #updated = AsyncObservable<[peer: Peer]>();
+    #addressCheckTimer: Timer;
+    #probing?: Promise<void>;
```
lauckhart: Maybe put all of the probing stuff into a single utility class that focuses on that concern

PR #3445, opened 2026-03-22
Fix parsing error in parser -> 1.5

Apollon77:


support/codegen/src/mom/spec/translate-device.ts
```diff
@@ -254,6 +254,17 @@ function addClusters(device: DeviceTypeElement, deviceRef: DeviceReference) {
         }
     }
 
+    // Also index implicitly-created cluster requirements (e.g. Descriptor) so element
+    // requirements referencing them can be attached
+    if (device.children) {
```
lauckhart: Can requirements be inherited?  Should this maybe using scope members?  Typically that would be a getter that uses scope.membersOf() internally, not sure if it's implemented for devicetype requirements, I usually add as-needed

lauckhart (approved):

Apollon77:
Re: cluster index comment — changed to build the index from all `device.children` that are server/client cluster requirements, rather than only from the table-parsed clusters + a separate implicit fixup. This naturally includes both inherited and implicitly-created entries like Descriptor.

PR #3458, opened 2026-03-23
De-TLV part2

lauckhart:
Followup to #3431. Subsumes part 1 (commits 1-4) so we could potentially just merge this one. I'd like to avoid squashing to preserve the evolution history.

This PR removes all legacy mechanics from the cluster namespaces and heavily optimizes both the type system and runtime implementation.

On the type side we now generate optimized .d.ts files directly. Definitions are simpler, smaller, look closer to "normal" javascript definitions and work better in the IDE.

On the runtime side everything is built lazily from the model. The .js files corresponding to above have two statements, one of which is a deprecated compatibility alias.

What this PR doesn't do is introduce major breaking changes to public API. What has been removed is lower-level "advance use" stuff that is likely unused externally. Namely:

* ClusterRegistry is gone
* We still support static definitions of clusters but I've stripped out the mutation support and supporting logic
* Statically generated TLV is gone
* Runtime data for command/attribute/event objects is limited to ID, name and schema (model)

Support for static definitions is for folks that have defined custom clusters the old way. Dropping those and the compile-time typing from TlvSchema would be step 3.

### DIFF
--- a/packages/node/src/behaviors/level-control/LevelControlServer.ts
+++ b/packages/node/src/behaviors/level-control/LevelControlServer.ts
@@ -57,8 +57,8 @@ const LevelControlBase = LevelControlBehavior.with(LevelControl.Feature.OnOff, L
  */
 export class LevelControlBaseServer extends LevelControlBase {
     declare protected internal: LevelControlBaseServer.Internal;
-    declare state: LevelControlBaseServer.State;
-    declare events: LevelControlBaseServer.Events;
+    declare readonly state: LevelControlBaseServer.State;
+    declare readonly events: LevelControlBaseServer.Events;
 
     /** Returns the minimum level, including feature specific fallback value handling. */
     get minLevel() {
--- a/packages/node/src/behavior/cluster/ClusterBehavior.ts
+++ b/packages/node/src/behavior/cluster/ClusterBehavior.ts
@@ -296,6 +296,16 @@ export namespace ClusterBehavior {
         B extends Behavior.Type = Behavior.Type,
         N extends ClusterTyping = ClusterInterface.InterfaceOf<B>,
         NS extends ClusterNamespace = ClusterNamespace.Concrete,
+        // I and ID are redundant — they duplicate information already derivable from B.  They exist because tsc < 6
+        // deeply expands B's structure when it encounters indexed access (B["Internal"]) or infer...extends
+        // (B extends { id: infer S extends ... }) during declaration emit.  This transitively triggers contravariant
+        // checking on N, breaking `.with().alter()` chains with TS2684.  Pre-computing as separate parameters avoids
+        // the expansion.  tsgo handles this correctly; tsc 6 has not been tested.  Once tsc < 6 is no longer
+        // supported, these parameters can be inlined back into the property types.
+        I = B["Internal"], // inline: B["Internal"]
+        ID extends Uncapitalize<string> = B extends { readonly id: infer S extends Uncapitalize<string> } // inline: B extends { readonly id: infer S extends Uncapitalize<string> } ? S : Uncapitalize<string>
+            ? S
+            : Uncapitalize<string>,
     > {
         new (agent: Agent, backing: BehaviorBacking): Instance<B, N>;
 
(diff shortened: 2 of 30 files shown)


packages/matter.js/src/cluster/client/InteractionClient.ts
```diff
@@ -472,27 +470,27 @@ export class InteractionClient {
...
-        const event = getClusterEventById(cluster, eventId);
+        const clusterModel = Matter.clusters(clusterId);
+        const event = clusterModel?.events(eventId);
 
         return {
             endpointId,
             clusterId,
             eventId,
-            eventName: event?.name ?? `Unknown (${Diagnostic.hex(eventId)})`,
+            eventName: event ? camelize(event.name) : `Unknown (${Diagnostic.hex(eventId)})`,
```
Apollon77: propoertyName?

packages/matter.js/src/cluster/client/InteractionClient.ts
```diff
@@ -472,27 +470,27 @@ export class InteractionClient {
...
-        const attribute = getClusterAttributeById(cluster, attributeId);
+        const clusterModel = Matter.clusters(clusterId);
+        const attribute = clusterModel?.attributes(attributeId);
 
         return {
             endpointId,
             clusterId,
             attributeId,
-            attributeName: attribute?.name ?? `Unknown (${Diagnostic.hex(attributeId)})`,
+            attributeName: attribute ? camelize(attribute.name) : `Unknown (${Diagnostic.hex(attributeId)})`,
```
Apollon77: propertyName?

packages/matter.js/src/device/PairedNode.ts
```diff
@@ -621,9 +620,11 @@ export class PairedNode {
                     // We need to determine the attribute name for the API
                     let attributeName = this.#attributeIdToNameMap.get(attrKey);
                     if (attributeName === undefined) {
-                        const clusterDef = getClusterById(clusterId);
-                        const attributeDef = getClusterAttributeById(clusterDef, attributeId as AttributeId);
-                        attributeName = attributeDef?.name ?? `Unknown (${Diagnostic.hex(attributeId)})`;
+                        const clusterModel = Matter.clusters(clusterId);
+                        const attrModel = clusterModel?.attributes(attributeId);
+                        attributeName = attrModel
+                            ? camelize(attrModel.name)
```
Apollon77: propertyName?

packages/protocol/src/interaction/AttributeDataDecoder.ts
```diff
@@ -207,7 +204,7 @@ export function normalizeAttributeStatus(
             return;
         }
         result.push({
-            path: { nodeId, endpointId, clusterId, attributeId, attributeName: attributeDetail.name },
+            path: { nodeId, endpointId, clusterId, attributeId, attributeName: camelize(attributeModel.name) },
```
Apollon77: propertyName (and more down this file)

packages/types/test/cluster/ClusterTypeOfModelTest.ts
Apollon77, line 1: Deleted because?

packages/node/src/behaviors/level-control/LevelControlServer.ts
```diff
@@ -575,9 +571,9 @@ export class LevelControlBaseServer extends LevelControlBase {
     }
 
     #calculateEffectiveOptions(
-        optionsMask: TypeFromPartialBitSchema<typeof LevelControl.Options>,
```
Apollon77, line 578: Are "TypeFromPartialBitSchema" and such still there as types, potentially to just being "pass through types" for easier migration?

lauckhart: It's still there but unmodified, it's still used for typing from actual TLV definitions.  We could probably modify to be a passthrough type as a fallback if the input isn't TlvSchema, you think that's important?

packages/types/test/cluster/ClusterTypeOfModelTest.ts
lauckhart: Accidental

lauckhart: Oh nvm it wasn't accidental.  Removed because it compared ClusterTypeOfModel against the statically generated equivalents that no longer exist.

packages/node/src/behaviors/level-control/LevelControlServer.ts
lauckhart: OK nvm here too, there's no `LevelControl.Options` object anymore so passthrough wouldn't work anyway

PR #3469, opened 2026-03-26
Add default network profile for device nodes

Apollon77:
## Summary
- Adds `defaultProfile` property to `NetworkProfiles` — when set, peers with unknown physical properties use this instead of the conservative fallback
- `NetworkServer.State` gains `defaultNetworkProfile` for explicit override
- `ServerNetworkRuntime` auto-detects the profile at startup: if the node has application endpoints, it derives "fast" or "thread" from the root endpoint's `NetworkCommissioning` features

Devices using `Peers.forAddress` (OTA providers, subscription reestablishment) previously got the conservative profile (4 exchanges, 100ms delay) because `physicalProperties` aren't available yet. Once properties are set later, `forPeer` picks the real profile — this only covers the initial unknown period.

### DIFF
--- a/packages/node/src/behavior/system/network/ServerNetworkRuntime.ts
+++ b/packages/node/src/behavior/system/network/ServerNetworkRuntime.ts
@@ -8,6 +8,7 @@ import { NetworkCommissioningBehavior } from "#behaviors/network-commissioning";
 import type { ServerNode } from "#node/ServerNode.js";
 import { InteractionServer } from "#node/server/InteractionServer.js";
 import {
+    AddressInUseError,
     ConnectionlessTransport,
     ConnectionlessTransportSet,
     Crypto,
@@ -43,6 +44,8 @@ import { NetworkRuntime } from "./NetworkRuntime.js";
 import { ServerGroupNetworking } from "./ServerGroupNetworking.js";
 
 const logger = Logger.get("ServerNetworkRuntime");
+
+const MAX_PORT_ASSIGNMENT_RETRIES = 10;
 
 function convertNetworkEnvironmentType(type: string | number) {
     const convertedType: InterfaceType =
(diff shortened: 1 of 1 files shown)


packages/node/src/behavior/system/network/NetworkServer.ts
```diff
@@ -146,5 +146,17 @@ export namespace NetworkServer {
...
+         * The default network profile to use for peers with unknown physical properties.
+         *
+         * When undefined, the profile is auto-detected at startup: if the node has application endpoints it is treated
+         * as a device and the profile is derived from the local node's network capabilities (e.g. "fast" for WiFi or
+         * Ethernet, "thread" for Thread-only).  If no application endpoints are present it remains unset, preserving
+         * the conservative default.
+         *
+         * Set explicitly to override auto-detection, e.g. `"fast"` or `"unlimited"`.
+         */
+        defaultNetworkProfile?: string = undefined;
```
lauckhart: "Network" is the name is unnecessary and inconsistent with "profiles"

packages/protocol/src/peer/NetworkProfile.ts
```diff
@@ -64,6 +64,14 @@ export interface NetworkProfile extends ConcreteNetworkProfile {
 export class NetworkProfiles {
     #networks = new Map<string, NetworkProfile>();
     #defaults: NetworkProfiles.Templates;
+    /**
```
lauckhart: Blank line before this please

packages/protocol/src/peer/NetworkProfile.ts
```diff
@@ -64,6 +64,14 @@ export interface NetworkProfile extends ConcreteNetworkProfile {
...
+     * The default profile used for peers with unknown physical properties.
+     *
+     * When set, peers without physical properties use this profile instead of "conservative".  This is useful for
+     * devices that know their own network capabilities — e.g. a WiFi device should not throttle outgoing exchanges to
+     * peers it hasn't yet fully characterized.
+     */
+    defaultProfile?: string;
```
lauckhart: "fallback" might be better terminology as there's a lot of "defaults" floating around here already

Rather than creating a (partially implemented) config side channel, could we instead create a "fallback" network profile, using the conservative tempkate for defaults?

packages/protocol/src/peer/NetworkProfile.ts
```diff
@@ -64,6 +64,14 @@ export interface NetworkProfile extends ConcreteNetworkProfile {
 export class NetworkProfiles {
     #networks = new Map<string, NetworkProfile>();
     #defaults: NetworkProfiles.Templates;
+    /**
+     * The default profile used for peers with unknown physical properties.
+     *
+     * When set, peers without physical properties use this profile instead of "conservative".  This is useful for
```
lauckhart: I think I would state as `Defaults to "conservative".` in a separate paragraph

packages/protocol/src/peer/NetworkProfile.ts
```diff
@@ -163,8 +171,24 @@ export class NetworkProfiles {
             id = "group";
             defaults = this.#defaults.fast;
         } else if (pp === undefined) {
-            id = "unknown";
-            defaults = this.#defaults.conservative;
+            if (this.defaultProfile !== undefined) {
+                const template = this.#defaults[this.defaultProfile as keyof NetworkProfiles.Templates];
+                if (template === undefined) {
```
lauckhart: Config should be validated on write, not read, and misconfiguration should be fatal.

packages/protocol/src/peer/NetworkProfile.ts
```diff
@@ -163,8 +171,24 @@ export class NetworkProfiles {
             id = "group";
             defaults = this.#defaults.fast;
         } else if (pp === undefined) {
-            id = "unknown";
-            defaults = this.#defaults.conservative;
+            if (this.defaultProfile !== undefined) {
```
lauckhart: This all gets much simpler if we don't create a side channel for just one of the profiles

packages/protocol/src/peer/NetworkProfile.ts
lauckhart: Wait, this is already playing the role of the fallback profile.  Why do we need a separate mechanism for this?  If the issue is that we want a different fallback in different contexts then we should just split this into two with names specific to the contexts

Otherwise implementation is more complicated and semantics are confusing

PR #3470, opened 2026-03-26
Fix IPv4 UDP port mismatch: use same port as IPv6 for consistent operational port

Apollon77:
When no port is configured, IPv6 and IPv4 UDP sockets were each getting different OS-assigned random ports. The operationalPort (used in DNS-SD SRV records and by TCP) was set from the IPv6 port, causing IPv4 UDP to listen on a different port that peers couldn't discover.

Now IPv4 UDP reuses the operationalPort assigned to IPv6, ensuring all transports (IPv6 UDP, IPv4 UDP, TCP) share the same port number.


packages/node/src/behavior/system/network/ServerNetworkRuntime.ts
```diff
@@ -163,11 +163,13 @@ export class ServerNetworkRuntime extends NetworkRuntime {
 
         if (netconf.ipv4) {
             try {
+                // Use the same port as IPv6 to ensure consistent operational port across transports
+                const ipv4Port = this.owner.state.network.operationalPort || netconf.port;
```
lauckhart: Can we be sure the port assigned to the ipv6 IP is available?  Seems like we should maybe handle that case and recreate the sockets if there's a conflict

Apollon77: good catch

PR #3472, opened 2026-03-27
TCP: Add Matter TCP transport support

### LINKED ISSUE #40: Matter camera enabler - Matter JS 1.5.(1)
### Problem statement

As stated in the overarching roadmap opportunity #6 and the accompanying [Thoughts on a way forward with cameras in Home Assistant document](https://docs.google.com/document/d/11CO56Vdzc1R8lP9fskaPOO-ynyWhoaCAA5nczAEKxrU/edit?tab=t.0) we need to establish the technical foundation first to enable Matter camera support going forward.

Currently, work on the Matter JS server is progressing but we're still lacking full support for Matter cameras to proceed. This includes leveraging the Matter 1.5(.1) base schema and having a defined path to also support TCP to cover the full scope going forward.

### Scope & Boundaries

- In Scope
  - Basic-Cluster and DataModel support for Matter 1.5.0
  - Basic-Cluster and DataModel support for Matter 1.5.1
  - TCP Support with Matter 1.5.x
  - Having a first shot at a refined camera UI (ref. https://github.com/OpenHomeFoundation/ux-design/discussions/19)
  - Gather learnings and refine the way forward from a UI/UX perspective

- Out of scope
  - Having all Matter camera features already implemented and available to users to interact with
  - Focus on the technical grounds to establish that going forward

### Foreseen solution

Aim to have the technical foundation ready to work on integrating Matter cameras into UI flows in a consecutive step

### Community signals

We see first Matter cameras hitting the market.

While Matter-controller cameras are available for quite some time already native Matter camera support is missing but requested as an alternative to using e.g. Homekit integrations to establish core functionality.

### Risks & open questions

Not having TCP support just yet leaves us with a limited feature set but enough to proceed with further UX work.

### Appetite

- 4-6 weeks

### Execution issues

_No response_

### Decision log

| Date | Decision | Outcome |
|------|----------|---------|
|      |          |         |

Apollon77:
## Summary

Adds full TCP transport support to matter.js for both client and server side, following the Matter 1.5.1 specification. This replaces the draft PR #3465 with a clean, rebased commit history.

### Commit overview (8 commits):

1. **Generalize transport abstractions** — Rename `ConnectionlessTransport` → `Transport`, add `ChannelType` enum (UDP/TCP), `ServerAddressTcp` type
2. **TCP socket platform layer** — `TcpSocket`/`TcpServerSocket` interfaces, Node.js implementation (`NodeJsTcpSocket`/`NodeJsTcpServer`), React Native support, mock infrastructure
3. **TCP transport core** — `TcpConnection` (4-byte LE length-prefix framing, stream reassembly, receive buffer cap), `TcpTransport` (connection pool, max 3 connections/peer IP, 10s idle timeout)
4. **Server/client runtime wiring** — `NetworkServer` TCP config (`tcp: boolean | {incoming, outgoing}`, `transportPreference`), `NetworkClient` per-peer preference, `ServerNetworkRuntime` port retry loop (UDP IPv4/IPv6 + TCP), DNS-SD T key, multicast interface fallback
5. **Session/exchange/peer integration** — Session-to-connection binding (connection drop → evict sessions, last session → close connection), `MessageChannel` TCP lifecycle, `Peer.newestSession(type?)` with soft/hard transport preference, `PeerConnection` transport constraint, group sessions always UDP
6. **Invoke & Large Message Quality** — `largeMessage` quality → `requiredTransport: TCP`, `CommissioningController`/`MatterController` TCP option forwarding, PASE always UDP
7. **Test suite** — 85+ tests across 8 files (TcpConnection, TcpTransport, MockTcp, TcpSessionBinding, InvokeLargeMessage, NodeJsTcpSocket, NodeJsTcpServer, ClientNodeTcp, DeviceAdvertiser)
8. **Shell, examples, CI** — Shell TCP enabled by default, `nodes tcp`/`nodes descriptor` commands, `--tcp` device flag, `TEST_PREFER_TCP=1` env var, `matterjs-tests-core-tcp` CI job, PICS `MCORE.SC.TCP=1`

### Key design decisions

- **N:1 session-to-connection mapping** matching CHIP SDK pattern
- **Soft transport preference** (TCP preferred → fall back to UDP) vs **hard requirement** (`requiredTransport` for Large Message Quality, no fallback)
- **TCP incoming connections** use ephemeral remote ports — operational address sourced from mDNS, not TCP connection
- **DoS protections**: receive buffer cap (2x maxPayloadSize), max 3 connections per peer IP, 10s idle timeout on new connections, zero-length message rejection
- **Port retry loop**: when auto-assigning ports, retries up to 10 times if IPv4 or TCP port conflicts with IPv6

### Follow-up items (not in this PR)

- Socket abstraction refactoring (StreamSocket/UdpSocket with AsyncIterable)
- `MESSAGE_TOO_LARGE` status report before close
- BDX auto-prefer TCP when peer supports it

Replaces #3465.


packages/general/src/net/Transport.ts
```diff
@@ -12,21 +12,21 @@ import { Channel, ChannelType } from "./Channel.js";
 import { ServerAddress } from "./ServerAddress.js";
 
 /**
- * A local network endpoint associated with a specific {@link ServerAddress} for a connectionless protocol.
+ * A local network endpoint for message transport.
  */
-export interface ConnectionlessTransport {
-    onData(listener: (socket: Channel<Bytes>, data: Bytes) => void): ConnectionlessTransport.Listener;
+export interface Transport {
+    onData(listener: (socket: Channel<Bytes>, data: Bytes) => void): Transport.Listener;
```
lauckhart, line 18: If this interface has "onData" then it's connectionless, right?  What would this method do for TCP?

Or is this the actual communication channel?  In that case, why does the "connection oriented" one have "onConnect"?

This is very confusing, would be good to either clean up or give a better explanation of what a "transport" is and why "onData" and "onConnect" make sense in the same interface

packages/general/src/net/Transport.ts
```diff
@@ -40,19 +40,28 @@ export namespace ConnectionlessTransport {
...
+ * Adds lifecycle hooks for connection/disconnection events.
+ * Methods are optional — a transport may only implement one direction.
  */
-export class ConnectionlessTransportSet<T extends ConnectionlessTransport = ConnectionlessTransport>
-    extends BasicSet<T>
-    implements ConnectionlessTransport.Provider<T>
-{
+export interface ConnectionOrientedTransport extends Transport {
+    /** Fires when a remote peer connects (server role). */
+    onConnect?(listener: (channel: Channel<Bytes>) => void): Transport.Listener;
```
lauckhart, line 49: "Listener" is usually the socket that's listening for connections so this is a little confusing

packages/general/src/net/tcp/TcpSocket.ts
```diff
@@ -0,0 +1,69 @@
...
+
+    /** Address to bind to. undefined = all interfaces. */
+    listeningAddress?: string;
+}
+
+/**
+ * Platform-agnostic TCP client socket abstraction.
+ * Wraps a connected TCP socket with matter.js-compatible event handling.
+ */
+export interface TcpSocket {
```
lauckhart: Would be good to call "TcpConnection" so we aren't redefining "tcp socket" to mean "tcp connection", otherwise things get even more confusing

packages/general/src/net/Network.ts
```diff
@@ -70,6 +71,16 @@ export abstract class Network {
     abstract getIpMac(netInterface: string): MaybePromise<NetworkInterfaceDetails | undefined>;
     abstract createUdpChannel(options: UdpChannelOptions): Promise<UdpChannel>;
 
+    /** Create a TCP server socket. Override in platform implementations that support TCP. */
+    createTcpServer(_options: TcpServerOptions): Promise<TcpServerSocket> {
```
lauckhart: I liked the abstraction of having registered transports rather than hard-coding, but having explicit transport-specific methods maybe makes more given how dependent higher layers are on specifics of the transport.

Buuut having a mix of transport-specific APIs and registered transports is weird.  Should we choose one or the other?

packages/nodejs/src/net/NodeJsNetwork.ts
```diff
@@ -74,6 +81,18 @@ export class NodeJsNetwork extends Network {
         return this.getNetInterfaceZoneIpv6Internal(netInterface, netInterfaceInfo);
     }
 
+    /** Get the first non-internal IPv6 network interface zone ID as fallback for multicast. */
+    static getDefaultNetInterface(): string | undefined {
```
lauckhart, line 89: How's this used?  What does "fallback" mean?

packages/nodejs/src/net/NodeJsNetwork.ts
```diff
@@ -164,6 +183,42 @@ export class NodeJsNetwork extends Network {
...
+                settle(() => resolve(new NodeJsTcpSocket(socket)));
+            });
+
+            const rejectOnce = (error: Error) =>
+                settle(() => {
+                    socket.destroy();
+                    reject(error);
+                });
+
+            socket.setTimeout(TCP_CONNECTION_TIMEOUT_MS);
```
lauckhart: Can we make timeout a parameter?  Presumably we'll want others too, like keepalive

packages/nodejs/src/net/NodeJsTcpServer.ts
```diff
@@ -0,0 +1,107 @@
...
+        throw new NetworkError(`TCP server address is not an IP address: ${addr}`);
+    }
+    return addr.port;
+}
+
+/**
+ * Node.js implementation of the {@link TcpServerSocket} interface.
+ * Wraps a `net.Server` that accepts incoming TCP connections.
+ */
+export class NodeJsTcpServer implements TcpServerSocket {
```
lauckhart: We've got a mix of socket-suffix and no-socket-suffix here...  I'd be inclined to say just do "TcpServer' for the interface, socket doesn't add anything

packages/general/src/net/tcp/TcpConnection.ts
```diff
@@ -0,0 +1,245 @@
...
+/**
+ * TCP channel implementing Matter message framing.
+ *
+ * Each message is prefixed with a 4-byte little-endian uint32 length header indicating the number of message bytes
+ * that follow. Incoming data is buffered and reassembled into complete messages before being emitted.
+ *
+ * The {@link maxPayloadSize} represents the maximum total frame size (header + message). The maximum message content
+ * size is therefore `maxPayloadSize - 4`.
+ */
+export class TcpConnection implements IpNetworkChannel<Bytes> {
```
lauckhart: So here "tcp connection" means "a matter channel over a tcp connection".  Could we reserve "tcp connection" for the layer 4 thing and use TcpChannel here instead?

packages/general/src/net/tcp/TcpConnection.ts
```diff
@@ -0,0 +1,245 @@
...
+    readonly #socket: TcpSocket;
+    readonly #messageListeners = new Set<(data: Bytes) => void>();
+    readonly #closeListeners = new Set<() => void>();
+    readonly #workers = new BasicMultiplex();
+
+    /** Maximum allowed receive buffer size before connection is closed for protection. */
+    readonly #maxReceiveBufferSize: number;
+
+    /** Receive buffer — accumulated chunks not yet consumed. */
+    #receiveChunks = new Array<Uint8Array>();
```
lauckhart: Is there a reason to not use Bytes?

packages/general/src/net/tcp/TcpConnection.ts
```diff
@@ -0,0 +1,245 @@
...
+        return { type: "tcp", ip: this.#socket.remoteAddress, port: this.#socket.remotePort };
+    }
+
+    /** Maximum message content size (total frame size minus the 4-byte length header). */
+    get maxMessageSize(): number {
+        return this.maxPayloadSize - FRAMING_HEADER_SIZE;
+    }
+
+    /**
+     * Send a framed message — prepends a 4-byte little-endian length header.
```
lauckhart: Do we want these comments to spell out details of message framing?

packages/general/src/net/tcp/TcpConnection.ts
```diff
@@ -0,0 +1,245 @@
...
+            for (const listener of this.#messageListeners) {
+                listener(message);
+            }
+        }
+    }
+
+    /**
+     * Collapse all buffered chunks into a single contiguous Uint8Array.
+     */
+    #flatten(): Uint8Array {
```
lauckhart: Don't we have a Bytes utility for this?  If not should add one

packages/general/src/net/tcp/TcpConnection.ts
```diff
@@ -0,0 +1,245 @@
...
+                this.#workers.add(this.close());
+                return;
+            }
+
+            const totalNeeded = FRAMING_HEADER_SIZE + messageLength;
+            if (this.#receiveLength < totalNeeded) {
+                return;
+            }
+
+            const message = flat.slice(FRAMING_HEADER_SIZE, totalNeeded);
```
lauckhart: Flattening then slicing is inefficient.  Would be better to create message buffer, fill it, then slice just the remaining chunk

packages/general/src/net/tcp/TcpTransport.ts
```diff
@@ -0,0 +1,213 @@
...
+    /** Port to listen on (server role). undefined = no server. */
+    listeningPort?: number;
+    /** Address to bind to. */
+    listeningAddress?: string;
+    /** Max TCP frame size (header + message). Default DEFAULT_MAX_TCP_MESSAGE_SIZE. */
+    maxMessageSize?: number;
+}
+
+/**
+ * TCP transport layer combining server and client roles with a connection pool.
```
lauckhart: I think I see how we're making it work...  "onData" gives you "channel" + "data".  Which...  why wouldn't you just register from the channel if you want data from it?  Are we doing it this way for a reason?  Or just because we're shoehorning into a broken abstraction?

Also, how does a pool make sense given that we know peers have sessions tied to specific connections and thus connections are not interchangeable?

packages/general/src/net/tcp/TcpTransport.ts
```diff
@@ -0,0 +1,213 @@
...
+            await connection.close();
+        }
+    }
+
+    #registerConnection(connection: TcpConnection): void {
+        const { ip, port } = connection.networkAddress;
+        const key = `${ip}:${port}`;
+
+        // Close any existing connection with the same key to prevent orphaning (M-1 fix)
+        const existing = this.#connections.get(key);
```
lauckhart: This seems like it's the opposite of a pool, it's more of a registry of singular connections to different resources (that are definitely not interchangeable)

packages/node/src/behavior/system/network/NetworkClient.ts
```diff
@@ -261,6 +276,12 @@ export namespace NetworkClient {
          * The highest event number seen from this node for the default read/subscription.
          */
         maxEventNumber = EventNumber(0);
+
+        /**
+         * Per-peer transport preference override. If set by the user, persisted.
```
lauckhart: No need to mention persisting, all config values do that.  Or we should mention it eveywhere

packages/node/src/behavior/system/network/NetworkServer.ts
```diff
@@ -144,10 +141,39 @@ export namespace NetworkServer {
...
+         * Fine-grained control via `{ incoming?: boolean; outgoing?: boolean }`.
+         * Disabled by default.
+         */
+        tcp?: boolean | { incoming?: boolean; outgoing?: boolean } = undefined;
+
+        /**
+         * Preferred transport for outgoing connections. Default "udp".
+         * Revisit once TCP device stability is proven.
+         */
+        transportPreference?: "tcp" | "udp" = "udp";
```
lauckhart: Should make optional or set default, not both, I think?  And be consistent with.tcp

packages/node/src/behavior/system/network/NetworkServer.ts
```diff
@@ -144,10 +141,39 @@ export namespace NetworkServer {
         };
         subscriptionOptions?: ServerSubscriptionConfig = undefined;
 
+        /**
+         * Enable TCP transport. `true` enables both incoming (server) and outgoing (client).
+         * Fine-grained control via `{ incoming?: boolean; outgoing?: boolean }`.
+         * Disabled by default.
+         */
+        tcp?: boolean | { incoming?: boolean; outgoing?: boolean } = undefined;
```
lauckhart: We should have @field annotations on these so they're type checked and configurable

...which makes the type union problematic since matter semantics don't have that...  I do think it'd be interesting to add a "union" type.  Is on my list.  I can look into that if you want

packages/protocol/src/protocol/DeviceAdvertiser.ts
```diff
@@ -21,6 +22,8 @@ const logger = Logger.get("DeviceAdvertiser");
 export interface DeviceAdvertiserContext {
     fabrics: FabricManager;
     sessions: SessionManager;
+    /** TCP transport support bitmap for operational DNS-SD advertisement. */
+    tcp?: SupportedTransportsBitmap;
```
lauckhart: Maybe call supportedTransports in case they ever put something else in it?

packages/protocol/src/peer/Peer.ts
```diff
@@ -120,12 +128,23 @@ export class Peer {
...
             });
 
-            // Ensure the operational address is always set to the most recent IP
             if (!session.isClosed) {
                 const { channel } = session.channel;
                 if (isIpNetworkChannel(channel)) {
-                    updateNetworkAddress(channel.networkAddress);
-                    channel.networkAddressChanged.on(updateNetworkAddress);
+                    if (channel.type === ChannelType.TCP) {
+                        // For incoming TCP the remote port is ephemeral — use the mDNS port
```
lauckhart, line 129: MDNS port seems like a very weird choice for a guard value.  Maybe `const UNKNOWN_PORT = -1`?  (or 0 would work too)

packages/protocol/src/peer/Peer.ts
```diff
@@ -258,7 +277,7 @@ export class Peer {
         }
 
         while (true) {
-            const session = this.newestSession;
+            const session = this.newestSession();
```
lauckhart: Should this be respecting the transport constraint?

packages/protocol/src/peer/PeerConnection.ts
```diff
@@ -566,6 +582,13 @@ export namespace PeerConnection {
         network?: string;
         kicker?: Observable<[KickOrigin]>;
 
+        /**
+         * Constrain the transport type for this connection. When set to `ChannelType.TCP`, the
+         * connection will be established over TCP (requires peer TCP Server support and local
+         * TCP outgoing enabled). When undefined, the default transport (UDP/MRP) is used.
+         */
+        transportConstraint?: ChannelType;
```
lauckhart: I think "constraint" is unnecessary in this name.

packages/node/src/behavior/cluster/ClusterBehaviorType.ts
```diff
@@ -34,79 +35,58 @@ import { introspectionInstanceOf } from "./cluster-behavior-utils.js";
  * This factory performs runtime class generation of behaviors associated with a Matter cluster.  It implements
  * ClusterBehavior.for() directly and is a core component of PeerBehavior().
  */
-export function ClusterBehaviorType({
-    namespace,
+export function ClusterBehaviorType<const C extends ClusterType>({
```
lauckhart: What's going on here?  Can we do a separate PR for this?  Or does this branch need to be rebased or something?

packages/protocol/src/peer/ControllerCommissioningFlow.ts
```diff
@@ -874,7 +875,7 @@ export class ControllerCommissioningFlow {
             const regulatoryResult = await this.#invokeCommand(
                 {
                     endpoint: RootEndpointNumber,
-                    cluster: GeneralCommissioning,
+                    cluster: GeneralCommissioning.Complete,
```
lauckhart: This is definitely reverting changes, something is screwed up with this commit

packages/node/src/behavior/cluster/ClusterBehaviorType.ts
```diff
@@ -35,58 +35,79 @@ import { introspectionInstanceOf } from "./cluster-behavior-utils.js";
  * This factory performs runtime class generation of behaviors associated with a Matter cluster.  It implements
  * ClusterBehavior.for() directly and is a core component of PeerBehavior().
  */
-export function ClusterBehaviorType<const C extends ClusterType>({
-    cluster,
+export function ClusterBehaviorType({
```
lauckhart, line 38: More weirdness.  I think claude might have reverted and recreated some of the changes on main

PR #3473, opened 2026-03-27
Skip deprecated/disallowed elements in protocol service

Apollon77:
## Summary

- Skip deprecated and disallowed cluster members (e.g. `EventList` with conformance `"D"`) in `clusterTypeProtocolOf` before attempting TLV generation
- Fixes crash `"No metabase for model EventList"` that occurs on the client side when a remote device reports `EventList` (0xFFFA) in its `attributeList`
- Deprecated elements have no type definition, so `TlvOfModel` cannot generate a schema for them — they are now treated like unknown elements, invisible to the protocol layer

## Root Cause

`EventList` is a global attribute injected into every cluster via `injectGlobalAttributes`. It has `conformance: "D"` (deprecated) and **no type definition**. On the server side this was harmless because `supportedElements` never includes it. On the client side, when a remote device reports `EventList` in its attribute list, the client adds it to `supportedElements`, causing `clusterTypeProtocolOf` to call `TlvOfModel` on the typeless model — which throws.

## Changes

- **`ProtocolService.ts`**: Added early `continue` for `member.isDeprecated || member.isDisallowed` in the member iteration loop
- **`ProtocolServiceTest.ts`**: Added test verifying deprecated `EventList` is excluded from protocol attributes and confirming `TlvOfModel` throws on the deprecated model (proving the guard is necessary)

## Test plan

- [x] New test: `excludes deprecated global attributes like EventList from protocol`
- [x] Verified test passes with the fix
- [x] Verified `TlvOfModel` throws without the guard (test asserts this directly)
- [x] All existing ProtocolServiceTest tests continue to pass (10/10)

### DIFF
--- a/packages/nodejs-shell/src/MatterNode.ts
+++ b/packages/nodejs-shell/src/MatterNode.ts
@@ -138,9 +138,12 @@ export class MatterNode {
     }
 
     async close() {
-        await this.commissioningController?.close();
-        this.#observers?.close();
-        await this.#storageManager?.close();
+        try {
+            await this.commissioningController?.close();
+        } finally {
+            this.#observers?.close();
+            await this.#storageManager?.close();
+        }
     }
 
     async start() {
--- a/packages/nodejs-shell/src/shell/Shell.ts
+++ b/packages/nodejs-shell/src/shell/Shell.ts
@@ -107,6 +107,20 @@ export class Shell {
                         process.exit(1);
                     });
             })
+            .on("SIGINT", () => {
+                // Readline in terminal mode intercepts CTRL-C as a keypress and does not
+                // propagate it as a process signal, so we need to handle it explicitly.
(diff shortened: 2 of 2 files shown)


packages/node/test/endpoint/server/ProtocolServiceTest.ts
```diff
@@ -332,6 +332,38 @@ describe("ProtocolServiceTest", () => {
         expect(sent?.status?.status.status).deep.equals(Status.Success);
     });
 
+    it("excludes deprecated global attributes like EventList from protocol", async () => {
+        const node = await MockServerNode.createOnline(undefined, {
+            device: OnOffLightDevice.with(OnOffServer),
+        });
+
+        const EVENT_LIST_ID = 0xfffa;
```
lauckhart: Copilot is correct here, use EventList.id

PR #3474, opened 2026-03-27
Upgrade to Matter specification 1.5.1

### LINKED ISSUE #40: Matter camera enabler - Matter JS 1.5.(1)

Apollon77:
### Spec 1.5.0 parser and model
- Upgrade spec parser and codegen tooling for Matter 1.5 HTML structure changes
- Add parsed Matter 1.5 specification and model overrides
- Fix codegen for composite device types with no direct cluster requirements (FloodlightCamera, VideoDoorbell)
- Fix spec parser for renamed "Derived Cluster Namespace" section in mode-derived clusters
- Fix `[object Object]` rendering in validation error messages
- Add missing generated file patterns to `.prettierignore`

### Spec 1.5.1 upgrade
- Upgrade from 1.5.0 to 1.5.1 spec, regenerate all cluster/endpoint/forward files
- Add `Rev >= vN` revision conformance support (parser, AST, compiler, variance)
- Add `L` (Large Message) quality support
- Update spec references from v1.4.2 to v1.5.1 in all generated JSDoc
- Add MatterSpecification namespaces for v1.5 and v1.5.1

### Runtime 1.5.1 compliance
- **AccessControl**: Set capability attribute defaults in initialize() for tightened constraints
- **LevelControl**: Set minLevel/maxLevel defaults for Rev >= v7 mandatory conformance
- **DoorLock**: Remove legacy Zigbee commands (SetPINCode, GetPINCode, etc.) and vestigial supervision
- **UserLabel**: Add RESOURCE_EXHAUSTED validation for LabelList writes (configurable max via State.maxLabels)
- **RevokeCommissioning**: Run PASE session cleanup before window check per spec 1.5.1 ordering
- **OpCreds**: Invoke RevokeCommissioning before and after RemoveFabric (SHOULD per spec)
- **WildcardPathFlags**: Reject in write interactions with InvalidAction
- **PairingHintBitmap**: Add bits 20-22 (1.5.0/1.5.1 additions)
- **Remove REQUIRED_CAT_MISMATCH** secure channel status code (removed from spec)
- **RVC example**: Support DirectModeChange feature bypass in CustomRvcRunModeServer

### Override cleanup
- Mark ColorControl ColorCapabilities bitmap override as `until: "1.4.2"` (spec 1.5 adds ColorCapabilitiesBitmap)
- Mark DoorLock GetYearDayScheduleResponse field ID override as `until: "1.4.2"` (spec 1.5 has correct IDs)

Replaces #3445 — rebased on current main.

## Breaking changes

- Some Namespaces were renamed and now have a "Common*" prefix
- Several previous "Zigbee only" features, attributes and commands were removed because they were never allowed for Matter
- DoorLock: Legacy non-USR commands removed (SetPINCode, GetPINCode, ClearPINCode, SetRFIDCode, etc.)
- OperationalCredentials: `supportedFabrics` default changed from 254 to 5 (spec minimum). Set explicitly in ServerNode config if you need more.
- Spec JSDoc references updated from v142 to v151


packages/model/src/models/RequirementModel.ts
```diff
@@ -91,6 +93,7 @@ export class RequirementModel extends Model<RequirementElement, RequirementModel
     override toElement(omitResources = false, extra?: Record<string, unknown>) {
         return super.toElement(omitResources, {
             element: this.element,
+            instance: this.instance,
```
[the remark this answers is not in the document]
lauckhart: I completely disagree with copilot here.  Having instance present unconditionally will keep the number of object shapes down and make it less likely that v8 will deeoptimize consuming functions to polymorphic or megamorphic.  And if we're using "in" we deserve what we get.  And damn that's an ugly way to set a property

support/models/src/local/OccupancySensingOverrides.ts
```diff
@@ -0,0 +1,247 @@
+/**
+ * @license
+ * Copyright 2022-2026 Matter.js Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/**
+ * The OccupancySensing cluster (0x406) is defined in Matter 1.5 spec section 2.7 but was not parsed by
```
lauckhart: Why do this instead of fix the parser?  This drops all docs and xrefs and is going to tend toward bitrot.

.prettierignore
```diff
@@ -10,7 +10,9 @@ packages/types/src/globals/*
 packages/node/src/tags/*
 packages/node/src/endpoints/*
 packages/node/src/devices/*
+packages/node/src/behaviors/index.ts
 packages/node/src/behaviors/*/index.ts
 packages/node/src/behaviors/*/*Behavior.ts
+packages/node/src/behaviors/*/*Server.ts
```
lauckhart: This wasn't in prettierignore because there's a huge corpus of hand-written servers

support/codegen/src/util/TsFile.ts
```diff
@@ -130,7 +130,7 @@ export abstract class Entry {
             }
 
             return `${linePrefix}/**\n${lines
-                .map(l => `${linePrefix} * ${l}`.trimEnd())
+                .map(l => `${linePrefix} * ${l.replace(/\*\//g, "* /")}`.trimEnd())
```
lauckhart, line 133: Oh man, where did this happen?

support/codegen/src/mom/spec/translate-cluster.ts
```diff
@@ -45,6 +45,29 @@ export function* translateCluster(definition: ClusterReference) {
 
     const metadata = translateMetadata(definition, children);
     if (!metadata) {
+        if (definition.ids) {
+            // Section has a cluster ID table but IDs couldn't be parsed — this is a real cluster that
+            // needs manual handling via overrides (e.g. OccupancySensing).  Skip entirely.
```
lauckhart: No, this is a real cluster that needs the parser fixed.  Reverting to manual cluster definition isn't OK

lauckhart (approved):
I think it's good to go.  Only big issue is with occupancy sensing (claude decided to just bail because the ID was apparently hard to parse).

And just noting we need to make sure #3471 reversion gets fixed

packages/node/src/behaviors/door-lock/NonUserDoorLockServer.ts
lauckhart: Does it still make sense to have user/non-user versions of the behavior?  The reason we had them was because there were a bunch of methods that were disabled if !USR.  If they've been removed then this should just becomena single server implementation like other servers

lauckhart (approved):

packages/node/src/behaviors/door-lock/NonUserDoorLockServer.ts
Apollon77: yes I will merge them

PR #3586, opened 2026-04-13
Restore Cluster.with() as typing-only compat shim

Apollon77:
## Summary

- Restores the pre-PR #3466 `PowerSource.Cluster.with(Feature.X, Feature.Y)` call site that external users still rely on.
- New reusable helper type `ClusterType.WithCompat<NS, T>` layers a `.with()` method onto the deprecated `Cluster` const. Codegen emits a single-line declaration per feature-bearing cluster.
- Runtime install in `ClusterType.ts` returns an `Object.create(ns)` clone with a `supportedFeatures` marker. Feature selection has no further runtime effect — attribute maps are not culled by conformance.
- Deprecated from day one. Tracked for removal in 0.18.

## Design notes

- **Why "born @deprecated".** Compat shim, not a new API. The `@deprecated` JSDoc is intentional — restores a removed call site while signalling the path forward. TS surfaces `[6387]` deprecation warnings on every call.
- **Why `.with()` despite the name collision with `ClusterBehavior.with()`.** The two operations have different semantics: the behavior-layer `.with()` performs real schema-level feature derivation via `applyFeatureSelection`; this namespace shim is typing-only. Keeping the pre-#3466 name is the whole point — renaming would defeat the shim.
- **Runtime compatibility with `ClusterBehaviorType.syncFeatures`.** `syncFeatures` reads `namespace.supportedFeatures` and feeds it to `new FeatureSet(...)`. `FeatureSet.normalize` matches feature names case-insensitively, so the shim's camelCase keys resolve correctly against schema feature short/long names.
- **Scope intentionally limited.** Only `.with()` is restored. `.alter() / .enable() / .set()` from the old `MutableCluster` are not re-added — no demand surfaced. `.with()` is attached only to the `Cluster` const (already `@deprecated`), not to the top-level namespace.

## Test plan

- [x] `npm run build` full repo — clean
- [x] `npm test -w @matter/types` — 297/297 pass, including new `ClusterTypeWithCompatTest` suite covering identity metadata, attribute pass-through, `supportedFeatures` population, empty selection, no source mutation, prototype-chain fallthrough, and fresh clone per call
- [x] `npm run format-verify` — clean
- [x] `npm test` full suite — no failures
- [x] TS `[6387]` deprecation diagnostic fires on `PowerSource.Cluster.with(...)` call sites — confirms IDE surfacing via the helper type
- [ ] Verified by the external user who reported the regression

### DIFF
--- a/packages/types/src/cluster/ClusterType.ts
+++ b/packages/types/src/cluster/ClusterType.ts
@@ -163,6 +163,21 @@ function installLazyProperties(ns: object, model: ClusterModel) {
             }
             return Object.freeze(result);
         });
+
+        // Compat shim for pre-PR #3466 call sites: `PowerSource.Cluster.with(Feature.X, Feature.Y)`.
+        // Returns a shallow prototype-based clone of the namespace with `supportedFeatures` set.  Feature
+        // selection has no other runtime effect (attribute maps are not culled by conformance).
+        // @deprecated Removal tracked for 0.18.
+        lazy("with", () => (...features: string[]) => {
+            const clone = Object.create(ns) as Record<string, unknown>;
+            const supportedFeatures: Record<string, true> = {};
+            for (const feature of features) {
+                // Feature enum values are PascalCase; SupportedFeatures keys are camelCase.
+                supportedFeatures[feature.charAt(0).toLowerCase() + feature.slice(1)] = true;
+            }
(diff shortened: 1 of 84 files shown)


Apollon77:
@Luligu Please have a look if this is sufficient for your needs

PR #3630, opened 2026-04-20
fix: remove interactionComplete observer on Datasource close

gtalusan:
DatasourceImpl registered an observer on session.interactionComplete each time a write occurred inside a remote-actor interaction, but close() never removed it. A datasource can participate in multiple concurrent interactions (different sessions writing the same resource), and the session's interactionComplete observer list kept a closure on the DatasourceImpl alive until each exchange closed.

Changes:
- Add observedInteractions: Set<AsyncObservable> to DatasourceImpl, tracking every interactionComplete observable this datasource has subscribed to.
- In RootReference.#startWrite(), add the session's interactionComplete to the set before calling .on().
- In close(), iterate the set and call .off() on each, then clear the set.
- In interactionObserver, remove ourselves from the firing observable, delete it from the set, and reset session.interactionStarted so a reused session can re-emit interactionBegin and re-register the observer.

Tests added to DatasourceTest.ts:
- Observer is registered after a write with a remote session
- Observer is removed from interactionComplete after close()
- interactionEnd event fires and observer deregisters when interactionComplete emits with the session
- Observer self-removes from interactionComplete when it fires naturally
- Observer re-registers when the same session is reused for a subsequent interaction


gtalusan:
@copilot apply changes based on the comments in [this thread](https://github.com/matter-js/matter.js/pull/3630#pullrequestreview-4148301013)

PR #3636, opened 2026-04-21
fix(mdns): fast-reject queries with no matching registered record

gtalusan:
On a busy LAN, `MdnsServer.#handleMessage` encodes and sends responses to every mDNS query, even for service types with no registered records.

Lazily populate a `#knownQueryNames` Set from the records cache and fast-reject queries where no name matches, skipping the encode/send path. Cleared in `setRecordsGenerator` when records change.

### DIFF
--- a/packages/protocol/src/mdns/MdnsServer.ts
+++ b/packages/protocol/src/mdns/MdnsServer.ts
@@ -63,6 +63,8 @@ export class MdnsServer {
             timestamp: Timestamp;
         }
     >();
+    /** Set of DNS record names we have registered; used to fast-reject non-matching queries. */
+    readonly #knownQueryNames = new Set<string>();
 
     readonly #socket: MdnsSocket;
 
@@ -98,6 +100,21 @@ export class MdnsServer {
         }
 
         const { sourceIntf, sourceIp, transactionId, queries, answers: knownAnswers } = message;
+
+        // Lazily build the set of record names we can answer from the first populated cache result.
+        if (this.#knownQueryNames.size === 0) {
+            for (const portRecords of records.values()) {
+                for (const record of portRecords) {
+                    this.#knownQueryNames.add(record.name);
+                }
+            }
+        }
+
+        // Fast-reject messages where no query could possibly match our records, avoiding
+        // the per-record iteration and encode/send work for irrelevant LAN traffic.
+        if (this.#knownQueryNames.size > 0 && !queries.some(q => this.#knownQueryNames.has(q.name))) {
+            return;
+        }
 
         for (const portRecords of records.values()) {
             let answers = queries.flatMap(query => this.#queryRecords(query, portRecords));
@@ -234,6 +251,7 @@ export class MdnsServer {
         await this.#records.clear();
         this.#recordLastSentAsMulticastAnswer.clear();
         this.#recentlyAnsweredQueries.clear();
+        this.#knownQueryNames.clear();
         this.#recordsGenerator.set(service, generator);
     }
 


Apollon77:
Do you have more details and example or logs on the issue? So especially examples about cases where we responded to "not owned" record names? Especially what was responded because we should not have any data.

I want to understand the problem first please

gtalusan:
Addressed both reviewer comments plus two additional race conditions found during review:

**1. Per-interface isolation** (original comment): Changed `#knownQueryNames` from `Set<string>` to `Map<string, Set<string>>` keyed by `sourceIntf`, so per-interface record names (e.g. hostname from interface MAC) don't cross-contaminate fast-reject decisions on other interfaces.

**2. Clear on expire/close** (original comment): Added `.clear()` calls in `#resetServices()` (called by `expireAnnouncements`) and `close()`.

**3. Stale-record race across the `await this.#records.get()` boundary**: Added `#knownQueryNamesVersion`, incremented as the *first synchronous operation* in each reset/close path. The version is captured before the await and re-checked immediately after — a changed version causes the handler to return without using stale records. The `#knownQueryNames.set()` write-back is also guarded by the same check.

**4. Stale records during the inter-packet send loop**: The `for (const portRecords...)` loop has an `await Time.sleep()` at the end of each iteration; added a version re-check at the top of each iteration so a concurrent reset aborts remaining sends.

**5. New generator installed before the async `clear()`**: Moved `#recordsGenerator.set(service, generator)` to before `await this.#records.clear()` in `setRecordsGenerator`, so handlers that start after the version bump always regenerate using the new generator rather than the old one.

Note: two pre-existing issues were identified during review but are out of scope for this PR: (a) `buildDnsRecordKey` ignores record RDATA so distinct records with the same name/type/class hash identically in duplicate-suppression tracking; (b) there is a narrow window after a version bump where `AsyncCache` may still return a not-yet-cleared stale entry to a newly-starting handler — inherent to the async cache design and warrants a separate fix.

gtalusan:
@Apollon77  yeah, sorry, ouch, this turned into something larger than I initially intended while trying to track down high CPU usage with homebridge beta-2.0.0....

I've been CPU profiling my child bridge and the multiple mDNS stacks are causing higher than expected CPU usage and warming up my network rack.

I had GPT-5.4 review and it pointed out more bugs than my initial stab at the problem.  It might be getting out of hand trying to point out bugs outside of the initial stab at this PR.

Apollon77:
Hehe. yeah I was optimizing the mdns stack already a lot. Homebridge which also adds an own mdns sender could make things interesting and sure go beyond what I analyzed so far. 

But I also know that just looking at proposed fixes in such situations without knowing and understanding the root issue and how it manifests is very hard so that's why more information on the "issue" would help a lot. 

But I can also look myself (and likely Claude) tomorrow into what you described. Let's see what I can find. But any additional information will help me to understand it. 

And yes AI sometimes go wild with thinks they (might) find :-)

gtalusan:
@Apollon77 LOL yeah, so I have a feeling the Sonnet/GPT-5.4 might be treating the symptoms rather than the root cause with this PR.

The CPU profiling analysis showed this code was one of the hottest spots, but I'm wondering if it has to do with homebridge/matter.js setting up UDP listeners on all of the interfaces of my host-network docker container.  I'll explore that avenue instead of this noise.

Thanks!

Apollon77:
Yeahh multiple interfaces is much especially when they stat to play ping pong. Do me a favor and just create an issue with the information you had about cause. I will have a look

gtalusan:
#3637 - thanks!

PR #3638, opened 2026-04-21
fix(mdns): fast-reject unrelated LAN queries + case-insensitive matching

### LINKED ISSUE #3637: MdnsServer loops inefficiently when looking for matching queries
`MdnsServer.#handleMessage` encodes and sends a response for every mDNS query it receives, including queries for Chromecast, Apple TV, printers, etc. matter.js should only respond to queries that match its own registered record names (`_matter._tcp.local`, `_matterc._udp.local`, etc.).

CPU profiling shows a hotspot #handleMessage — on a busy LAN this could compound with every new device.

Apollon77:
## Summary

Fixes #3637. Addresses the MdnsServer CPU hotspot reported for busy LANs where every Chromecast / AppleTV / printer mDNS query was entering the full truncation-handling + per-service record-scan path even though the downstream guards correctly suppressed any response.

The first check already proved the PR's stated premise wrong — we do NOT send responses to unrelated queries. But the CPU work per packet is real, and cleaning it up exposed a few related issues worth addressing in the same branch.

- **Fast-reject** by lowercased record-name set stored alongside the AsyncCache'd records (no separate invalidation bookkeeping — piggybacks on the existing cache lifecycle). Zero-query packets still fall through so RFC 6762 §7.2 TC continuations can merge their known-answer records into cached fragments.
- **Case-insensitive name matching** per RFC 6762 §16 / RFC 1035 §2.3.3 in `#queryRecords`, extended to `buildDnsRecordKey` and `#shouldSuppressResponse` so duplicate-suppression and multicast TTL tracking agree with the invariant.
- **Removed trailing inter-packet sleep** after the final send in `#handleMessage`, `broadcast`, and `expireAnnouncements`. The 20-120ms pacing between successive packets is kept; the trailing sleep after the last one was pure latency waste.
- **Install generator synchronously before awaiting cache clear** in `setRecordsGenerator` so cache refills triggered during the clear use the new generator rather than re-caching stale records with the old one.
- 6 new tests covering unknown-name rejection, A-record for host qname (differs from SRV qname), mixed known/unknown queries, case-insensitive match, TC continuation known-answer merging, and generator replacement.

### DIFF
--- a/packages/protocol/src/mdns/MdnsServer.ts
+++ b/packages/protocol/src/mdns/MdnsServer.ts
@@ -276,10 +276,14 @@ export class MdnsServer {
     }
 
     #queryRecords({ name, recordType }: { name: string; recordType: DnsRecordType }, records: DnsRecord<any>[]) {
+        // DNS names are case-insensitive per RFC 6762 §16 / RFC 1035 §2.3.3.
+        const queryName = name.toLowerCase();
         if (recordType === DnsRecordType.ANY) {
-            return records.filter(record => record.name === name);
+            return records.filter(record => record.name.toLowerCase() === queryName);
         } else {
-            return records.filter(record => record.name === name && record.recordType === recordType);
+            return records.filter(
+                record => record.name.toLowerCase() === queryName && record.recordType === recordType,
+            );
         }
     }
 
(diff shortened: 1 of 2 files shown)


gtalusan:
👍

PR #3641, opened 2026-04-22
Avoid duplicate UDP multicast sends

gtalusan:
## Problem

`UdpMulticastServer.send()` iterates every IPv4 and IPv6 address on an interface even though broadcast channels are keyed only by `(netInterface, isIPv4)`, so interfaces with multiple addresses in the same family send the same multicast packet more than once. This change deduplicates by address family before looking up the broadcast channel, so each interface sends at most once for IPv4 and once for IPv6.


Apollon77:
Thanks! I will do dome dlight adjustments and merge it next. Great work!

Apollon77 (approved):

PR #3652, opened 2026-04-24
fix(node): strip feature-gated events from ClusterEvents under .for() narrowing

Apollon77:
## Summary
- Fix `ClusterEvents<N, BaseT>` Omit leaking feature-gated events when a wider-featured base is narrowed via `.for(Cluster)` (e.g. `SwitchBaseServer.for(Switch)`), so `endpoint.eventsOf(SwitchServer).initialPress` no longer type-checks on a bare `SwitchServer`.
- Switch the Omit key from `keyof Properties<N>` to `keyof CompleteProperties<N>` — stable across narrowing chains while `Properties<N>` still adds back only feature-applicable observables.
- Add a 22-case type-level regression harness under `packages/node/test/behavior/cluster/ClusterNarrowingTypingTest.ts` covering `stateOf` / `eventsOf` / `commandsOf` / `.alter()` / `endpoint.set()` / constructor `InputStateOf` / subclass `this.state` & `this.events` narrowing.

## Root cause

Introduced by commit a295568c6 ("Bug 14 fix") under the premise that *old* and *new* `Properties<N>` always share key sets. That's false when features narrow: `Properties<Switch-with-all-features>` contains `initialPress`/`switchLatched`/etc., while `Properties<Switch-with-no-features>` does not — so the Omit against the new key set failed to strip feature-gated events inherited from the wider base. The comment at the top of `ClusterEvents.ts` even documented the faulty assumption.

`CompleteProperties<N>` is derived from *all* components regardless of feature flags, so its `keyof` is constant for a given cluster — stripping correctly for same-cluster narrowing while leaving subclass-contributed observables (e.g. `SwitchServer.State.rawPosition$Changing/Changed`) intact.

## Known limitation

`CompleteProperties<N>` only covers keys from the *current* cluster's components. Cross-cluster `.for(OtherNamespace)` chains (different base cluster, not just narrower features) retain the pre-existing behavior where old-cluster-specific event names would not be stripped. None of matter.js's in-tree behaviors hit this; same-cluster narrowing — the matter-bridge case and every in-repo usage — is fully fixed.

### DIFF
--- a/packages/node/src/behavior/cluster/ClusterEvents.ts
+++ b/packages/node/src/behavior/cluster/ClusterEvents.ts
@@ -14,15 +14,16 @@ import type { ActionContext } from "../context/ActionContext.js";
 /**
  * Event instance type for ClusterBehaviors.
  *
- * The Omit key uses `keyof Properties<N>` rather than `keyof Properties<InterfaceOf<BaseT>>` (which would
- * mirror main's pattern of stripping OLD cluster events).  TypeScript can't resolve the latter through nested
- * ClusterEvents types.  Using N is equivalent because old and new are always the same cluster (same
- * attribute/event keys), just different feature selections.
+ * Omit key uses `keyof CompleteProperties<N>` — the full set of event-shaped keys the cluster can ever expose
+ * regardless of feature selection.  This keeps the Omit stable across narrowing chains (e.g.
+ * `Base.with(all features).for(Cluster)` → base's Events carries feature-gated keys that must be stripped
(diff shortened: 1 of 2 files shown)


Apollon77:
@luxni please verify

Luxni:
> @Luxni please verify

Copy that. I'll test it as soon as possible.

Apollon77:
lol @Luxni i did mention you wrongly ...sorry ... no need for you ...

@Luligu :-)

Luligu:
Thanks!

PR #3671, opened 2026-04-28
fix(general): decode TXT records as raw bytes per RFC 6763 §6

Apollon77:
## Summary

- **Codec layer (breaking):** `DnsCodec.decodeTxtRecord` now returns `Bytes[]` instead of UTF-8-decoded `string[]`, so binary TXT values (Thread MeshCoP `xa`, `xp`, `at`, `pt`, `sb`, `dd`) survive encode/decode losslessly per RFC 6763 §6. `encodeTxtRecord` and the `TxtRecord` factory accept `(Bytes | string)[]` for ergonomic ASCII senders.
- **Façade (additive):** `DnssdName.parameters` and `IpService.parameters` now return a `DnssdParameters` instance — implements `ReadonlyMap<string, string>` (zero-touch for existing consumers like `DiscoveryData`, `CommissionableMdnsScanner`, `Peer`) plus a new `.raw(key): Bytes` accessor for binary TXT consumers (driver: matterjs-server 406 Thread Border Router enrichment).
- **RFC 6763 conformance fixes:** `DnssdName.parameters` now first-wins on duplicate keys (§6.4), silently ignores zero-length entries (§6.5), and ignores entries with empty keys (§6.4).
- **Pre-existing bug:** `Bytes.of` had swapped `Uint8Array` constructor args (`(buffer, byteLength, byteOffset)` instead of `(buffer, byteOffset, length)`). Latent because every existing call site passed a `Uint8Array` and hit the early-return branch — fired on `DataView`/`Int8Array`/etc.

## Why

`DnsCodec.decodeTxtRecord` previously decoded each TXT entry via `readUtf8String`, which silently replaced any byte ≥ 0x80 with U+FFFD. That made arbitrary-bytes TXT values (per RFC 6763 §6: "The value can contain any eight-bit values including '='.") unrecoverable from the matter.js public API. The Thread Border Router `_meshcop._udp` and `_trel._udp` services advertise binary fields like `xa = 0x5AAF359C0501A1B0` that mojibaked to garbage; consumers couldn't match BRs to neighbor tables.

Internal Matter consumers see no observable change — Matter spec keys (D, CM, SII, SAI, SAT, T, ICD, VP, DN, …) are all ASCII; the `Map<string, string>`-shaped façade preserves the existing contract.

## Test plan

- [x] New round-trip test in `DnsCodecTest.ts` proves a TXT entry containing `0x5AAF359C0501A1B0` survives encode → decode bytewise (`Bytes.areEqual`).
- [x] New `DnssdParameters` unit tests (13) cover the full `ReadonlyMap` contract plus `.raw()` lossless retrieval, missing-key handling, iteration, `forEach` with `thisArg`.
- [x] New end-to-end test in `DnssdNamesTest.ts` installs a TXT record with both ASCII (`SII=5000`) and binary (`xa=…high-byte-bytes…`) entries; asserts `.raw("xa")` is bytewise equal to the original and `.get("SII") === "5000"`.
- [x] RFC 6763 §6.4 / §6.5 conformance tests: first-wins on duplicate keys, ignore empty-key entries, ignore zero-length entries, preserve `key=` empty-value entries.
- [x] `Bytes.of` regression test: a `DataView` with non-zero offset round-trips correctly.
- [x] Full repo test suite green.
- [x] `npm run format-verify` clean.
- [x] `npm run lint` clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,7 +22,9 @@ The main work (all changes without a GitHub username in brackets in the below li
 
 - @matter/general
     - Breaking: Blob/File-related storage methods were removed from the normal Storage implementation
+    - Breaking: `DnsCodec.decodeTxtRecord`/`encodeTxtRecord` and the `TxtRecord` factory now operate on `Bytes[]` per RFC 6763 §6 so binary TXT values (e.g. Thread MeshCoP `xa`, `xp`, `at`, `pt`, `sb`, `dd`) survive encode/decode losslessly; `encodeTxtRecord` and `TxtRecord` accept `(Bytes | string)[]` so ASCII-only senders are unchanged
     - Feature: Added a new `wal`-based storage engine (not yet the default) to optimize persistence
+    - Feature: `DnssdName.parameters` and `IpService.parameters` now return a `DnssdParameters` instance — a `ReadonlyMap<string, string>` plus a `.raw(key): Bytes` accessor for binary TXT consumers (e.g. Thread Border Router enrichment)
     - Enhancement: Added locking to storage implementations to prevent concurrent access issues and data corruption
     - Enhancement: Split out Blob-Storage into its own `dir`-based BlobStorage implementation
     - Enhancement: Added Storage Migration logic that can generically migrate between different storage engines
--- a/packages/general/src/codec/DnsCodec.ts
+++ b/packages/general/src/codec/DnsCodec.ts
@@ -448,7 +448,15 @@ export class DnsCodec {
     static encodeTxtRecord(entries: (Bytes | string)[]) {
         const writer = new DataWriter();
         entries.forEach(entry => {
-            const bytes = typeof entry === "string" ? Bytes.fromString(entry) : Bytes.of(entry);
+            let bytes = typeof entry === "string" ? Bytes.fromString(entry) : Bytes.of(entry);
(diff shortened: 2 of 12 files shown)


Apollon77:
Thanks @copilot — all three findings addressed in 1a5a782f2:

1. **TXT entry length > 255 (DnsCodec.ts)** — encoder now logs a warning and truncates to the RFC 6763 §6.1 limit instead of silently wrapping via `writeUInt8`. Throwing was my first instinct but `MdnsSocketTest`'s "sends oversized answer records with warning" deliberately exercises a 2000-byte entry to validate that the wire layer doesn't crash on bad data, so warn+truncate is the better fit. Test added in `DnsCodecTest`.

2. **`MdnsTest.sortKey` lossy on binary (MdnsTest.ts:952)** — applied your suggested patch verbatim: project entries through `Bytes.toHex` after normalizing strings via `Bytes.fromString`. Both fixture (`string[]`) and decoded (`Uint8Array[]`) sides now produce the same lossless hex sort key.

3. **First-wins across multiple TXT records can keep older data (DnssdName.ts)** — collect the TXT records from `#records`, sort newest-first by `installedAt`, then apply first-wins per RFC 6763 §6.4 within each record's entry order. An updated TXT record's keys now win over the not-yet-expired older copy while still honouring the spec rule within a single record.

PR #3718, opened 2026-05-11
fix: nullable attribute encoding in TlvOfModel and chip-testing controller

Apollon77:
### fix(types): TlvOfModel uses effectiveQuality for nullable check

`generateTlv` checked `model.quality.nullable` which reads only the **local** quality field.

**Root cause**: `PeerBehavior.generateDiscoveredType()` calls `maybeOverrideSupport()` for every attribute. That function adds an attribute to `attrSupportOverrides` whenever `isSupported && applicability !== Mandatory` — i.e., every non-mandatory (conditional/optional) attribute the device supports gets `attr.extend({ operationalIsSupported: true })`. This is correct behavior — it marks "this optional attribute is present on this device".

`startUpOnOff` has conformance `"LT"` (conditional on Lighting) → not Mandatory → always extended via `attr.extend(...)` when the device reports it, **regardless of FeatureMap**. The extended model carries no local quality (only `operationalIsSupported: true`), so `model.quality.nullable` returned `false` even though the original attribute is `nullable`.

Fix: use `model.nullable` which calls `effectiveQuality`, traversing the `operationalBase` inheritance chain back to the original attribute definition where `nullable = true`.

This affects **all nullable conditional/optional attributes** on peer devices, not just attributes whose feature appears absent.

Adds `TlvOfModel` round-trip tests for nullable scalar (enum) and struct attributes on extended models.

### fix(chip-testing): fix double-encode bug writing nullable attributes

`LegacyControllerCommandHandler.handleWriteAttributeById` pre-encoded values to `TlvStream` before passing to the old `InteractionClient`, which then re-encoded using the attribute schema via `TlvOfModel`.

For nullable enums (e.g. `startUpOnOff`), the second encode received a `TlvElement[]` array instead of `null` → `Number(array) = NaN → 0` → `UnsignedInt(0)` on the wire.

Fix: pass raw `value` directly; `InteractionClient` encodes once via `TlvOfModel` with correct nullable schema. For unknown attributes (not in the Matter model), `inferAttrModel` builds a typed `AttributeModel` from the JS value type so `TlvOfModel` produces a concrete schema instead of `TlvAny` (which requires a pre-encoded `TlvStream`).

Adds `ClientNode` tests covering `null`/`0`/`2` writes for a nullable enum attribute via both `agent.state` assignment and `setStateOf` paths.

### DIFF
--- a/packages/node/test/node/ClientNodeTest.ts
+++ b/packages/node/test/node/ClientNodeTest.ts
@@ -45,6 +45,7 @@ import { Specification } from "@matter/model";
 import { ControllerCommissioner, FabricAuthority, FabricManager, PeerSet, Val } from "@matter/protocol";
 import { FabricIndex, NodeId, Status, StatusResponseError } from "@matter/types";
 import { AccessControl } from "@matter/types/clusters/access-control";
+import { OnOff } from "@matter/types/clusters/on-off";
 import { WindowCovering } from "@matter/types/clusters/window-covering";
 import { MyBehavior } from "../behavior/cluster/cluster-behavior-test-util.js";
 import { MockSite } from "./mock-site.js";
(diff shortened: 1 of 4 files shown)


Apollon77:
Addressed Copilot's review comment in 2ee9bd2b4: added `inferAttrModel` which builds a typed `AttributeModel` from the JS value type (`"uint32"`, `"uint64"`, `"string"`, `"bool"`, `"octstr"`) so `TlvOfModel` returns a concrete schema instead of `TlvAny`. Primitive type names resolve for standalone models via `ModelTraversal.fallbackRoot → permanentDatatypes`.

Apollon77:
Both review comments addressed in e3173435d:
- **Comment 1** (TlvAny / unknown attrs): addressed in 2ee9bd2b4 — `inferAttrModel` builds a typed `AttributeModel` from the JS value type so `TlvOfModel` returns a concrete schema (TlvInt32, TlvUInt64, etc.) instead of TlvAny.
- **Comment 2** (uint32 vs int32): changed fallback numeric type to `int32` to preserve signed range for negative values.

Apollon77:
Fixed in 399871620: `Bytes.isBytes(value)` replaces `instanceof Uint8Array`, covering all `ArrayBuffer`/`ArrayBufferView` forms. `Bytes` was already imported.

Apollon77:
Fixed in 5e32ceddb: added explicit `typeof value === "number" || value === null` branch for `int32`; everything else throws `Error` immediately instead of silently coercing through int32 encoding.

PR #3748, opened 2026-05-16
fix(types): emit IM IB list members in schema-defined tag order

### LINKED ISSUE #3747: TlvTaggedList encodes IM IB members in caller property order, violating Matter Core §10.6.1
# `TlvTaggedList` encodes IM IB members in caller property order, violating Matter Core §10.6.1

## Summary

matter.js's `TlvTaggedList` encoder emits list members in the order in which
the caller assigned properties on the value object, not in the order defined
by the schema. For Interaction Model IBs (CommandPathIB, AttributePathIB,
EventPathIB, ClusterPathIB) the Matter Core specification §10.6.1 ("Tag
Rules") requires the schema-defined tag order. Receivers that follow the
spec read the resulting out-of-order context tags as missing, treating
fields that were actually present (notably `endpointId`) as wildcards.

In practice this causes commands sent from a matter.js controller / server
to a multi-endpoint device with two endpoints of the same device type to
fan out to *every* endpoint exposing the named cluster: toggling one Home
Assistant entity flips both endpoints' physical outputs simultaneously.

## Reproduction

A two-endpoint Matter OnOff light implemented with rs-matter (verified
against current `project-chip/rs-matter` `main`), commissioned from Home
Assistant in "Matter Server (beta)" mode (which runs `matter-js/matterjs-
server`). Both endpoints are `DEV_TYPE_ON_OFF_LIGHT` on the same device.

Steps:

1. Commission the device to Home Assistant.
2. Two `light.*` entities appear in HA, each with a `unique_id` ending in
   the correct endpoint number (`...-MatterNodeDevice-1-MatterLight-6-0`,
   `...-MatterNodeDevice-2-MatterLight-6-0`).
3. From HA, call `light.turn_on` on either entity.

Observed: both entities transition to `on`; the device's GPIO and WS2812
both light up.

Expected: only the addressed entity toggles.

The same fault reproduces with a single `onOff.on` invoke directly from
matterjs-server's logs (`Invoke » @1:27•aa4e⇵9efe 1.onOff.on`) — the
device logs two `handle_on` invocations, one for `EP=1` and one for
`EP=2`, with `cmd.endpoint_id` set correctly to each. So the dispatch is
spec-compliant on the device side: the receiver expanded a wildcard
`CommandPath` to every endpoint that exposes the cluster.

## Root cause

`packages/types/src/tlv/TlvObject.ts`, `ObjectSchema.#encodeList`:

```typescript
#encodeList(writer: TlvWriter, value: TypeFromFields<F>, options?: TlvEncodingOptions) {
    const encodedFields = new Set<string>();
    for (const name of Object.keys(value)) {           // ← caller insertion order
        this.#encodeEntryToTlv(writer, name, value, options);
        encodedFields.add(name);
    }
    for (const name in this.fieldDefinitions) {
        if (encodedFields.has(name)) continue;
        this.#encodeEntryToTlv(writer, name, value, options);
    }
}
```

`packages/protocol/src/action/request/Invoke.ts`, `Invoke.Command`,
constructs `commandPath` with `clusterId` and `commandId` first, then
conditionally appends `endpointId`:

```typescript
const result: InvokeCommandData = {
    commandPath: {
        clusterId: Specifier.clusterFor(request.cluster).id,
        commandId: requestId,
    },
    ...
};
const endpointId = Specifier.endpointIdOf(request as ConcreteCommandRequest);
if (endpointId !== undefined) {
    result.commandPath.endpointId = endpointId;
}
```

V8's `Object.keys` preserves property-insertion order for string keys, so
the encoder writes context tags `1, 2, 0` (cluster, command, endpoint).

Wire bytes captured from a current matterjs-server build (decoded
manually):

```
17                    List, anonymous tag
  37 00               nested List, ctx tag 0  (commandPath inside commandData)
    24 01 06          U8 ctx tag 1: clusterId  = 6   (OnOff)
    24 02 01          U8 ctx tag 2: commandId  = 1   (On)
    24 00 01          U8 ctx tag 0: endpointId = 1
  18                  end-of-list
…
18                    end-of-container
```

## Spec analysis

Matter Core §10.6.1 ("Tag Rules", Matter 1.3 — 1.5.1, identical wording
across revisions):

> Unless otherwise noted, all context tags SHALL be emitted in the order
> as defined in the appropriate specification. This is done to reduce
> receiver side complexity in having to deal with arbitrary order tags.

CommandPathIB is defined in §10.6.11 as a TLV list with three context-
tagged members in tag-numeric order: `Endpoint = 0`, `Cluster = 1`,
`Command = 2`. AttributePathIB (§10.6.2), EventPathIB (§10.6.8), and
ClusterPathIB (§10.6.7) follow the same pattern.

§A.5.3 ("Lists") additionally states that "the meanings of member
elements in a list are denoted by their position within the list in
conjunction with any associated tag value they may have," reinforcing
that a list member's wire position is semantically significant — not an
encoder implementation detail.

Together these clauses mean a receiver is permitted (and §10.6.1
explicitly encourages it) to scan IB tags in ascending order and treat
tags it has already passed as absent. matter.js's CommandPathIB encoding
defeats that scan for the optional `endpointId`.

## Scope of impact

Likely affects any §10 IB whose schema starts with an optional field that
callers set conditionally:

  - **CommandPathIB** (`TlvCommandPath`): optional `endpointId` first.
    Direct reproduction above.
  - **AttributePathIB** (`TlvAttributePath`): all-optional, starting with
    `enableTagCompression` and `nodeId`. Latent — affected callers in
    matter.js's controller code may already populate these in schema
    order by convention, but the encoder offers no guarantee.
  - **EventPathIB** (`TlvEventPath`): all-optional, starting with
    `nodeId`. Same shape, same risk.
  - **ClusterPathIB** (`TlvClusterPath`): optional `nodeId` first.

Inner lists of operational certificates (subject / issuer / extensions)
are TLV lists too but are *not* covered by §10.6.1 — those carry
application-defined positional meaning, and matter.js's current
caller-order encoding is what makes signed certificates round-trip
bit-identically. Any fix should preserve that.

## Proposed fix

Make schema-order list encoding opt-in. Add an `inTagOrder` parameter
to `TlvTaggedList`; when set, `#encodeList` iterates the schema's field
definitions instead of `Object.keys(value)`. Set the flag on the four
§10 IB schemas. Default remains insertion-order so cert encoding stays
bit-stable.

PR with the implementation and regression tests is ready; will link from
this issue once filed.

## Tested with

- `@matter/types` test suite (`npx matter-test esm`): 311/311 with the
  three new regression tests in `TlvObjectTest.ts`.
- `@matter/protocol` test suite: 841/841, including all
  `FabricBuilder` / `Fabric` / `FabricAuthority` / `Certificate*` /
  `CertificateAuthority` tests that exercise certificate round-trips.
- `npm run lint` and `npm run format`: clean.

## Environment

- matter.js: current `main` branch (commit `008f898`)
- Node.js: 24.x (also run in CI against 20 / 22)
- Receiver: rs-matter `main` (last verified at commit `2418d1d`), running
  on ESP32-C6 with esp-idf-matter
- Home Assistant Matter Server (beta), latest stable matterjs-server
- Network path: laptop ↔ HA OTBR ↔ Thread mesh

snabb:
# tlv: emit IM IB list members in schema-defined tag order

Fixes #3747.

## Summary

`TlvTaggedList`'s default encoder emits list members in the caller's
property-insertion order, which for the §10 Interaction Model IBs
violates Matter Core spec §10.6.1 ("Tag Rules") and §A.5.3. The wire
bytes for an outgoing `1.onOff.on` invoke from matter.js currently come
out as context-tag order `1, 2, 0` (cluster, command, endpoint), and
spec-compliant receivers treat the late-arriving `endpointId` as absent
— silently turning a unicast invoke into a wildcard-endpoint fan-out.

This PR adds an `inTagOrder` opt-in to `TlvTaggedList`, sets it on the
four §10 IB schemas (`TlvCommandPath`, `TlvAttributePath`, `TlvEventPath`,
`TlvClusterPath`), and adds regression tests. Default behavior is
unchanged, so list schemas whose member position is application data
(notably Matter operational certificate inner lists, whose signed bytes
must reproduce caller order on re-encode) continue to round-trip
bit-identically.

The full bug analysis, wire-byte capture, and spec citations are in the
companion issue.

## What changed

`packages/types/src/tlv/TlvObject.ts`
  - `ObjectSchema` gains a fourth constructor parameter `inTagOrder`
    (default `false`).
  - `TlvTaggedList` exposes the flag via a third positional parameter.
  - `#encodeList` consults the flag: schema-order iteration when set,
    insertion-order iteration otherwise.

`packages/types/src/protocol/types/`
  - `TlvCommandPath`, `TlvAttributePath`, `TlvEventPath`, `TlvClusterPath`
    all opt in with `inTagOrder = true`. Each schema gains a doc-comment
    explaining the §10.6.1 requirement; `TlvCommandPath` carries the
    detailed rationale and the other three refer to it.

`packages/types/test/tlv/TlvObjectTest.ts`
  - Three new tests in a `Tlv Lists with inTagOrder` block:
    1. CommandPath-shaped schema with caller order `clusterId, commandId,
       endpointId`: wire bytes are schema order `0, 1, 2`.
    2. Wildcard endpoint (`endpointId` omitted): remaining members stay
       in schema order.
    3. Default behavior unchanged when `inTagOrder` is not set.

## Why opt-in rather than blanket

The first iteration of this fix changed `#encodeList` unconditionally to
schema order and broke 58 `FabricBuilder` / `Fabric` / `FabricAuthority`
tests with `Signature verification failed` inside `Noc.verify`. Root
cause: Matter operational certificates use `TlvTaggedList` internally
(`TlvGenericMatterSubjectOrIssuerTaggedList`, `TlvCertificateExtension`),
and signature verification depends on parse-then-reencode reproducing the
originally signed bytes — which means caller-order encoding is part of
the contract there. The spec backs this up: §A.5.3 makes list-member
position semantically significant in general, and the §10.6.1
schema-order requirement applies only to §10 IBs ("Unless otherwise
noted, all context tags SHALL be emitted in the order as defined in the
appropriate specification.").

An opt-in flag keeps the two semantics cleanly separated and is a
minimal-surface change.

## Tests

- `npx matter-test esm` in `@matter/types`: **311/311** (+ the 3 new
  tests in this PR).
- `npx matter-test esm` in `@matter/protocol`: **841/841**, including
  all `FabricBuilder` / `Fabric` / `FabricAuthority` / `Certificate*` /
  `CertificateAuthority` cases that exercise certificate round-trips.
- `npm run lint`: 0 warnings, 0 errors.
- `npm run format`: clean.

## End-to-end verification

The fix was developed against a two-endpoint OnOff light implemented
with rs-matter + esp-idf-matter on ESP32-C6 NanoC6 hardware. Without
this fix and without a receiver-side workaround, toggling one Home
Assistant `light.*` entity flips both endpoints' physical outputs. The
device's serial log makes the dispatch concrete:

  Before fix (matter.js writes tags 1, 2, 0; rs-matter scans ascending):

      OnOffHandler(EP=1, hooks=0x408405a8).handle_on cmd.endpoint_id=1
      OnOffHandler(EP=2, hooks=0x408405dc).handle_on cmd.endpoint_id=2

  After fix (matter.js writes tags 0, 1, 2):

      OnOffHandler(EP=1, hooks=0x408405a8).handle_on cmd.endpoint_id=1

## Backwards compatibility

- `TlvTaggedList` callers that don't pass the new third positional
  argument keep the existing encoding behavior exactly.
- The `ObjectSchema` constructor's new fourth parameter is also
  defaulted; external subclasses (if any) compile unchanged.
- No public type signature changes.
- Wire format from matter.js is now spec-compliant for the four §10 IB
  schemas. Receivers that accepted the old non-conformant order
  continue to parse the new order correctly (the new bytes are simply
  what the spec already required).

## Spec references

  - Matter Core 1.5.1 §10.6.1 ("Tag Rules") — schema-defined tag order
    requirement.
  - Matter Core 1.5.1 §10.6.11 (CommandPathIB), §10.6.2
    (AttributePathIB), §10.6.8 (EventPathIB), §10.6.7 (ClusterPathIB).
  - Matter Core 1.5.1 §A.5.3 ("Lists") — list-member position is
    semantically significant.
  - Matter Core 1.5.1 Appendix B.5.1 — `[tag-order]` / `[schema-order]`
    / `[any-order]` qualifiers.

Verified against Matter 1.3, 1.4, and 1.5 / 1.5.1 (PDFs from CSA).
Section numbers and wording are stable across all four. matter.js
targets 1.5.1 (`Specification.REVISION` in
`packages/model/src/common/Specification.ts`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -118,6 +118,7 @@ The main work (all changes without a GitHub username in brackets in the below li
     - Breaking: Removed some special pre-bound TLV string/byte-string schema exports, including `TlvHardwareAddress`
     - Feature: We've rewritten the typing system for clusters to simplify types, consume less runtime memory and work better with IDEs
     - Enhancement: Increases Matter TlvString/TlvByteString default maximum length to 65536 to cover WebRTC cases
+    - Fix: (@snabb) `TlvTaggedList` now ensures schema/tag ordering when encoding (Matter Core §10.6.1). Added `TlvTaggedListPreservingOrder` for cases where ordering needs to be data driven (e.g. signed certificate sub-lists).
 
 - @project-chip/matter.js
     - Feature: CommissioningController allows to set `tcp` (boolean) to enable TCP support for the controller
--- a/packages/protocol/src/certificate/kinds/definitions/operational.ts
+++ b/packages/protocol/src/certificate/kinds/definitions/operational.ts
@@ -32,9 +32,7 @@ import { ExtensionKeyUsageBitmap } from "./base.js";
  */
 export const MAX_TLV_CERTIFICATE_SIZE = 400;
 
-// Encoded extensions are part of the certificate's signed bytes, so a
-// parse + re-encode round-trip must reproduce the original on-the-wire
-// member order. Use the order-preserving variant for that reason.
+// Order-preserving: extensions are part of the certificate's signed bytes; re-encode must reproduce wire order.
 export const TlvCertificateExtension = TlvTaggedListPreservingOrder({
     basicConstraints: TlvField(
         1,
@@ -108,9 +106,7 @@ export namespace OperationalCertificate {
             dnQualifierPs: TlvOptionalField(142, TlvString),
             pseudonymPs: TlvOptionalField(143, TlvString),
         };
-        // The subject and issuer DN sub-lists are part of the certificate's
-        // signed bytes, so encoding must reproduce the caller-supplied
-        // member order on re-encode for signature verification to pass.
+        // Order-preserving: subject/issuer DN sub-lists are part of the certificate's signed bytes; re-encode must reproduce caller order.
         return TlvTaggedListPreservingOrder(fields);
     };
 
--- a/packages/protocol/src/certificate/kinds/definitions/operational.ts
+++ b/packages/protocol/src/certificate/kinds/definitions/operational.ts
@@ -17,7 +17,7 @@ import {
     TlvOptionalField,
     TlvOptionalRepeatedField,
     TlvString,
-    TlvTaggedList,
+    TlvTaggedListPreservingOrder,
     TlvUInt16,
     TlvUInt32,
     TlvUInt64,
@@ -32,7 +32,10 @@ import { ExtensionKeyUsageBitmap } from "./base.js";
  */
 export const MAX_TLV_CERTIFICATE_SIZE = 400;
 
-export const TlvCertificateExtension = TlvTaggedList({
+// Encoded extensions are part of the certificate's signed bytes, so a
+// parse + re-encode round-trip must reproduce the original on-the-wire
+// member order. Use the order-preserving variant for that reason.
+export const TlvCertificateExtension = TlvTaggedListPreservingOrder({
     basicConstraints: TlvField(
         1,
         TlvObject({
@@ -105,7 +108,10 @@ export namespace OperationalCertificate {
             dnQualifierPs: TlvOptionalField(142, TlvString),
             pseudonymPs: TlvOptionalField(143, TlvString),
         };
-        return TlvTaggedList(fields);
+        // The subject and issuer DN sub-lists are part of the certificate's
+        // signed bytes, so encoding must reproduce the caller-supplied
+        // member order on re-encode for signature verification to pass.
+        return TlvTaggedListPreservingOrder(fields);
     };
 
     /**
--- a/packages/types/src/protocol/types/TlvAttributePath.ts
+++ b/packages/types/src/protocol/types/TlvAttributePath.ts
@@ -15,8 +15,13 @@ import { TlvBitmap, TlvUInt16, TlvUInt32 } from "../../tlv/TlvNumber.js";
 import { TlvOptionalField, TlvTaggedList } from "../../tlv/TlvObject.js";
 import { WildcardPathFlagsBitmap } from "./WildcardPathFlagsBitmap.js";
 
-/** @see {@link MatterSpecification.v13.Core}, section 10.6.2 */
-
+/**
+ * @see {@link MatterSpecification.v13.Core}, section 10.6.2
+ *
+ * `TlvTaggedList` emits members in schema/tag order, satisfying §10.6.1
+ * ("Tag Rules") for Interaction Model IBs; see the doc-comment on
+ * `TlvCommandPath` for the full rationale.
+ */
 export const TlvAttributePath = TlvTaggedList({
     // AttributePathIB
     enableTagCompression: TlvOptionalField(0, TlvBoolean),
--- a/packages/types/src/protocol/types/TlvAttributePath.ts
+++ b/packages/types/src/protocol/types/TlvAttributePath.ts
@@ -15,13 +15,7 @@ import { TlvBitmap, TlvUInt16, TlvUInt32 } from "../../tlv/TlvNumber.js";
 import { TlvOptionalField, TlvTaggedList } from "../../tlv/TlvObject.js";
 import { WildcardPathFlagsBitmap } from "./WildcardPathFlagsBitmap.js";
 
-/**
- * @see {@link MatterSpecification.v13.Core}, section 10.6.2
- *
- * `TlvTaggedList` emits members in schema/tag order, satisfying §10.6.1
- * ("Tag Rules") for Interaction Model IBs; see the doc-comment on
- * `TlvCommandPath` for the full rationale.
- */
+/** @see {@link MatterSpecification.v13.Core}, section 10.6.2 */
 export const TlvAttributePath = TlvTaggedList({
     // AttributePathIB
     enableTagCompression: TlvOptionalField(0, TlvBoolean),
--- a/packages/types/src/protocol/types/TlvClusterPath.ts
+++ b/packages/types/src/protocol/types/TlvClusterPath.ts
@@ -9,8 +9,13 @@ import { TlvEndpointNumber } from "../../datatype/EndpointNumber.js";
 import { TlvNodeId } from "../../datatype/NodeId.js";
 import { TlvField, TlvOptionalField, TlvTaggedList } from "../../tlv/TlvObject.js";
 
-/** @see {@link MatterSpecification.v13.Core}, section 10.6.7 */
-
+/**
+ * @see {@link MatterSpecification.v13.Core}, section 10.6.7
+ *
+ * `TlvTaggedList` emits members in schema/tag order, satisfying §10.6.1
+ * ("Tag Rules") for Interaction Model IBs; see the doc-comment on
+ * `TlvCommandPath` for the full rationale.
+ */
 export const TlvClusterPath = TlvTaggedList({
     // ClusterPathIB
     nodeId: TlvOptionalField(0, TlvNodeId),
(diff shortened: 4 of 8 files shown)


Apollon77 (commented):
Ok, while I agree with the goal to "fix" this I am not sure if this complexity is needed ... Why we do not simply ensure the tag order for all cases? We usually generate all data Structures as defined in the spec, so I guess it makes sense to also simply use this for all TlvObject cases - which simply removes the need for the additionally added parameters to control this "ordering" selectively.

PS ok when looking at the posted details again then certificates would be more the "special case" than all normal Matter messages ... so turning the idea around and have a special parameter when these get encoded might make more sense maybe then the other way around, that potentially still leaves "sub objects" in encoding in a unexpected order (Because clients that have issues with that in IMs likely also have issues in other objects ...)

So s asked in the issue it would be great to know more details about who exactly had issues with this

Apollon77:
Thinking again I would maybe introduce a second TaggedTlvList "helper" which preserves the original order and use this in the certificates and turn the logic to "preservceDataordering" and otherwise always respect the tag orders of the object type ... this should then solve this issue for all IM messages and all data (also e.g. the data of an invoke command).

Apollon77:
Re "the interaction model revision" issue. Ok interesting that chip tool allows it to be absent. I will also check and clarify that. Here i know clients that require it or will also fail. So matter.js is not alone here when I remember right.

snabb:
Pushed `90e14ba` (rebased onto current `main`) which restructures along the lines you suggested.

**What changed vs. the original push**

- `TlvTaggedList` now emits members in schema/tag order by default — no opt-in flag at the call site. Since matter.js declares fields in spec order, this satisfies §10.6.1 for all IBs and any other invoke/data sub-object that uses `TlvTaggedList`, not just the four §10 IBs the first version touched.
- Added a sibling helper `TlvTaggedListPreservingOrder` that keeps caller-supplied member order on encode. Used in exactly two places, both in `packages/protocol/src/certificate/kinds/definitions/operational.ts`: `TlvCertificateExtension` and `TlvGenericMatterSubjectOrIssuerTaggedList`. These are the inner lists of operational certificates whose signed bytes must round-trip bit-identically for signature verification — the cases that broke the 58 fabric tests when I first tried a blanket switch.
- The four `TlvCommandPath` / `TlvAttributePath` / `TlvEventPath` / `TlvClusterPath` files no longer carry a per-schema flag; doc-comments now point at `TlvCommandPath`'s rationale.
- `ObjectSchema`'s internal flag was renamed `preserveDataOrdering`. Three pre-existing `TlvTaggedList` tests that asserted caller-order preservation were retitled and updated to assert the new schema-order default; one new test covers `TlvTaggedListPreservingOrder`.

**Scope of the encode change**

This only affects the *sender* side — the decoder continues to accept context tags in any order, so nothing changes for matter.js as a receiver. Peers that send out-of-order tags keep working.

**Re: who had issues**

It's a rs-matter device commissioned to a matter.js controller (Project CHIP `rs-matter`, with `esp-idf-matter` on ESP32-C6 NanoC6 hardware). rs-matter's IM decode scans context tags in ascending order — spec-permissible per §10.6.1 ("to reduce receiver side complexity in having to deal with arbitrary order tags") — so the late-arriving `endpointId = 0` was read as absent and the invoke fanned out to every endpoint exposing the named cluster. Symptom only appears with two endpoints of the same device type, which is probably why it hasn't surfaced before. Agreed it's a gap in interop testing on both sides.

**Tests**

- `@matter/types`: 311/311.
- `@matter/protocol`: 842/842, including all `FabricBuilder` / `Fabric` / `FabricAuthority` / `Certificate*` / `CertificateAuthority` round-trip cases.
- `npm run lint`, `npm run format`: clean.

Apollon77:
@snabb That looks better and I thunk makes more sense ... ets see what testing says ... rest is mainly the usual "over documenting of AIs" but I will fix that

Apollon77 (approved):

PR #3786, opened 2026-05-23
fix(protocol): preserve EventNumber order in normalizeEventData

### LINKED ISSUE #3785: Events dispatched out of order: normalizeEventData regroups occurrences by path
## Summary

`normalizeEventData` (`packages/protocol/src/interaction/EventDataDecoder.ts:74-91`)
groups occurrences by `(nodeId, endpointId, clusterId, eventId)` into a Map
and returns `Array.from(map.values())`. JS Map iteration is insertion order,
so the outer array becomes "first occurrence per path". `InputChunk`
(`packages/protocol/src/action/client/InputChunk.ts:39-40`) then walks
outer-then-inner, dispatching events in the regrouped order — destroying
the wire-order (EventNumber) sequence sent by the publisher.

The annotation `// TODO CHECK` on the grouping line suggests the regrouping
was added speculatively and not since revisited.

## Reproduction

Stock ESP32-C6 Generic Switch (rs-matter based) wired as a triple-press
button. Verified with a temporary device-side log that the wire layout
is correct: events `InitialPress(en=N) → ShortRelease(en=N+1) → InitialPress(en=N+2) → MultiPressOngoing(en=N+3) → ShortRelease(en=N+4) → InitialPress(en=N+5) → MultiPressOngoing(en=N+6) → ShortRelease(en=N+7) → MultiPressComplete(en=N+8)`
go out in a single `IM::ReportData`, EventNumber-monotonic.

`ClientEventEmitter` then logs them in this order:

```
initialPress
shortRelease, shortRelease, shortRelease
initialPress, initialPress
multiPressOngoing(2), multiPressOngoing(3)
multiPressComplete(3)
```

The signature ("all SRs, then all IPs, then all MPOs, then MPC") is
exactly what the Map-grouping in `normalizeEventData` produces: groups
appear in first-occurrence order; occurrences within a group preserve
EventNumber order; outer/inner iteration in `InputChunk` flattens to
the broken sequence.

## Why it matters (and why HA still works today)

The Generic Switch cluster (Application Cluster 1.5.1 §1.13) defines its
semantics as ordered cycles:

> The above constraints imply that for a given activity detection cycle
> of a switch having MSM and/or MSL feature flags set, the entire
> activity is either a single long press detection cycle of
> *(InitialPress, LongPress, LongRelease)*, or a single multi-press
> detection cycle (ending in MultiPressComplete), where presses that
> would otherwise be reported as long presses are instead reported as
> a counted press in the MultiPressComplete event…

After regrouping, the cycle structure is unobservable: there is no
metadata in the events themselves that lets a consumer reconstruct
which cycle a given event belongs to. EventNumber is the only
ordering key, and we discard it before dispatch.

Matter Core 1.5.1 §10.7.3.1 backs this up — the EventReports wire layout uses
an "omit EventNumber if exactly one greater than the previous Event"
compression scheme, which is only safe if the report is read in order.
The spec's own example interleaves two event types (`Started`,
`PatternChange`) on the same path in sequential EventNumber order.
matter.js's existing precedent for attributes (#3290 "Preserve Attribute
update order") applies the same expectation.

HomeAssistant's matter integration happens to be immune because for
multi-press switches it ignores `InitialPress` / `ShortRelease` /
`MultiPressOngoing` entirely and reads only
`MultiPressComplete.totalNumberOfPressesCounted` — a self-contained
summary field. Custom matter.js consumers, automations that fire on
per-press events, and any cluster whose events repeat within a single
report and depend on order would all break.

The CHIP Python SDK delivers events through a per-event
callback (`SubscriptionTransaction.SetEventUpdateCallback`) invoked
once per `EventReadResult` in arrival order — see
`connectedhomeip/src/controller/python/matter/clusters/Attribute.py`,
the `handleEvent` path that calls `OnEventChangeCb(eventResult, …)`
for every received event. python-matter-server (the predecessor that
matter.js replaces in HA) consumes that API directly and so preserves
wire order by construction. matter.js is the outlier among Matter
controller implementations on this.

## Suggested fix

Drop the regrouping; return one occurrence per outer entry in wire order.
The regrouping was a decode-cache optimization (one schema lookup per
path tuple in `normalizeAndDecodeEventData` rather than per occurrence)
with no spec basis. `InputChunk` and `normalizeAndDecodeEventData` work
unchanged because the outer/inner iteration just flattens to wire order
now.

A candidate patch with this shape is at
<https://github.com/snabb/matter.js/tree/preserve-event-order>; PR to
follow. Protocol package tests pass (842/842 ESM), node package tests
pass (1147/1147 ESM). Existing `EventDataDecoderTest` cases that
codified the broken behavior are updated to assert wire-order output.

snabb:
Fixes #3785.

`normalizeEventData` grouped occurrences by `(nodeId, endpointId, clusterId, eventId)` into a Map and returned `Array.from(map.values())`, so the outer iteration became "first occurrence per path". `InputChunk` walks outer-then-inner, which destroyed wire (EventNumber) order whenever the same path repeated within a single `ReportData`.

Matter Core 1.5.1 §10.7.3.1 emits `EventReports` in EventNumber order (the EventNumber-omitted compression rule depends on it), and Application Cluster 1.5.1 §1.13 defines the Generic Switch cycle (`InitialPress → MultiPressOngoing → ShortRelease → … → MultiPressComplete`) as an ordered sequence; the CHIP Python SDK's `SubscriptionTransaction.SetEventUpdateCallback` API contract delivers per-event in arrival order.

This change returns one occurrence per outer entry, in wire order. `normalizeAndDecodeEventData` and `InputChunk` work unchanged; their outer/inner iteration now simply flattens to wire order. The previous grouping was a decode-cache optimization with no spec basis. Tests that codified the broken behavior are updated.

Test results: protocol 842/842 ESM, 1147/1147 node ESM.

### DIFF
--- a/packages/protocol/src/interaction/EventDataDecoder.ts
+++ b/packages/protocol/src/interaction/EventDataDecoder.ts
@@ -74,20 +74,15 @@ export function normalizeAndDecodeReadEventReport(data: TypeFromSchema<typeof Tl
 export function normalizeEventData(
     data: TypeFromSchema<typeof TlvEventData>[],
 ): TypeFromSchema<typeof TlvEventData>[][] {
-    // Put all returned values into a map to group by path
-    const responseList = new Map<string, TypeFromSchema<typeof TlvEventData>[]>(); // TODO CHECK
-    data.forEach(value => {
-        if (!value) return;
-        const {
-            path: { nodeId, endpointId, clusterId, eventId },
-        } = value;
-        const mapId = `${nodeId}-${endpointId}-${clusterId}-${eventId}`;
-        const list = responseList.get(mapId) || [];
-        list.push(value);
-        responseList.set(mapId, list);
-    });
-
-    return Array.from(responseList.values());
+    // One occurrence per outer entry, preserving wire (EventNumber) order
+    // per Matter Core §10.7.3.1. Earlier code regrouped by path tuple,
+    // which broke ordered event cycles such as Generic Switch multi-press.
+    const result: TypeFromSchema<typeof TlvEventData>[][] = [];
+    for (const value of data) {
+        if (!value) continue;
+        result.push([value]);
+    }
+    return result;
 }
 
 export function normalizeAndDecodeEventData(
--- a/packages/protocol/test/protocol/interaction/EventDataDecoderTest.ts
+++ b/packages/protocol/test/protocol/interaction/EventDataDecoderTest.ts
@@ -62,7 +62,7 @@ describe("EventDataDecoder", () => {
             ]);
         });
 
-        it("normalize data with all paths given for multiple events", () => {
+        it("preserves wire (EventNumber) order across interleaved event types", () => {
             const data: TypeFromSchema<typeof TlvEventData>[] = [
                 {
                     path: { endpointId: EndpointNumber(0), clusterId: ClusterId(0x28), eventId: EventId(0) },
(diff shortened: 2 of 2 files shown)


Apollon77:
The fix makes sense, we should not modify the order in  any way, this was a bad idea and sure with switch it really makes issues

Apollon77:
In fact I need to look into it if the normalization now makes any sense anymore or if I not do a bigger refactor to solve it

Apollon77:
Superseded by #3787, which folds the event-ordering fix into a wider refactor of the DataReport decode pipeline (streaming `InputChunk`, legacy `DecodedDataReport` relocation, type tightening, async iteration). Your diagnosis and original patch are credited in the new PR's body and in the first commit's message. Thanks @snabb!

PR #3787, opened 2026-05-24
refactor(protocol,matter.js): streaming InputChunk + DecodedDataReport relocation

### LINKED ISSUE #3785: Events dispatched out of order: normalizeEventData regroups occurrences by path

Apollon77:
## Summary

- Rewrites `@matter/protocol`'s DataReport decode pipeline as a streaming async generator (`InputChunk` → `ReadResult.Chunk`). Decode runs per fully-qualified attribute path; chunked-array reassembly and tag compression stay spec-compliant; events stream per-occurrence in wire (EventNumber) order. Fixes #3785 (Generic Switch multi-press cycle was destroyed by the previous Map-based grouping). Thanks to @snabb for the diagnosis and original fix proposal in #3786 — this PR supersedes it with a wider refactor.
- Relocates the legacy 4-array `DecodedDataReport` shape (interface + `Decoded*` types + `normalize*` helpers) into `@project-chip/matter.js/cluster` where the only consumers live (legacy `InteractionClient`, `PairedNode`, `ClusterClient`, `EventClient`). Protocol-side copies stay with `@deprecated` JSDoc for one release and are scheduled for hard removal in 0.18.
- `ReadResult.Chunk` is now `AsyncIterable<Report> | Iterable<Report>`. Sync server-side producers (`AttributeReadResponse.process` et al.) keep their `yield [report]` shape; the streaming wire-decode producer uses async to let TCP transports deliver large DataReports without blocking the event loop.
- `ReadResult.EventValue` gains the four explicit wire timestamp variants (`epochTimestamp`, `systemTimestamp`, `deltaEpochTimestamp`, `deltaSystemTimestamp`) alongside the existing collapsed `timestamp: number` (Matter Core §10.7).
- Bug fixes folded in: `attr-status` / `event-status` no longer double-emit when both `status` and `clusterStatus` are set; chunked-array leftover handoff no longer mutates the caller's `attributeReports` array via splice; dead exports removed.

Closes #3785. Supersedes #3786.

## Commits

1. Streaming `InputChunk` + dead-code cleanup
2. matter.js `DecodedDataReport` adapter + consumer rewires + `@deprecated` shims
3. Stale `#clusterClients` field deleted from `ControllerCommissioningFlow`
4. Type tightening: status fields required, four timestamp variants on `EventValue`
5. `ClusterClientTypes.ts` relocated to matter.js
6. Greg-review fixes: shared chunk-to-decoded helpers, narrowed `UnexpectedDataError` catches, naming, deprecation tag wording
7. `ReadResult.Chunk` → async-iterable; `InputChunk` → async generator
8. Changelog entries

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,12 @@ The main work (all changes without a GitHub username in brackets in the below li
 - @matter/general
     - Fix: `causedBy`/`asError`/`errorOf`/`repackErrorAs` no longer crash with "undefined is not an object" when invoked with `undefined`/`null` as error object
 
+- @matter/protocol
+    - Deprecation: Internally used `DecodedDataReport`, `DecodedAttributeReport{Value,Status,Entry}`, `DecodedEventReport{Value,Status,Entry}`, `DecodedEventData`, and the `normalize*` / `normalizeAndDecode*` helpers moved to `@project-chip/matter.js/cluster`. Scheduled for removal in 0.18
+    - Enhancement: `ReadResult.EventValue` exposes the four wire timestamp variants (`epochTimestamp`, `systemTimestamp`, `deltaEpochTimestamp`, `deltaSystemTimestamp`) alongside the existing collapsed `timestamp: number`
+    - Adjustment: `ReadResult.Chunk` and `InputChunk` are now async generators; should be only used internally
+    - Fix: Event reports are now decoded in wire (EventNumber) order
+
 ## 0.17.0 (2026-05-20)
 
 - Breaking: Matter 1.5/1.5.1 specification introduces some changes, as always with new Matter specification versions. You might need to adjust your code.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,7 @@ The main work (all changes without a GitHub username in brackets in the below li
 - @matter/protocol
     - Deprecation: Internally used `DecodedDataReport`, `DecodedAttributeReport{Value,Status,Entry}`, `DecodedEventReport{Value,Status,Entry}`, `DecodedEventData`, and the `normalize*` / `normalizeAndDecode*` helpers moved to `@project-chip/matter.js/cluster`. Scheduled for removal in 0.18
     - Enhancement: `ReadResult.EventValue` exposes the four wire timestamp variants (`epochTimestamp`, `systemTimestamp`, `deltaEpochTimestamp`, `deltaSystemTimestamp`) alongside the existing collapsed `timestamp: number`
-    - Adjustment: `ReadResult.Chunk` and `InputChunk` are now async generators; should be only used internally
+    - Adjustment: `ReadResult.Chunk` may now be an async iterable (`InputChunk` is an async generator); consumers iterate chunk contents with `for await … of chunk`. Mainly internal
     - Fix: Event reports are now decoded in wire (EventNumber) order
 
 ## 0.17.0 (2026-05-20)
--- a/packages/matter.js/src/cluster/client/ClusterClient.ts
+++ b/packages/matter.js/src/cluster/client/ClusterClient.ts
@@ -6,7 +6,7 @@
 
 import { capitalize, Diagnostic, Duration, ImplementationError, Logger } from "@matter/general";
 import { AttributeModel } from "@matter/model";
-import { ClusterClientObj, DecodedEventData } from "@matter/protocol";
+import { ClusterClientObj } from "@matter/protocol";
 import {
     AttributeId,
     ClusterId,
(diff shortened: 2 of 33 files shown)


Apollon77:
Addressed Copilot + CodeQL review comments in 91464beab:

- **CodeQL (InputChunk.ts:114) — Unneeded defensive code.** `buildAttrStatus` return type tightened to `ReadResult.AttributeStatus` (drops `| undefined`); callsite's dead `if (status !== undefined)` guard removed.
- **Copilot (InputChunk.ts:210) — `dataVersion as number` cast.** Replaced with `dataVersion ?? 0` sentinel; chunk now honors its `version: number` contract. Comment notes spec-compliant publishers always set dataVersion in practice.
- **Copilot (CHANGELOG.md:20) — wording.** Adjusted to clarify `Chunk` is a type alias permitting async-iterable values and `InputChunk` is the async generator; consumers use `for await … of chunk`.

PR #3841, opened 2026-06-04
fix(node): resolve self-bindings to runtime-added cluster servers

Apollon77:
## Problem

`BindingManager.#endpointHasClusterServer` read `endpoint.type.behaviors` to decide whether a self-binding target serves a cluster. But cluster servers added at runtime via `endpoint.behaviors.require(...)` land in `endpoint.behaviors.supported`, never in `type.behaviors` (`Behaviors` clones off the shared `type.behaviors` reference on first inject). So self-bindings to runtime-installed cluster servers were wrongly rejected.

## Fix

- Read `endpoint.behaviors.supported` instead of `endpoint.type.behaviors`.
- Exclude client cluster behaviors via `isClientBehavior` — `behaviors.supported` also holds client `ClusterBehavior`s (installed for bindings), and a client shares its server counterpart's `cluster.id`, so a pure id match would false-positive on a client-only endpoint.
- Dropped the dead `=== undefined` guard (`supported` is always an object).

The mirror `#selectClientClusters` was reviewed and is correct as-is: client cluster declarations are static metadata living only in `type.clientClusters`, never runtime-injected into a source endpoint's `supported`.

## Tests

Two regression tests (both verified red→green):
- `kind=server` accepts a cluster server added at runtime via `behaviors.require`.
- `kind=server` rejects when the endpoint only has the cluster as a client, not a server.

Full `@matter/node` suite green (ESM/CJS/Web), lint + format clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,9 @@ The main work (all changes without a GitHub username in brackets in the below li
 - @matter/general
     - Fix: An empty mDNS TXT record is now encoded as a single zero byte as required by RFC 6763 §6.1
 
+- @matter/node
+    - Fix: Self-bindings now detect cluster servers added to an endpoint at runtime via `behaviors.require`, and no longer treat a client cluster as satisfying the server check
+
 - @matter/protocol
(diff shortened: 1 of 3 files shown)


Luligu:
Thanks

Issue #3843, opened 2026-06-05
Example for toggling a commissioned device not working


NauchtanRobotics:
https://github.com/matter-js/matter.js/blob/b2c9f3f658838b096258c13415c38ebdcc02e960/examples/control-onoff/src/OnOffController.ts#L44

To reproduce this failure, commission a device using temporary pairing code and confirm that files have been added on Mac under /Users/<username>/.matter/node0/*

Run command:
`node examples/control-onoff/src/OnOffController.ts toggle 1

Response:
"Cannot toggle because there is no commissioned device".

Debugging:
case "toggle":
    ...
    await controller.run();  // Does not help, but tried it anyway.
    console.log("Peers:", controller.peers);

Response is:
Peers: {}

What am I doing wrong?

NauchtanRobotics:
I am not running matterjs-server during this test as I assumed that matter.js works at a lower level and does not require the socket communications provided my matter-server.

NauchtanRobotics:
From the matter.js/README.md:

"Building Matter controllers:
Develop Matter controllers with JavaScript/TypeScript for OS-based IoT products. Post-commissioning (including BLE support), controllers can access all data from devices and control them. The API supports development with known typings of all clusters when endpoints and device types are known during development. It also allows generic programmatic access to all clusters and attributes. Typing information of custom clusters can be added to the controller or accessed generically with generic data parsing support. If you plan to build/certify a controller based on matter.js please see note below.
"
My use case is creating a simple controller that switches on a Matter smartplug whenever I am producing excess solar power.  Is this controller example not working because a Smartplug is not an OS-based IoT product?  So I'd need to be running matterjs-server for this to work?

Apollon77:
Yes, matterjs-server is one implementation of a whole and very convenient controller with a lot of features around it. matter.js is a library, so you can use it to build your own logic.

Ok in fact it should work, even that this example - as stated in the file - is using the "new" API of the controller which is basically in finalization still and not s convenient as the "old" one .

So you first did "commission " and commissioned a peer, right?
What device you used? if you also used a matter.js example that might produce collissions in the storage leading potentially to removing the commissioning. so I would need more details what you did any how and such. Maybe try to explicitly pin the storage dir --storage-path=PATH-HERE.

So or so I found a bug in the argument logic, but when you come that far it seems you already fixed that one :-)

jbaron:
[...]

PR #3847, opened 2026-06-05
fix(examples): correct argv parsing and isolate storage in OnOffController

Apollon77:
## Problem

The `control-onoff` controller example never dispatched commands correctly.

Line 19 dropped **two extra arguments**:

```js
const [, , command, ...args] = process.argv.slice(2);
```

`process.argv.slice(2)` already removes `node` + the script path; the
`[, , command, ...]` destructure then skips the first two *real* args. So
`node OnOffController.ts toggle 1` resolves `command` to `undefined` and falls
through to `default` → `"Unsupported command"`. Verified across plain `node`,
`tsx`, and `node --import tsx` — all give the same broken parse.

Secondarily, the controller called `ServerNode.create()` with no id. The
fallback id is `node${counter}` where the counter is in-memory and resets to
`0` each process, so the controller shares `~/.matter/node0` with the device
examples (e.g. the on/off light), which can clobber its peer storage.

## Fix

- Parse with `const [command, ...args] = process.argv.slice(2);`
- Report the actual `command` in the unsupported-command message (was printing `process.argv[1]`, the script path)
- Give the controller an explicit id (`ServerNode.create({ id: "controller" })`) so its storage is isolated from device examples

## Migration note

The explicit id moves controller storage from `~/.matter/node0` to
`~/.matter/controller`. Anyone who already commissioned with the old example
will need to re-commission once.

Related to #3843

### DIFF
--- a/examples/control-onoff/src/OnOffController.ts
+++ b/examples/control-onoff/src/OnOffController.ts
@@ -14,9 +14,9 @@ import { ServerNode } from "@matter/main";
 import { OnOffClient } from "@matter/main/behaviors/on-off";
 
 // Create the controller
-const controller = await ServerNode.create();
+const controller = await ServerNode.create({ id: "controller" });
 
-const [, , command, ...args] = process.argv.slice(2);
+const [command, ...args] = process.argv.slice(2);
 
 switch (command) {
     case "commission":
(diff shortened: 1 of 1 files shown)


Apollon77:
Paired with #3874, which fixes the library-side root cause: peers commissioned with a custom id (`commission({ id })`) were dropped from storage on restart, so `peers.get(id)` returned nothing in a fresh process. Together these fully resolve #3843 — this PR fixes the example's argv parsing and storage-id isolation, #3874 fixes the persistence.

PR #3849, opened 2026-06-05
fix(protocol): evict least-recently-active exchange instead of oldest-created

Apollon77:
## Summary

When a session exceeds `MAXIMUM_CONCURRENT_OUTGOING_EXCHANGES_PER_SESSION`, `ExchangeManager.#cleanupSessionExchanges` previously closed `sessionExchanges[0]` — the **oldest created** exchange in Map order, regardless of how recently it was used. A long-lived but still-active exchange created first could be killed in favor of a freshly-created idle one.

Root cause: a `MessageExchange` had no own activity timestamp — it only forwarded `session.notifyActivity()`, which updates the *session*, not the exchange. Nothing to sort by.

## Changes

- `MessageExchange` now tracks a per-exchange `lastActive` timestamp, stamped via a `#notifyActivity()` wrapper at the three sites that forward to `session.notifyActivity()` (one receive, two send/retransmit).
- `#cleanupSessionExchanges` now evicts the **least-recently-active** exchange (`reduce` over min `lastActive`) instead of the oldest-created one. Log text updated to match.
- Clarified the stale comment that said "5 concurrent exchanges" (spec value) while the constant is 30 (practical limit).

Strictly better than the previous FIFO-by-creation eviction. A back-pressure creation queue (replacing the hard limit entirely) remains a separate, larger follow-up.

## Test

New `MessageExchangeTest.ts` "activity tracking" tests verify `lastActive` updates on receive and on send (MockTime-driven). Full protocol suite passes (1101/1101), format and lint clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ The main work (all changes without a GitHub username in brackets in the below li
     - Fix: MRP retransmissions now use the idle interval when the peer left its active window mid-exchange, matching CHIP SDK behavior
     - Fix: TcpChannel now closes the connection on a zero-length stream frame instead of silently skipping it, preventing a peer from holding a connection slot open indefinitely
     - Fix: A forward message-counter jump of exactly the window size no longer retains stale replay-bitmap bits, which could wrongly reject a later legitimate message as duplicate
+    - Fix: When the per-session concurrent-exchange limit is exceeded, the least-recently-active exchange is now closed instead of the oldest-created one
 
 ## 0.17.1 (2026-06-03)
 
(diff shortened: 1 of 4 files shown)


Apollon77:
Thanks @copilot — addressed in b39093486:

1. **`reduce` param name** — renamed `oldest` → `leastRecentlyActive` to match the min-`lastActive` selection.
2. **Log wording** — dropped "outgoing" from the eviction log. The filter counts all non-closed exchanges for the session (not just initiator/outgoing); restricting it to outgoing would be a behavioral change and is out of scope here.
3. **Targeted eviction test** — skipped intentionally. Testing `#cleanupSessionExchanges` requires a fully-wired `ExchangeManager` (TransportSet, SessionManager, ExchangeCounter, lifetime, multiplex) with no existing test harness, and we avoid extracting the selection into a helper purely for testability. The selection is a standard min-by `reduce`, and the new `lastActive` stamping (the actual mechanism) is covered by the added unit tests.
4. **Verification** — ran the full required gates locally, all green:
   - `npm run build -- --clean` → exit 0 (type check + transpile)
   - `npm run lint` → exit 0
   - `npm run format-verify` → all files conform
   - `npm test -- -p packages/protocol` → 1101/1101 (ESM/CJS/Web)

PR #3964, opened 2026-06-23
fix(protocol, node): accept CASE/PASE session intervals above the 1-hour DNS-SD cap

### LINKED ISSUE #3962: Adjust CASE session parameter checks
... and allow even higher values than the 1h limits for MDNS

fixes https://github.com/home-assistant/core/issues/174542

Apollon77:
## Problem

A device whose CASE session parameters carry an idle/active interval **above one hour** was declined by matter.js during PASE/CASE establishment (see home-assistant/core#174542). The 1-hour `SII`/`SAI` maximum is a property of the **DNS-SD advertisement encoding only** — the CASE/PASE `session-parameter-struct` carries `SESSION_IDLE_INTERVAL`/`SESSION_ACTIVE_INTERVAL` as full-range `uint32` values with no bound. This matches the reference SDK (see CHIP-Specifications/connectedhomeip-spec#13177, Reading A).

fixes matter-js/matter.js#3962

## Changes

- **`SessionIntervals()`** no longer throws on out-of-range intervals; it applies defaults only. The receive path (`Session.timingParameters` setter, fed by peer Sigma1/Sigma2/PBKDFParamReq/Resp) now accepts the peer's full-range values. `SAT` stays naturally bounded by its `uint16` wire type.
- **`SessionIntervals.forAdvertisement()`** (new) clamps `SII`/`SAI` to one hour and `SAT` to 65535 ms for DNS-SD advertisement — capping instead of rejecting, matching the SDK. `MdnsAdvertisement` uses it.
- **Controller side** (`RemoteDescriptor.toLongForm`): a peer's mDNS advertisement at the one-hour cap no longer lowers a higher CASE-negotiated interval already on record. The guard only triggers when the advertised value is exactly the cap **and** the stored value is higher; a genuine interval change (re-advertised below the cap) still applies.

## Tests

- `SessionIntervalsTest`: accept >1h intervals / >65535 ms threshold via the base function; `forAdvertisement` clamps each to its spec maximum.
- `RemoteDescriptorTest`: capped advertisement does not lower a higher session-derived value; a below-cap advertisement still applies; a capped value applies when nothing higher is on record.

Full `@matter/protocol` and `@matter/node` suites pass (ESM/CJS/Web).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ The main work (all changes without a GitHub username in brackets in the below li
 
 - @matter/node
     - Enhancement: Free a behavior's persisted seed values from memory once the datasource has loaded them
+    - Fix: A peer's mDNS advertisement at the 1-hour SII/SAI cap no longer lowers a higher CASE-negotiated idle/active interval already on record
     - Fix: Ensure that a peer's FeatureMap change rebuilds the affected client cluster behavior
     - Fix: Ensure to always sanitize fabric-scoped attribute data (e.g. stale AccessControl ACL entries) at node startup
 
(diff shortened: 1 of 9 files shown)


Apollon77:
Addressed the Copilot feedback in d4a4dbf09: the three `forAdvertisement` cap log messages now derive the limit from the constants via `Duration.format(maxAdvertisedInterval)` / `Duration.format(maxActiveThreshold)` instead of hard-coding "1 hour"/"65535ms", so they stay correct if the caps ever change.

Same commit also adds up-front validation of a locally-configured Session Active Threshold (uint16 max) at the `SessionManager` config boundary, rather than failing late at message encoding.

PR #4211, opened 2026-08-07
fix: prevent subscription report failures from tearing down controller sessions

Apollon77:
## Summary

Fixes a critical regression where a single unacked subscription report would trigger peer loss inference and force-close every session with the controller. The issue is compounded when controllers disappear — sessions now accumulate without bound instead of being reaped by peer-loss teardown.

## Changes

**Core fix (@matter/protocol):**
- `MessageExchange.Options.suppressPeerLoss`: Waive peer-loss inference for exchanges that inherently have zero prior contact (e.g., subscription reports). The "never heard from this peer" heuristic doesn't apply to fresh exchanges on established sessions.
- Convey the failing exchange through peer-loss teardown so subscriptions can disambiguate subscription-specific failures from peer loss.

**Session management (@matter/protocol):**
- Session eviction by least-recently-used activity when a peer establishes more than five sessions. TCP disconnect still evicts all sessions bound to that connection.
- Wind down sessions explicitly for counter rollover and re-establish state names.

**Subscription handling (@matter/node):**
- InteractionServer now opens subscription exchanges with `suppressPeerLoss: true`.
- SubscriptionsServer abandons subscriptions after repeated send failures (using proper `causedBy()` check for MRP exhaustion errors).
- Fixed: `establishFormerSubscription` now closes the exchange it opens (prevents session from hanging in closing state) and uses the provided session instead of resolving peer address.
- Fixed: Stop spending a full MRP window on each subscription when a peer is unreachable.

## Testing

All tests pass (1491 protocol + 1645 node tests). Includes regression tests for session eviction and subscription abandonment.

## Checklist

- [x] All tests pass (full suite + protocol + node)
- [x] Format verification passes
- [x] Lint passes  
- [x] Build succeeds
- [x] CHANGELOG.md updated

🤖 Generated with Claude Code


Apollon77:
already merged

PR #4223, opened 2026-08-12
fix(model): feature conformance analysis rejects valid selections

### LINKED ISSUE #4215: IllegalFeatureCombinations inverts bracketed OR feature conformance
Since the 0.18 nightlies enforce FeatureSelectionErrors at endpoint add, a feature with conformance `[A | B]` is rejected when selected alongside its disjuncts.

IllegalFeatureCombinations routes the optional-if OR case through addExclusivityRequirement, emitting {feature, A, B all true} as illegal. Per the conformance grammar `[A | B]` means optional while A or B is supported, so the illegal combination is the feature with neither disjunct selected.

Repro on 0.18.0-alpha.0-20260807-335c3291b: mounting a CameraAvStreamManagement server with video + snapshot + imageControl (ImageControl conformance is `[VDO | SNP]`) fails with

```
Error in CameraAvStreamServer.features: features ImageControl, Video and Snapshot cannot be selected together
```

One note for the fix: `[A | B | C]` parses as nested ORs, so routing through addDependencyRequirement's two-name OR branch would silently drop the rule for CameraAvSettingsUserLevelManagement.MPRESETS (`[MPAN | MTILT | MZOOM]`). Flattening the disjuncts via extractDisjunctFeatures into one bitmap ({feature: true, every disjunct: false}) handles both shapes, that is what I am running as a local patch right now.

Apollon77:
Feature selection assessment reports several conforming selections as illegal. Four distinct defects in `IllegalFeatureCombinations`, each rejecting a valid feature set.

## The defects

**Bracketed disjunction was inverted.** `[A | B]` routed through `addExclusivityRequirement`, which merged the disjuncts into one flag set with every name `true` — "illegal when the feature and *all* alternatives are selected". The grammar means the feature is available while *any* alternative holds, so the illegal state is the feature with none of them.

**Nested disjunction emitted nothing.** `[A | B | C]` parses as nested ORs and took a two-`Name` branch that silently added no rule, dropping the constraint on `CameraAvSettingsUserLevelManagement.MechanicalPresets` entirely. Both shapes now flatten through `extractDisjunctFeatures`.

**Choice sets discarded their gating expression.** A gated set demanded a selection even where the gate admits no member, and never enforced the gate itself. Members now carry a participation gate; the requirement holds only where a member remains, and the gate contributes its own exclusion.

**Otherwise-list entries all applied unconditionally**, though the first *applicable* entry governs. Entries are now evaluated in order, conditioned on the inapplicability of those preceding them (a disjunction of flag sets), and a feature is disallowed where no entry applies. This replaces the `optionalIf` special case, which qualified one entry by the others irrespective of position.

**Provisional choice sets were enforced.** A set whose members are all provisional required a selection, forcing adoption of conformance the specification has yet to settle. Such a set now requires nothing; mutual exclusivity still holds. `Conformance.isProvisional` makes the test available beside `isMandatory` and `isDisallowed`.

## Effect

`DeviceEnergyManagement` now satisfies the four rules its specification states in prose (§9.2.4):

| selection | before | after |
| --- | --- | --- |
| `[]` | at least one of PFR, SFR or PA | unchanged |
| `[PA]` | ❌ at least one of PFR or SFR | ✅ accepted |
| `[STA]` | ❌ PFR is mandatory | at least one of PFR, SFR or PA |
| `[STA, SFR]` | ❌ PFR is mandatory | ✅ accepted |
| `[PA, SFR]` | ✅ accepted | ❌ cannot be selected together |
| `[PA, STA]` | PFR is mandatory | unchanged |

`CameraAvStreamManagement` with video, snapshot and image control mounts. `ClosureDimension` with only motion latching mounts. `Groupcast` and `AmbientContextSensing` instantiate without features.

## Scope

Every standard cluster's illegal combinations were compared before and after. Seven change, each intended: `BooleanStateConfiguration`, `CameraAvSettingsUserLevelManagement`, `CameraAvStreamManagement`, `ClosureDimension`, `DeviceEnergyManagement`, `Groupcast`, `AmbientContextSensing`. No cluster becomes unanalyzable.

`requiresFeatures` turns false for `Groupcast` and `AmbientContextSensing`, so their generated documentation no longer claims a selection is required — the three regenerated files come from `npm run generate-endpoints`, not hand edits.

## Tests

528 → 549. Each production hunk was verified by reverting it and confirming the corresponding tests fail.

`build --clean`, `format-verify`, `lint` and the full suite pass.

## Review

Two independent adversarial passes and one maintainer-profile pass over the cumulative diff. Findings taken: a gated choice inside an otherwise list produced a false positive (now refused rather than mis-analyzed); the traversal context bundled; the boolean algebra hoisted out of the tree walk; `NotImplementedError` in place of bare `InternalError`; the choice-member probe isolated from live state.

Known limitation, deliberately refused rather than approximated: a gated choice set whose members close under differing conditions, or that an earlier otherwise entry makes conditional, cannot be expressed as one flag set per member. Both throw `NotImplementedError`, which `FeatureSelectionErrors` logs and abstains on. No specification cluster produces either shape.

Fixes #4215 — the bracketed-disjunction inversion and the dropped nested-disjunction rule. The three further defects were found in the same code path while fixing it and are addressed here as well.

Unrelated, included by request: `CNET/4.11` joins the excluded Wifi commissioning tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,11 @@ The main work (all changes without a GitHub username in brackets in the below li
     - Fix: A cluster's feature table no longer selects features; a feature the specification makes unconditionally mandatory is always selected
     - Fix: Feature selection records against `operationalIsSupported` so a feature's `default` conveys only the specification's fallback value
     - Fix: A feature mandated by any of several alternatives, such as `DoorLock.User`, is now reported as required
+    - Fix: A feature conformance that is a bracketed disjunction, such as `[VDO | SNP]`, is illegal only when none of its alternatives is selected
+    - Fix: A choice set whose members are gated on a conformance expression, such as `[!PA].a` or `[PS].b`, requires a selection only where the gate admits a member, and rejects a member the gate excludes
+    - Fix: An entry of an otherwise conformance applies only where the entries preceding it do not, and a feature is disallowed where no entry applies
+    - Fix: A choice set whose members are all provisional, such as the `Groupcast` and `AmbientContextSensing` feature sets, no longer requires a selection
+    - Enhancement: New `Conformance.isProvisional` tells whether conformance describes what an element requires only once it leaves provisional state
 
 - @matter/node
     - Breaking: Default server exports no longer inherit the features their base implementation enables internally.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,7 +29,6 @@ The main work (all changes without a GitHub username in brackets in the below li
     - Fix: A choice set whose members are gated on a conformance expression, such as `[!PA].a` or `[PS].b`, requires a selection only where the gate admits a member, and rejects a member the gate excludes
     - Fix: An entry of an otherwise conformance applies only where the entries preceding it do not, and a feature is disallowed where no entry applies
     - Fix: A choice set whose members are all provisional, such as the `Groupcast` and `AmbientContextSensing` feature sets, no longer requires a selection
-    - Enhancement: New `Conformance.isProvisional` tells whether conformance describes what an element requires only once it leaves provisional state
 
 - @matter/node
     - Breaking: Default server exports no longer inherit the features their base implementation enables internally.
--- a/packages/model/src/aspects/Conformance.ts
+++ b/packages/model/src/aspects/Conformance.ts
@@ -100,30 +100,6 @@ export class Conformance extends Aspect<Conformance.Definition> {
     }
 
     /**
-     * Is the associated element provisional?
-     *
-     * Conformance following a provisional flag states what the element requires once it leaves provisional state, so it
-     * does not bind an implementation today.
-     */
-    get isProvisional() {
-        const conformance = this.ast;
-        if (conformance.type === Conformance.Flag.Provisional) {
-            return true;
-        }
-        if (conformance.type === Conformance.Special.Otherwise) {
-            for (const c of conformance.param) {
-                if (c.type === Conformance.Flag.Provisional) {
-                    return true;
-                }
-                if (c.type === Conformance.Flag.Mandatory) {
-                    return false;
-                }
-            }
-        }
-        return false;
-    }
-
-    /**
      * Is the associated element disallowed?
      */
     get isDisallowed() {
--- a/packages/model/src/aspects/Conformance.ts
+++ b/packages/model/src/aspects/Conformance.ts
@@ -100,6 +100,30 @@ export class Conformance extends Aspect<Conformance.Definition> {
     }
 
     /**
+     * Is the associated element provisional?
+     *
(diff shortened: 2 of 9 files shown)


Apollon77:
All three were correct — reproduced each before changing anything. Fixed in 16ca5b2b.

**Contradictory merge.** Real bug. `[FOO], FOO` emitted `{FOO: true, BAZ: false}`, a requirement drawn from a state where `FOO` is at once true and false, because the merge used object spread rather than the `conjoin` written for exactly this. Both the entry merge and the exhaustion rule now conjoin, so the output is `{FOO: false, BAZ: true}` alone.

**Choice reached conditionally with an empty gate.** Real gap. My earlier reasoning about this shape only held because the preceding entry was a bare feature name, which makes the members mandatory and the choice rule vacuous where that entry governs. With `[FOO], O.a` the preceding entry leaves them optional, so the set bound in states it does not govern. The registration now requires either that the choice is reached unconditionally or that the preceding entries already mandate the feature, and refuses otherwise. `Thermostat`'s `AUTO, O.a+` is the mandating case and is unaffected.

**Positional provisional.** Correct, and the fix removed API rather than adding it. Whether a choice follows a `P` is a property of its position in the otherwise list, which the walk already knows, so the member carries that fact instead of consulting a whole-conformance getter. `[FOO].a, P, O` now keeps its requirement, guarded by `FOO`. That left `Conformance.isProvisional` without a caller, so it is gone and `aspects/Conformance.ts` is back to its state on main.

None of the three affects a cluster in the specification — every standard cluster's illegal combinations and `requiresFeatures` are unchanged by this commit. Three regression tests added, each verified by reverting its fix and confirming it fails.

Apollon77:
Correct, and the specification is explicit — § 7.3.14: "It is invalid to use mandatory conformance for choice such as `M.a`, `AB.a+` or `!AB.a-`." Fixed in 3be8a582.

`participationGate` assumed any exclusion it received disallows the feature, which holds for `O`, `[FOO]` and `[!PA]` but not for a name or `M` expression, where the exclusion requires it instead. Stripping the feature from `{AB: true, X: false}` left `{AB: true}`, read as a removal gate with the polarity reversed:

```
AB.a  before  {X: true, AB: true}            X disallowed where it is in fact mandatory
              {X: false, Y: false, AB: false} the set required where it does not apply
M.a   before  {X: false, Y: false}            mandatory conformance stripped to an empty gate
```

The gate now requires an exclusion that disallows the feature and refuses anything else, so both shapes raise `NotImplementedError` and `FeatureSelectionErrors` abstains rather than emitting inverted rules. The file already asserted this in a comment without enforcing it.

`O.a`, `[PS].b` and `[!PA].a` are unaffected, as is every cluster in the specification.

PR #4244, opened 2026-08-14
fix: keep client storage metadata out of cluster state, and log shell attribute changes per attribute

Apollon77:
Two fixes that showed up together in shell logs of a controller talking to real devices.

## A client node's data version leaked into cluster state

A client node caches a peer's cluster values together with the data version the peer reported for them. That version travels alongside the values under the key `__version__`, and the datasource — which only knows the members its schema defines — treated it as one more attribute. It ended up in cluster state, and any report carrying a new version counted as a changed property.

Anything listening for changes therefore saw one whenever the peer bumped the version, even when no value differed at all. In the shell that reads:

```
Node 67: Attribute 1/1026 changed to {"__version__":"undefined"}
```

The peer had repeated an unchanged measurement; the only "change" was the version, and its value prints as undefined because state does not expose the key.

The fix treats `__`-prefixed keys as storage metadata wherever the datasource takes values from a store: they are no longer seeded into state when values are loaded (which already held for the feature key), and no longer considered when a report is integrated. The version still reaches the datasource through the store's own version property, so a report that carries nothing else updates it and emits no change.

One follow-through in the cache: it reads the values it persists back from the datasource, so a flush whose only pending key was the version would find nothing there and re-queue itself indefinitely. It now writes the version from its own record in that case.

## Shell logged one line per cluster report, labelled as one attribute

Peers report attribute changes per cluster, and the shell logged one line per report — labelled `Attribute <endpoint>/<cluster>`, with an object holding every attribute that changed. It reads as a single attribute while showing several, and names none of them:

```
Node 11: Attribute 1/145 changed to {"periodicEnergyImported":{...},"periodicEnergyExported":{...}}
```

Now one line per attribute, naming node, endpoint, cluster and attribute, as the shell did before the controller API changed:

```
Node 11: Attribute 11/1/145/periodicEnergyImported changed to {...}
```

That format is also what the bundled web UI looks for, and it had been unable to parse attribute changes since the change, so device tiles stopped showing values. Its pattern for connection state was stale in the same way — it still expected the prefix of a callback that no longer exists — so node tiles never updated and the subscription the UI starts when a node connects never ran. Both now match the text the shell actually logs.

Event lines additionally carry the peer's node id, spell out the priority instead of printing its number, and render the timestamp as a date, or as a formatted duration for a device that timestamps against its own uptime. A bare number gave no indication which of the two it was.

## Tests

Four datasource tests covering metadata seeding, changed-property reporting, a version-only report and a report repeating a value with a new version, plus one cache test for the version-only flush. Every production change was verified by reverting it and confirming the tests fail.

`npm run build`, `npm run format-verify`, `npm run lint` and the full `npm test` all pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -72,6 +72,7 @@ The main work (all changes without a GitHub username in brackets in the below li
     - Fix: Writing a fabric-scoped list entry that stems from a cluster whose schema could not be resolved no longer produces two conflicting fabricIndex fields
     - Fix: A rejected write to an attribute served by dynamic properties restores the previous value instead of deleting the property, and a rejected write to a previously absent attribute leaves no slot behind instead of one holding `undefined`
     - Fix: Factory reset removes commissioned peers and the certificate authority's key material; a peer that cannot be torn down no longer blocks the reset
+    - Fix: A client node's storage metadata no longer surfaces as state: a peer report that only bumps the data version emits no change notification, and `__version__` no longer appears among the changed properties or in cluster state
 
 - @matter/nodejs
     - Breaking: `FileStorageDriver`'s constructor no longer accepts a `clear` argument; clearing is handled by `StorageService`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -65,7 +65,7 @@ The main work (all changes without a GitHub username in brackets in the below li
     - Fix: Struct validation resolves a member stored under its TLV tag number, so a constraint violation there is caught and a mandatory member present only at its id no longer raises a conformance error
     - Fix: `endpoints.size` no longer double-counts the root endpoint
     - Fix: A commissioned peer's connection state leaves `Connected` as soon as its last operational session is lost
-    - Fix: `ChangeNotificationService` event occurrences carry a `timestampKind` (epoch vs system) so a consumer forwarding an event no longer has to guess which clock its timestamp came from
+    - Fix: `ChangeNotificationService` event occurrences carry a `timestampKind` naming which of the four wire variants the timestamp is (`epoch`, `system`, `epoch-delta`, `system-delta`), so a consumer forwarding an event no longer has to guess its clock or whether it is absolute or a delta from the previous event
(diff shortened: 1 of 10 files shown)


Apollon77:
Reviewed the three comments; two applied, one declined.

**Timestamp variants — correct, fixed properly.** The finding holds: `InputChunk.ts:389` collapses all four wire variants into `timestamp` without rebasing, and the previous classification folded each delta in with its absolute counterpart, so an epoch delta rendered as a 1970 date and a system delta as an absolute device uptime.

The shell could not fix this on its own, since the delta-ness never reached it, so `ChangeNotificationService.TimestampKind` now names the variant itself — `epoch`, `system`, `epoch-delta`, `system-delta`. That type shipped days ago and is unreleased, so no consumer breaks. The shell formats each accordingly, and a new test drives all four through the client event emitter; reverting the classification makes it fail with `[ 'epoch', 'system', 'epoch', 'system' ]`. Downstream consumers such as matterjs-server need the same distinction to choose between `epochTimestamp` and `systemTimestamp` when republishing.

**Cast in the cache test — applied.** The mock does satisfy the consumer interface, so it is now `satisfies Datasource.ExternallyMutableStore.Consumer` and a future contract change fails the test at compile time.

**Comment on the connection-state regex — keeping it.** It labels which log line that regex targets, and both siblings in the same block carry exactly that (`// attribute changed`, `// from "read attribute"`). Dropping only the new one makes a wall of near-identical regexes harder to read, and the guidance quoted is about narrating code rather than labelling which line a pattern is for.

Apollon77:
Valid, and it goes further than the formatting. Fixed in 30f17655f.

`TlvEventData.epochTimestamp` is `TlvPosixMs`, which is `TlvUInt64`, so a peer can send a time far beyond the roughly ±8.64e15 ms a JavaScript date spans; `InputChunk` passes it through as a plain number and `toISOString()` then throws `RangeError`. The epoch branch now checks the constructed date and prints the raw offset when it is not renderable.

The blast radius was wider than the log line. `Observable`'s default error handler rethrows from inside the emit loop, so the throw would have skipped every remaining observer of that change and propagated back through `broadcastEvent` into the peer's report processing — one unrenderable timestamp taking out unrelated consumers and the interaction that produced it. Logging a change is therefore guarded as a whole now, with the failure logged, so no future formatting bug in a diagnostic can fail an interaction.

Not covered by a test: `@matter/nodejs-shell` has no test setup, and I did not want to stand one up inside this PR.

Issue #4374, opened 2026-08-30
CommodityTariff.DayEntryStruct.RandomizationType has both default: 0 and conformance [RNDM], so setStateOf() rejects its own auto-fill


lboue:
### matter.js version

0.17.9

### Node.js version

v24.18.0

### Operating system

Ubuntu Linux 6.17.0-35-generic, x86_64

### Description

**What I did:**
Built a `CommodityTariff` cluster server (feature `Pricing` only, `Randomization` NOT enabled) on an
endpoint, then used `endpoint.setStateOf(CommodityTariffServer, { dayEntries: [...], ... })` (via
Matterbridge's `setAttribute()` helper) to publish a flat-rate day schedule after the endpoint was
already installed/active. The `dayEntries` value was:

```js
dayEntries: [{ dayEntryId: 1, startTime: 0 }]
```

i.e. a plain `DayEntryStruct` with only the two mandatory fields (`DayEntryId`, `StartTime`) — no
`RandomizationOffset`/`RandomizationType` key at all.

**What I expected:**
The write to succeed, since `RandomizationType`'s conformance (`[RNDM]`) means the field must be
*absent* when the `Randomization` feature isn't enabled — and I never supplied it.

**What happened:**
The write is rejected:

```
Validating Matterbridge.Matterbridge.electricalMeter1.electricalMeterCurrent.commodityTariff.state.0.randomizationType:
Conformance "[RNDM]": Matter does not allow you to set this attribute (135)
```

Full stack trace (from the attached log, `matterbridge-commoditytariff-bug.log`):

```
Error: Validating Matterbridge.Matterbridge.electricalMeter1.electricalMeterCurrent.commodityTariff.state.0.randomizationType: Conformance "[RNDM]": Matter does not allow you to set this attribute (135)
    at disallowValue (file:///.../node_modules/@matter/node/dist/esm/behavior/state/validation/conformance-compiler.js:432:13)
    at Object.randomizationType (file:///.../node_modules/@matter/node/dist/esm/behavior/state/validation/conformance.js:20:19)
    at validateStruct (file:///.../node_modules/@matter/node/dist/esm/behavior/state/validation/ValueValidator.js:275:23)
    at file:///.../node_modules/@matter/node/dist/esm/behavior/state/validation/ValueValidator.js:30:7
    at file:///.../node_modules/@matter/node/dist/esm/behavior/state/validation/conformance.js:28:22
[...]

PR #4375, opened 2026-08-30
fix(model): remove conflicting default on CommodityTariff RandomizationType

lboue:
## Type of change

## Description

`CommodityTariff.DayEntryStruct.RandomizationType` declared both conformance `"[RNDM]"` (only allowed
when the `Randomization` feature is supported) and `default: 0`. Writing a `dayEntries` list entry that
omits `RandomizationType` auto-filled it with that default before validating conformance, then rejected
the very same fill:

```
```

This only happens along the list-of-struct write path (`Endpoint.setStateOf` → `supervisor.patch` →
`StructManager`) — a plain state property assignment (`agent.get(...).state.dayEntries = [...]`) or a
bare, non-list struct attribute with the identical field shape (e.g. `CurrentDayEntry`) don't trip it.

Per the spec text for both `RandomizationOffset` and `RandomizationType` (Application Cluster
Specification § 9.12.5.10.4–5, as transcribed in `commodity-tariff.resource.ts`): *"If this field is not
indicated, randomization shall use the value in the Default\* attribute."* Absence is spec-meaningful
(fall back to the cluster's own `DefaultRandomizationType`), not "value is `None`". `RandomizationOffset`
already has no `default` in the model, matching that — `RandomizationType`'s `default: 0` was
inconsistent with its sibling field, and collapsed that fallback semantic even when `Randomization` was
enabled, on top of causing the conformance crash when it wasn't.

**Fix**: remove the conflicting `default: 0`.

## Backing evidence

- Related issue: #4374
- [x] This fix/change is backed by a **complete log file** — attached here or in the linked issue (not just a few quoted lines).

(Log attached on #4374, from the real-world crash in a Matterbridge plugin. This PR also adds a
self-contained regression test — `CommodityTariffServerTest.ts` — that reproduces the exact same error
via `MockServerNode` + `Endpoint.setStateOf()`, with and without the fix.)

## Checklist

- [x] `npm test` passes — ran `@matter/model`'s build, `@matter/node`'s full `test/behaviors/**` suite (390/390, including the new regression test) and `@matter/general`'s suite (1330/1330). Could not complete a full root `npm test` in this sandbox: the browser/web test step fails on a missing Playwright browser binary, unrelated to this change.
- [x] `npm run format-verify` and `npm run lint` pass (both run at the repo root, not just on the touched files)
- [x] CHANGELOG updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,9 @@ The main work (all changes without a GitHub username in brackets in the below li
     - Fix: A certification run's `result.json` no longer reports a passing verdict for a run that failed
     - Fix: A failure to attach a certification run's device logs fails the run instead of only warning
 
+- @matter/model
+    - Fix: `CommodityTariff.DayEntryStruct.RandomizationType` no longer declares a `default` value that conflicted with its own `[RNDM]` conformance gate; a `dayEntries` list entry that omits `RandomizationType` when `Randomization` is unsupported no longer fails validation for a value the write never asked to set
+
 - @matter/node
     - Fix: Validation honors the conformance and quality an element inherits, so an element overridden by an operational extension is no longer judged as if it stated neither
     - Fix: A write from local code that adds an entry to a fabric-scoped list without a `fabricIndex` now fails validation instead of storing an entry that belongs to no fabric. Writes from a peer are unaffected — the accessing fabric is supplied for them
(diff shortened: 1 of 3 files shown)


Apollon77:
nooo, we never change the model files directly!

And in fact the default: "None" is basically defined by spec ... so have you more details like exact code to check that and verify? But the fix - if we change the default at all is a model override but this more looks like some other root cause. I close this PR and will check the issue later

PR #4380, opened 2026-08-31
fix(node): filter getDefaults() to conformant members

lboue:
## Type of change

## Description

Supersedes #4375, which was correctly closed: that first pass hand-edited the generated
`commodity-tariff.element.ts` to drop `RandomizationType`'s `default: 0`/`"None"` — but that default is
spec-mandated (Application Cluster Specification § 9.12.5.10, `DayEntryStruct` field table: `RandomizationType`'s
Fallback column is `None`), so removing it "fixed" the crash by breaking the spec-correct read fallback for
when `Randomization` *is* supported and the field is omitted.

**Actual root cause**: `ValuePatcher.ts`'s `getDefaults()` calls `supervisor.membersOf(schema)` with no
conformance filter, so it considers every declared member's `default` regardless of whether that member is
even conformant given the currently active features. Writing a `dayEntries` list entry that omits
`RandomizationType` synthesizes its default (`None`) even when `Randomization` is unsupported — where the
field can't exist at all — then rejects that very same synthesized value for violating the conformance gate
it was never asked to satisfy:

```
```

`SelectDefaultValue()` (`@matter/model`) already guards this correctly with `scope.hasOperationalSupport(member)`;
`StructManager.ts`'s mandatory-default synthesis path already calls `membersOf(member, { conformance: "conformant" })`
for the same reason. `getDefaults()` has its own, separate member-default computation that skipped that
filter. This only surfaces on the list-of-struct write path — a bare, non-list struct attribute with the
identical field shape (e.g. `CurrentDayEntry`) doesn't trip it.

**Fix**: pass `{ conformance: "conformant" }` to `membersOf()` in `getDefaults()`, matching the filter already
used elsewhere. No model files touched; `commodity-tariff.element.ts`'s `default: 0` stays as originally
declared.

## Backing evidence

- Related issue: #4374
- [x] This fix/change is backed by a **complete log file** — attached here or in the linked issue (not just a few quoted lines).

(Log attached on #4374, from the real-world crash in a Matterbridge plugin.) Two tests in
`CommodityTariffServerTest.ts` cover both sides: `Randomization` unsupported (the crash — omitting
`RandomizationType` must be accepted) and `Randomization` supported (omitting it must still read back as the
spec's `None` fallback, not `undefined`) — both fail without this fix, both pass with it.

## Checklist

- [x] `npm test` passes — ran `@matter/model`'s build, `@matter/node`'s full `test/behaviors/**` suite (391 tests, including both new regression tests) and `@matter/general`'s suite (1330/1330).
- [x] `npm run format-verify` and `npm run lint` pass (both run at the repo root, not just on the touched files)
- [x] CHANGELOG updated


Apollon77:
Thank you for chasing this down — your second analysis is the one that got the root cause right, and it
saved us the detour: the model's `default: 0` is a faithful transcription of the spec's Fallback column,
and `getDefaults()` in `ValuePatcher.ts` is where the value is invented. We picked that up and took it a
step further, so I'm going to close this in favour of #4386 — not because the diagnosis was wrong, but
because the filter it lands on stops one case short.

`{ conformance: "conformant" }` removes members whose conformance evaluates to *disallowed*. It keeps
members that are merely *optional*, and `[RNDM]` is optional whenever the `Randomization` feature is
supported. So on a cluster with `Randomization` enabled, a day entry that omits `RandomizationType` still
gets `None` written into it — and that is not a harmless fill:

> § 9.12.5.10.5: "If this field is not indicated, randomization SHALL use the value in the
> DefaultRandomizationType attribute."

Absent means "use `DefaultRandomizationType`". An explicit `None` means "apply no randomization". Writing
the fallback in turns the first into the second, silently, for a device that never asked for it. The test
in this PR that asserts `randomizationType === None` after a write that omitted the field pins exactly that
behaviour, which is why we could not just merge it as-is.

More generally, the core specification defines the Fallback column as a reader-side value:

> § 7.18.4: "A fallback value is defined to complete dependencies when the actual data field value is not
> present."

That is what a consumer assumes for an absent field, not something a server must materialize. And when a
field is optional, whether it is present is the implementation's decision, not ours.

So #4386 gates the default on whether the cluster actually supports the member — mandatory given the
active features, or explicitly implemented — rather than on conformance alone. It keeps your two real
findings verbatim: the model file stays untouched, and the defaults cache is keyed by supervisor for the
reason you identified (peer supervisors share nested schema identity). It also fixes two neighbours the
same loop was causing: struct defaults were stored in the schema's notation rather than the datatype's
units (a Thermostat preset without a cooling setpoint got `{type: "celsius", value: 26}` instead of `2600`),
and the same `[FEATURE]`-gated fill was breaking writes in `ClosureControl`, `ClosureDimension` and
`ElectricalEnergyMeasurement`.

Your issue report and the two rounds of analysis on #4374 are what made this findable — thank you. Closing
here; please do follow #4386 and push back if you disagree with the reasoning above.

PR #4386, opened 2026-09-01
fix(node): a struct write fills in only a default the cluster supports

### LINKED ISSUE #4374: CommodityTariff.DayEntryStruct.RandomizationType has both default: 0 and conformance [RNDM], so setStateOf() rejects its own auto-fill

Apollon77:
## Type of change

fixes #4374

## Description

Writing a struct — a struct-typed attribute, or an entry of a list of structs — starts from a set of
default values taken from the schema and then applies what the caller supplied. That starting set was
built from every field the schema declares, whether or not the cluster supports the field.

**A field the cluster cannot have was filled in, then rejected for existing.** `CommodityTariff`'s
`DayEntryStruct` declares `RandomizationType` under conformance `[RNDM]`, so the field may only appear
when the `Randomization` feature is enabled. The schema also states a fallback value for it. Writing a
day entry that omitted the field on a cluster without `Randomization` filled the fallback in and then
failed validation on the value it had just invented:

```
```

The same shape exists in `ClosureControl`, `ClosureDimension` and `ElectricalEnergyMeasurement`, whose
feature-gated struct fields are nullable, so the write filled in `null` and hit the same wall.

**An optional field nobody set was reported as if the device had set it.** A fallback is not a value a
device must publish. The Matter core specification (§ 7.18.4, "Fallback Column") defines it as the value
that "completes dependencies when the actual data field value is not present" — what a *reader* assumes
for an absent field. Filling it in on the writer's behalf can invert the meaning of the data: for
`RandomizationType`, an absent field means "use the cluster's `DefaultRandomizationType` attribute"
(Application Clusters § 9.12.5.10.5), while an explicit `None` means "apply no randomization at all". A
device that never asked for randomization behaviour was made to announce a specific one. Where a field
is optional, whether it appears is the implementation's decision, not ours.

A field now contributes a default only where the cluster actually supports it: mandatory given the
active features, or explicitly implemented. An optional field the implementation does not set stays
absent, which is what tells a reader to apply the fallback.

Three further defects in the same function, found while fixing that one:

- **Defaults were stored in the schema's notation rather than the datatype's units.** A `Thermostat`
  preset created without a cooling setpoint received the literal `{ type: "celsius", value: 26 }` the
  schema states, not `2600`, and a bitmap default arrived as a raw number rather than decoded. Defaults
  now decode through the model's `DefaultValue()`, as every other default path already does.
- **Every struct created from one schema shared one instance of each container default.** The cached
  default set was handed out with a one-level spread, so two entries of a list of structs received *the
  same* array for a list-typed member, and the patcher then wrote into it — leaving both entries holding
  the second one's content and polluting the cached default for the rest of the process. Each struct now
  gets its own copy.
- **A patch key that names a property of `Object.prototype` passed the member check.** Membership was
  tested with `key in memberPatchers` against a plain object literal, so `toString` was accepted as a
  member and stored as `"[object Object]"`, and `__proto__` escaped as an untyped `TypeError` instead of
  the documented `WriteError`. Reachable from `Endpoint.set` and from the remote API's
  `PropertyResource.patch`; the Matter protocol write path does not use this code. Those maps now carry
  no prototype.

**The default set was also cached against the schema alone.** It now depends on which features are
active, and two feature variants of one cluster share the identity of their nested struct schemas, so
whichever variant was written first served its defaults to the other. The cache is keyed by supervisor
first.

Supersedes #4380, which fixed the reported crash by filtering the same loop to conformant members. That
filter keeps optional members, so a cluster *with* `Randomization` still had `None` written into every day
entry that omitted the field — and #4380's second test asserted that as the expected result.

### Observable change

An optional struct field that a device does not set is now absent where it previously carried the schema's
fallback. In the standard clusters this affects `Thermostat` `PresetStruct.Name`, `Descriptor`
`SemanticTagStruct.Label`, `MediaPlayback` track attributes, and the feature-gated fields named above —
all cases where the specification's own fallback for the absent field is the value that was being written.

## Backing evidence

- Related issue: #4374 (complete log attached there, from a Matterbridge plugin)

`ValuePatcherTest.ts` covers each branch: a field the features disallow, a field they leave optional, a
field they make mandatory, container-default copying, unit decoding, the prototype-chain key, and
per-feature-set caching. `CommodityTariffServerTest.ts` reproduces the reported
`setStateOf({ dayEntries: [...] })` write in both feature states. Every new assertion was checked against
the unmodified baseline and fails there.

Gates run from the repository root on the final tree: `npm run build-clean`, `npm run format-verify`,
`npm run lint` and `npm test`, all passing.

## Checklist

- [x] `npm run build-clean` passes from the repository root
- [x] `npm test` passes
- [x] `npm run format-verify` and `npm run lint` pass
- [x] CHANGELOG updated


Apollon77:
Four further commits, addressing the review comments and a profile-driven review pass.

**Nullable member that a feature expression makes mandatory** — correct, and the fix is a layer down. Neither `member.mandatory` (unconditional `M` only) nor `scope.isMandatory` was usable as written, because `Conformance.computeApplicability` reported a term that compares a value as *Optional*, and `AND` then read that as a settled answer: `TriggerType == Motion & !PERZONESENS` claimed *Mandatory* whenever `PERZONESENS` is unsupported, whatever the record holds. `NOT` had no case for `Optional` at all, so `!(Field == Value)` fell through to the `InternalError` at the end of the evaluation. Such a term is now *Conditional*, which both operators already handle and which says what is true — the record's contents decide. With that, `scope.isMandatory` means what it says and the patcher uses it, so `CommodityTariff`'s `Currency` under `Pricing` holds `null` when a write omits it. The elements whose applicability changes are the revision-gated ones (`BasicInformation.SpecificationVersion`, `MaxPathsPerInvoke`, `ConfigurationVersion`, `GeneralDiagnostics.DeviceLoadStatus`, `Chime.ChimeStartedPlaying`); element validation resolves a conditional element as present exactly when it is implemented, which is what they already did as optional elements, and there is now a test for that.

**`deepCopy` and `Date`** — correct; fixed in `@matter/general` with a test. No shipped cluster has a struct member with a date, duration or bytes default, so nothing reached it through this PR's call sites, but `StructManager` copies schema defaults on the read path and had the same exposure. The wider gap — `RegExp`, `URL`, `Error` and other exotic objects degrade to a plain object, and `DataView` throws — is left alone here as a behaviour change for every caller.

**Build gate** — `npm run build-clean`, `npm run format-verify`, `npm run lint` and `npm test` all run from the repository root on each pushed state, and all pass.

A separate review pass also produced these, now included:

- The stored-default policy moved to `@matter/model` as `StoredDefaultValue()`, beside `SelectDefaultValue()` and `MandatoryDefaultValue()` and documented against them, rather than being a third copy of the same question in the node package. It omits the datatype-level fallbacks deliberately: a mandatory member a write leaves out must fail validation naming the member rather than read as a silent zero, and the read path synthesizes the specification's fallback regardless.
- `__proto__`, not `toString`, is the key that mattered in the prototype-chain fix — it passed the member check and reassigned the prototype of the value being written. Named in the CHANGELOG and covered by a test that writes it as a computed key, since `{ __proto__: x }` in a literal is prototype syntax rather than a property.
- The CHANGELOG claimed a bitmap default arrives decoded, and no shipped cluster states a default on a bitmap struct member, so nothing exercised it. The synthetic fixture now has one.
- One line of documentation per `Applicability` member, `Conditional` stated as "features alone do not decide".

Two items are deliberately left for their own change: `deepCopy`'s general exotic-object handling, and moving the defaults cache onto `ValueSupervisor`, which already caches on the same key.

PR #4469, opened 2026-09-18
Stop a rebooting device's farewell report from cancelling fast re-subscription

Apollon77:
## The failure

`RebootResubscribeArmer` decides, 30 seconds after a rebooted peer's new session appears, whether the peer's subscription survived the reboot. If it did not, the controller forces re-subscription instead of waiting out the full subscription liveness timeout.

It made that decision by comparing the peer's last report timestamp against the returning session's creation time. A rebooting device flushes its dying subscription as it shuts down, and that flush can be stamped in the same millisecond as the returning session. From a CI run of the OTA reboot test:

```
13:08:59.600 ServerSubscription   Flushing subscription 02020202 (for closing)
13:08:59.600 ClientEventEmitter   Received event basicInformation.shutDown on controller1.@1:1
13:08:59.600 SecureSession        Created secure CASE session ... @1:1•0103
13:09:00.000 …notifyUpdateApplied Success
13:09:31.000 Node controllerA going offline      ← teardown, nothing in between
```

The flush report and the returning session carry the same timestamp, so `lastReportStartedAt >= session.createdAt` read it as "the subscription is being fed", the forced re-subscription was skipped, and the controller waited out the full timeout. A real device hits the same tie on a real clock.

## The change

A timestamp cannot carry the one thing the decision needs: which session the report arrived over. The armer now observes the report itself, and judges it by session identity.

- `ClientSubscriptions.reportStarted` is an observable carrying the session a pushed report arrives over, emitted from the point in `ClientSubscriptionHandler` that already noted an inbound report. A session names its own peer, so that is the whole payload. It has an error handler, because it is emitted while a report is being read and an observer that throws must not abort the report.
- The armer's arm state names the sessions that count as evidence rather than the ones that do not. The first session after arming is the return; later sessions with the peer join it. A report over one of them marks the subscription live, and the grace window keeps it.
- A device opens sessions for its own reasons, so a later one is not a second reboot: it must not discard evidence already gathered or restart the grace window.

`ClientSubscriptions.lastReportStartedAtFor()` and `PeerSubscription.lastReportStartedAt` had no remaining reader once the decision stopped using timestamps, and are removed. That is the Breaking part; `reportStarted` replaces them.

## Not changed

These predate this change, and the timestamp version shared all but the last. They are listed so the scope is clear, not fixed here:

- A report arriving before the peer returns does not count as evidence. That includes a session the controller opened between arming and the return: it joins no return set, so a report over it is not evidence until the peer has announced itself. The cost is one unnecessary re-subscribe in that interleaving, which is the safe direction — the alternative counts pre-reboot traffic, which is the defect this PR fixes.
- A peer that reboots a second time inside the grace window keeps the evidence of its first return, so its subscription is left in place and recovers on the subscription's own liveness timeout. Telling a second reboot apart from a second session needs a reason for the session that the session cannot give, and the fallback is what happens without this component at all. A characterization test records it. Reboot followed by one session, or none, is the case this is built for.
- Evidence is peer-scoped rather than per-subscription.
- The grace window restarts on each return.
- `SoftwareUpdateManager[Symbol.asyncDispose]` leaves `internal.rebootResubscribeArmer` set. Not reachable in the current lifecycle.

## Tests

21 tests in `RebootResubscribeArmerTest` (was 13), 4 in `ClientSubscriptionsTest` (was 1), 26 OTA tests.

Each added branch was mutated and the mutation run against the suite:

| mutation | test that fails |
|---|---|
| restore the old `>=` against the returning session's `createdAt` | re-subscribes when the only report arrives over a pre-reboot session |
| any session counts, not only the current return's | that test, plus the second-reboot and re-arm cases |
| a later session restarts the return | keeps the subscription when the peer opens a second session after reporting |
| a later session does not join the return | the two "keeps the subscription…" cases |
| reports before the peer returns count as evidence | re-subscribes when the only report arrives before the peer returns |
| `arm()` does not drop the previous cycle's state | re-arming requires fresh evidence for the new cycle |
| re-arm leaves the previous return deadline running | re-arming replaces the previous return deadline rather than leaving it to fire |
| the armer registers no report observer | three "keeps the subscription…" cases |
| the `reportStarted` error handler rethrows | contains an observer that throws |
| `close()` does not dispose the observable | drops observers once closed |

## Verification

```
=== npm run build-clean ===
  ✓ Type check
  ✓ Transpile

=== npm run format-verify ===
All matched files use the correct format.
Finished in 53ms on 2396 files using 14 threads.

=== npm run lint ===
> oxlint --type-aware --tsconfig tsconfig.json examples packages support
(no findings)

=== npm test ===
(full suite, no failures)
```


Apollon77:
All three addressed in 0226c16b8.

**Observer interception** — confirmed against the contract at `Observable.ts:22-29` and the emit loop: a non-`undefined` return ends emission and a promise delays it, and the armer's listener is registered when a peer is first armed, so it sits behind anything registered earlier. A concise arrow returns whatever its body evaluates to, so this needs no intent to trigger. This is the third finding on the same construct (earlier rounds raised a throwing listener and a dropped promise), so rather than add a third guard I removed the observable: `ClientSubscriptions.onReport()` keeps its own listeners and calls each for effect, so a return value stays with its listener and a thrown error is logged without reaching the others or the report being read. A test pins that a listener returning a value does not stop later ones.

**Class documentation** — correct, the rule is "any session the peer did not already hold when it was armed", which includes a controller-initiated session opened after the reboot. Reworded.

**Test cast** — correct, the `{} as SecureSession` was mine and is gone; the tests build a real session through `SessionManager`. The remaining `as unknown as PeerSubscription` in that file predates this PR.

Apollon77:
Both correct, addressed in ce6c45888.

**Stale session from an earlier return** — confirmed: `preRebootSessions` was only ever assigned in `arm()`, so a session opened by a first return stayed outside it, and after a second return a delayed farewell report over that session still set the new cycle's `fed`. This is the third round of findings on the same field, so rather than widen the exclusion set I inverted it: the return record now lists the sessions that count — the session that announced the return, plus later sessions with the peer — and a further return starts a new set. A session from an earlier return is then not evidence for a later one without anything having to remember to exclude it, and `preRebootSessions` is gone. The second-reboot test now emits over the first return's session as you suggested.

**Async listener's discarded promise** — correct. `ReportListener` now returns `MaybePromise<void>` and a rejection is logged via `MaybePromise.catch`. The promise is still not awaited: waiting would let a listener delay the report it is only being told about, which is the property `onReport` exists to guarantee. Two tests cover it — later listeners run without an earlier asynchronous one settling, and a rejecting async listener is contained. Reverting the rejection handling makes the second fail with the rejection surfaced.

Apollon77:
Half right, and the half that is right was my PR text rather than the code.

**Correct:** `arm()` installs empty state, and a session the controller opens between arming and the return joins no return set, so a report over it is not evidence. The PR body still described an arm-time snapshot and a mutation case named "a session we opened after arming" — both were removed in ce6c45888, and I had not updated the text. Fixed: the body now describes the allowlist the code implements, and the corrected mutation table. The test is `keeps the subscription when a report arrives over a session we opened after the return`.

**Declining the code change.** Reinstating an arm-time session baseline is what ce6c45888 removed, and it is what your previous round correctly identified as broken: a baseline captured at `arm()` never contains a session opened by a *first* return, so after a second reboot a delayed farewell report over that session still counted as evidence. Four rounds of findings have now landed on this one field — the timestamp, the snapshot's timing, then its contents — each one a different way for pre-reboot traffic to qualify. Listing what counts makes that class unrepresentable; widening an exclusion set only moves it.

The behaviour you describe is therefore deliberate, and it is now stated in the PR body under "Not changed". Its cost is one unnecessary re-subscribe in that interleaving. The alternative is counting traffic from before the peer announced its return, which is the defect this PR exists to fix, so the trade runs the right way: a spurious re-subscribe is recoverable, a missed one leaves the controller blind for the full subscription timeout.

Apollon77:
Reverted to a plain `Observable` in c2ea7205b, and the earlier reasoning here was wrong in a way worth stating.

Nothing needs a value back from this notification, and `RebootResubscribeArmer` is its only listener. So the case behind the finding — some other listener returning a value and starving the armer — needs a second listener that does not exist, on a protocol-internal observable. That is a correct-but-theoretical finding, and the right answer was to say so rather than build a mechanism around it. The private listener set also cost more than it bought: a second notification idiom in a codebase that uses observables for this, no `once`, no async iteration, and a separate `Disposable` on the armer instead of its existing `ObserverGroup`.

What stays: `reportStarted` is an `Observable<[peer, session]>`, disposed in `close()` so observers do not outlive shutdown, and the armer registers it alongside its `sessions.added` observer.

The general question your finding surfaced — whether an observer's return value should stop later observers at all, given that TypeScript accepts a value-returning function where a `void` observer is expected — is now tracked as its own task against `@matter/general`. It is a property of `Observable`, not of this PR.

Apollon77:
This is the same finding as the earlier round, and the answer has not changed — it is a deliberate decision, not an oversight.

`reportStarted` is `Observable<[session: SecureSession]>` with no return value in its contract, and `RebootResubscribeArmer` is its only listener. The scenario needs a second listener that does not exist, registered on a notification internal to `@matter/protocol`, and written so it returns a value. That is a correct-but-unreachable finding, so the right response is to say so rather than build a mechanism around it.

We tried the mechanism. An earlier revision of this PR replaced the observable with a private listener set for exactly this reason (0226c16b8), and it was reverted in c2ea7205b: nothing here needs a value back, so the guarantee bought nothing, while the bespoke registry cost `once`, async iteration, `ObserverGroup` integration and a second notification idiom in a codebase that uses observables for this everywhere else.

The observable does carry an error handler now, matching `Peer.ts` for events raised from the wire path, so an observer that throws cannot abort the report being read. That one was reachable in the same sense and is covered by a test.

The general question your finding raises — whether an observer's return value should stop later observers at all, given that TypeScript accepts a value-returning function where a `void` observer is expected — is real and is tracked as its own task against `@matter/general`. It is a property of `Observable` shared by every observable in the codebase, and the place to fix it is there, not by working around it in one caller.

PR #4484, opened 2026-09-20
The Software Update certification cases whose DUT is the OTA provider

Apollon77:
Adds the first block of Software Update certification cases: TC-SU-3.1, TC-SU-3.2, TC-SU-3.3 and TC-SU-3.4, where the DUT answers a requestor's `QueryImage` and serves it an image, and TC-SU-1.1, where the DUT is the administrator that tells a requestor about a provider. The topology is the one TC-BDX-1.4 and TC-BDX-2.1 already use: the controller is the DUT, the device is the TH.

## What the cases needed

**A provider's own answer cannot be observed from outside it.** The requestor's log states what it received, and neither requestor this suite runs renders the update token's length or the image URI's exact text — both of which these cases assert on. `CertNodeApi.serveOtaUpdate()` therefore also reports `exchanges`: every `QueryImage`, `ApplyUpdateRequest` and `NotifyUpdateApplied` the controller's provider answered, with the fields of both sides. `OtaExchangeRecording` owns that record — it clears it and attaches its observer inside one `act`, so no answer falls between the two, and reading it copies, so an answer the requestor sends afterwards cannot reach a case still holding the earlier one.

**`CertNodeApi.announceOtaProvider()`** announces the controller, or another commissioned node, without staging an image. TC-SU-3.2 needs a provider with nothing to offer; TC-SU-1.1 needs the administrator's announcement of a third node.

**`certTest`'s `appArgs`** gives each device role its own app arguments, on the chip binary's command line and in a matter.js subject's `TestInstanceConfig`. The evidence bundle records them per role, so a reader can tell a flag that took effect from one that meant nothing on the flavor that ran.

**`CertNodeApi.scriptOtaProvider()`** queues answers the provider gives in place of its own, one per command, falling back to its real answer once a queue is spent. Five plan steps are about an answer the provider reaches only in a state nothing here can produce — a `Busy` carrying a `DelayedActionTime`, a `QueryImageResponse` carrying `UserConsentNeeded`, an `ApplyUpdateResponse` deferring or refusing the apply — and this is how they run. A scripted status or action is answered without asking the base implementation at all: its answer is a side effect as much as a value, staging an in-progress entry and registering the peer for BDX, and a status written over the top afterwards would leave the provider expecting a transfer the requestor was just told not to start.

## Timing

The delay such an answer names is the TH's to wait out, not the DUT's to be judged on, and the plans name three minutes of it. `OtaSoftwareUpdateRequestorServer`'s two-minute floors on re-querying a provider and on re-sending an `ApplyUpdateRequest` become `minimumQueryInterval` and `minimumApplyDelay`. `OtaRequestorTestInstance` lowers them under `MATTER_CERT_OTA_FAST_RETRY`, which the workflow sets on every leg, and the case then scripts a one-second delay and records in the check's own detail that it did. A product lowering these does not conform, which the JSDoc says.

chip's requestor floors the same waits at compile time. There the steps cost the plan's three minutes, so they declare `longRunning` and skip unless `MATTER_CERT_LONG_RUNNING` is set — which only the daily `schedule` trigger sets. `RunRecord.longRunningSkips` counts what a run left out, so a bundle without it covers the plan and one with it covers the plan minus what its reasons name. `npm run test-cert-long` runs them locally.

## Subject and PICS

The OTA requestor subject deletes the file it verified: a real device reboots into the new image, this one does not, and a downloaded file that outlives its update short-circuits every later query, so without the delete a case could drive exactly one update. It also declares `canConsent`, since it implements `requestUserConsent`; without the declaration it sends no `RequestorCanConsent` and refuses any update whose provider asks for consent.

The controller declares the OTA provider and administrator roles the cases rest on. chip-tool declares `0` for them, so a case it cannot drive skips before commissioning rather than failing its first step. `MCORE.OTA.HTTPS` is `0`: `SoftwareUpdateManager` serves from its own catalog over BDX and answers no https URI.

TC-SU-3.4 declares the matterjs device flavor: chip's `ota-requestor-app` exits once it has applied an image, which `--autoApplyImage` asks it to do, and every step of that case is about the apply. TC-SU-1.1 runs everywhere — chip's `ota-provider-app` exits at startup unless given `-f` or `-o`, and the harness now supplies a placeholder image where the case named none.

## Notes

`assertAppArgsRoles` in `cert-dsl.ts` throws a plain `Error`. `@matter/testing` carries no `@matter/general` dependency by design, and the three declaration guards beside it throw the same way.

The plan-literal three-minute path has not been exercised: every local attempt died on a shared Docker container another session kept recreating. The daily scheduled run is its first real exercise. A shortened run says so in its own evidence: the check that the DUT echoed the scripted `DelayedActionTime` is a real pass or fail on every run, and a separate check records whether the value scripted was the plan's.

## Verification

`npm run build`, `npm run format-verify`, `npm run lint` and `npm test` all clean. The hermetic SU specs pass 30/30 and `test-cert-framework` passed in CI. All five SU cases and the four BDX cases pass on the matterjs device flavor, nine evidence bundles for nine cases.


Apollon77:
All three Copilot findings verified against the code and fixed in 07e0e4823.

**Stale blob after delete — correct, and worse than reported.** `PersistedFileDesignator.openBlob()` caches its `Blob`, and neither `delete()` nor `writeFromStream()` cleared that cache. The instance is reused for the life of the node (`OtaSoftwareUpdateRequestorServer.getDownloadLocation()` holds it in `internal.downloadLocation`), so after the first download every later `openBlob()` answered with the first image — not only after a delete, but after any subsequent write. TC-SU-3.3 drives three updates against one requestor, so its second and third validated the first image rather than what had just arrived. It passed because all three stage the same payload, which makes it exactly the kind of check that proves nothing.

Fixed in the library rather than in the subject: both `writeFromStream()` and `delete()` now clear the cache. `PersistedFileDesignatorTest.ts` covers write-open-write, delete-write-open, and open-delete-open; reverting either line turns three of its four tests red (`expected 'first' to equal 'second'`).

**Missing coverage for the long-running gate — correct.** Added two tests in `cert-test.test.ts` (skipped and counted with the flag absent, run with it set) and two in `evidence.test.ts` (`longRunningSkips` persisted, and absent when nothing was skipped that way). Replacing the gate's condition with `false` fails the first: `expected [ 1, 2 ] to deeply equal [ 2 ]`.

**JSDoc attachment — correct.** `longRunningEnabled()` had been inserted between the malformed-PICS comment and `stepPicsMet()`. Moved above it.

Gates after the fixes: `build`, `format-verify`, `lint` and `npm test` all clean; the five SU cases still pass on the matterjs flavor.

Apollon77:
All three fixed in 5813d58b6.

**Shared recording and shared script queue — correct, and one defect rather than two.** The provider's record of what it answered and its queue of scripted answers both lived on the controller-wide endpoint, while `CertNodeApi` that reads and writes them is a single node's. Both are now `Map`s keyed on the `PeerAddress` the command arrived from, and `OtaExchangeRecording` resolves only on its own peer's query. Reachability today is theoretical — no case in this PR commissions two requestors against one controller — but TC-SU-2.x and TC-SU-4.1 both will, and the failure is silent: a foreign query settles the wait, consumes the scripted answer, and lands in another node's evidence.

**The `DelayedActionTime` assertion — correct, and the sharpest of the three.** A shortened run scripted 1s and compared against the same helper, so TC-SU-3.2/3.4 passed while the plans' "DelayedActionTime - 3 minutes" went untested.

The first suggested remedy does not work here: both requestors compute `max(delayedActionTime, floor)`, so scripting 180 means waiting 180 whatever the floor is. Accelerating only the TH's notion of elapsed time would mean letting a requestor *ignore* a provider's stated delay, which is a worse conformance hole than the floor override this PR already adds.

So the second: `delayedActionTimeCheck()` records `unverified` with an `accepted` reason on a shortened run — naming that the plan's value is not what the DUT sent, and that the `MATTER_CERT_LONG_RUNNING` run scripts and tests it — and a real pass/fail on a plan-literal one. TC-SU-3.2 and TC-SU-3.4 each carry that check now, so the bundle states what the run gave up. The other claims in those steps are unaffected.

Separately, chip's `ota-provider-app` starts again: it exits at startup unless given `-f` or `-o`, and `-f` is `access(R_OK)`-checked as it parses (`[SWU] Either an OTA file or image list file must be specified`, then `chipDie`). The harness writes a placeholder into the device's own directory and names it where the case named none, so TC-SU-1.1 no longer declares a flavor.

Gates: `build`, `format-verify`, `lint` and `npm test` clean; all five SU cases and the four BDX cases pass on the matterjs flavor.

Apollon77:
Three more Copilot findings, all correct, fixed in 128380eee — plus the cert-bins failure from the previous run, which turned out to be a library defect this PR armed.

**The `DelayedActionTime` check came back a second time, so I reshaped it rather than patched it again.** The previous fix made it return `unverified` unconditionally on a shortened run, which meant a dropped or altered field could no longer fail. The shape was wrong: one helper both chose the scripted value and judged the answer, with a verdict that changed with an environment variable. It is now two checks with one job each — `delayedActionTimeCheck()` asks whether the DUT echoed the value the case scripted, a pass or fail on every run, and `planDelayCoverageCheck()` separately records whether that value was the plan's three minutes. A dropped field now fails everywhere, and the coverage gap is still stated.

**`RunDeviceRecord.appArgs` recorded the declaration rather than what the app got.** `CertDevice` now reports the arguments it was started with and `deviceRecordsFor` prefers them, exactly as it already did for `appVariant` and for the same reason: the harness adds what an app cannot start without, and a bundle naming only the declaration omits an argument that changed the app's behaviour.

**The PR body's TC-SU-1.1 flavor claim was stale** after the restriction was removed. Corrected.

**The `matterjs vs cert-bins` failure is a pre-existing defect in `Peers.ts`, which this PR does not otherwise touch.** `#instrumentBasicInformation` attaches an observer that watches for a node's structure read and removes itself once seeding latches. A node that goes away before it ever seeds keeps that observer, and `close()` emits `lifecycle.changed` after the behaviors are gone, so `#evaluateSeeded` reads state that throws `uninitialized-dependency`:

```
ERROR PartLifecycle Unhandled error in dut.peer2.lifecycle.changed handler: [uninitialized-dependency]
  at Peers.#evaluateSeeded (packages/node/src/node/client/Peers.ts:707)
  at onChanged (packages/node/src/node/client/Peers.ts:690)
  at ClientNode.close (packages/node/src/endpoint/Endpoint.ts:1000)
  at ClientNode.decommission
```

TC-SU-1.1 is the first case in the suite to decommission two peers, which is what made it reachable — whether the second peer seeds before teardown is a race, which is why only one leg failed. The observer now detaches when the node stops being readable as well as when it seeds, and the seeding check returns early for a node that is no longer active.

That fix has no unit test. Reproducing "commissioned but never seeded" needs either a reset on `ClientNodeLifecycle` or a device shape `MockSite` cannot express, and adding production API for testability is the wrong trade. The reproduction is the cert-bins leg, which this run should turn green.

Gates: `build`, `format-verify`, `lint` and `npm test` all clean. The SU cases could not be re-run locally — the Docker daemon on this machine is down — so CI is what exercises them.

Apollon77:
Both fixed in 983c4f1ee, and the one claim this PR could not verify is now verified.

**The `Peers` fix has a regression test.** Naming `NodeLifecycleSeededTest` is what made it possible — it holds the recipe for an unseeded peer (delete it, then `peers.forAddress()` recreates it with no structure read). I had searched the wrong directory and grepped for the wrong symbol, concluded it could not be reproduced, and said so; that was wrong.

The first attempt was also vacuous: it closed the peer directly and reverting the guard did not fail it. The CI trace goes through `decommission() → delete() → close()`, and `delete()` is what tears the behaviors down before `lifecycle.changed` is emitted. With `delete()` the test reproduces exactly — reverting the guard fails with the line CI printed:

```
Unhandled error in controller1.peer2.lifecycle.changed handler: [uninitialized-dependency] controller1.peer2 is not initialized
```

**The certification guidance described the two checks the wrong way round** — written before the split and not updated. `delayedActionTimeCheck` is the verdict that always passes or fails against the scripted value; `planDelayCoverageCheck` is the one that reports `unverified` on a shortened run. Corrected.

**The plan-literal three-minute path has now run.** Every earlier attempt died on a shared Docker container; with the daemon restarted, `MATTER_CERT_LONG_RUNNING=1` against the matterjs device took 6m 8s and both cases pass, with the DUT sending and echoing `DelayedActionTime 180s` and `planDelayCoverageCheck` reporting `pass` rather than `unverified`. The PR body's note that this path was unexercised no longer applies.

All nine SU and BDX cases also pass locally with the `Peers` fix in. Gates: `build`, `format-verify`, `lint` and `npm test` clean.

Apollon77:
Fixed in a552f8f9e, though not the way the comment proposed.

The finding is correct and reachable: TC-SU-3.4 step 3 scripts `Discontinue` and then drives another update, so it exercised exactly this. It passed only because a later `QueryImage` clears the stale in-progress entry — the case's "the entire OTA process is restarted" claim rested on that recovery rather than on a provider left in a usable state.

Factoring the cleanup into a helper and calling it before returning would not work from here: the in-progress bookkeeping (`#updateInProgressDetails`, `#removeInProgressDetails`) is private to `OtaSoftwareUpdateProviderServer`, so a subclass cannot perform those transitions. It would also leave the bypass in place and duplicate side effects the base class already owns.

The question the finding really raises is why the bypass exists at all. Only `AwaitNextAction` needs the provider's own handling suppressed, because it must keep the peer's BDX registration so the requestor's next attempt can use what it already downloaded. Every other answer should settle normally. So a deferral still bypasses, and any other scripted action now runs `super.applyUpdateRequest` and replaces only the `Action` in the response. The provider is no longer left holding an update it has answered the last word on.

TC-SU-3.4 step 3 still answers `Discontinue` and still restarts from offset 0, now without depending on the stale-entry path. All five SU cases and the four BDX cases pass on the matterjs flavor; `build`, `format-verify`, `lint` and `npm test` are clean.

Apollon77:
All three fixed in 3a9de8998. Two of them were findings in code the previous round changed, so both were reshaped rather than guarded again.

**The `Peers` seeding observer — my own regression, and the comment is right about the consequence.** I used one predicate for two different questions. Restricting it to `Active` made the observer detach while a node was merely `Initializing`, and since a construction-status change is not a structure change, nothing would reattach it: such a node could never seed.

The two questions are now separate. `isReadable` decides whether state can be read *now* and stays `Active`-only, because an initializing node has nothing to answer with yet. `isGone` decides whether to give up and covers `Destroying`/`Destroyed`/`Crashed` — the same triple `runCommissioning` already uses. Neither direction of the mistake survives: a reader cannot treat "not yet" as "go ahead", and a waiter cannot give up on a node that is still coming up.

**The scripted `Discontinue` — correct, and the right fix is not to sequence the side effects but to stop overlaying at all.** The provider has a refusal path of its own: it answers `Discontinue` for an update it holds no consent for. So the controller now withdraws that update's consent and lets the provider refuse for itself. Nothing is overlaid, the provider's own cancellation bookkeeping runs, and `#applyAllowed` no longer reports an apply the requestor was refused.

TC-SU-3.4 step 3's evidence shows the difference: the refused apply is now recorded `unverified` ("the TH had not asked to apply the image when the DUT stopped waiting") instead of "the TH asked to apply the image and the DUT allowed it", while the DUT still answers `Discontinue` and the restart still begins at offset 0.

`AwaitNextAction` still states its answer directly. That is the one action the provider has no path to, which is also why suppressing its side effects is correct rather than convenient — the requestor's next attempt needs the BDX registration and the image it already downloaded. The scripted-answer type says which two actions are meaningful and what each does.

**The 1024-byte floor** now reports `unverified` where the TH proposed less, with the reason. A pass there was certification evidence for a requirement nothing exercised.

All five SU cases and the four BDX cases pass on the matterjs flavor. `build`, `format-verify`, `lint` and `npm test` clean.

